PR #2, opened 2020-12-04
Starting to move components from prototype repo

arschles:
Moving most of the code from https://github.com/osscda/kedahttp over. Includes:

- the scaler
- the interceptor (called a proxy in the old repo)
- the CLI
- the operator

## Todo

- [x] Fix errors like this (when you create a new `HTTPScaledObject`: 
```
│ manager 2021-01-21T00:44:48.996Z    ERROR    controllers.HTTPScaledObject    Creating service    {"reconciler.appObjects": "addObject │
│ s", "HTTPScaledObject.name": "myapp", "HTTPScaledObject.namespace": "kedahttp", "error": "the namespace of the provided object does n │
│ ot match the namespace sent on the request"}                                                                                          │
```
- [x] Make ancillary image names configurable (interceptor, scaler, etc...)
- [x] ~Make the operator (optionally) install an ingress controller and create an `Ingress`~ - will do in a follow-up. issue forthcoming
- [x] ~Make plan for writing unit tests (begin implementing if possible)~ - will do in a follow-up. issue forthcoming
- [x] Do not fail if controller tries to delete underlying resource that doesn't exist
- [x] Manual testing
- [x] Fix docker build for operator (#5)
- [x] Make scaler ping interceptors for queue size (#4)
- [x] Update the external scaler protobuf file and regenerate code (#3)
- [x] Fix the helm chart for the operator - it currently installs properly but the deployment never scales up with the following problem in its status field:

```shell
  Type           Status  Reason
  ----           ------  ------
  Available      False   MinimumReplicasUnavailable
  Progressing    True    ReplicaSetUpdated
OldReplicaSets:  <none>
NewReplicaSet:   keda-http-addon-controller-manager-7bbd5fd985 (1/1 replicas created)
Events:
  Type    Reason             Age    From                   Message
  ----    ------             ----   ----                   -------
  Normal  ScalingReplicaSet  2m41s  deployment-controller  Scaled up replica set keda-http-addon-controller-manager-7bbd5fd985 to 1
```

## Architecture

See below for reference:

arch

~References for the Rust-based operator:~ We've decided to write the operator in Go, so the below no longer applies.

~- https://github.com/deislabs/krustlet/blob/master/crates/kubelet/src/kubelet.rs#L157-L198~
~- https://dzone.com/articles/writing-a-kubernetes-crd-controller-in-rust~


tomkerkhove:
I'll add @zroubalik for this since it's mainly code, let me know if we have some docs ready for review

arschles:
@tomkerkhove @zroubalik we have some code and docs but the system is not yet moved over and fully working (the main TODO is finishing the operator). Is it ok for us to merge to `main` with a disclaimer on the README?

tomkerkhove:
Fine by me, what do you think @zroubalik ?

zroubalik:
@arschles so you plan to use Rust for the operator? Any particular reason behind this? We are usin Operator SDK (Kubebuilder) framework for our KEDA operator. I think that it makes a lot of sense to unify.

tomkerkhove:
I tend to agree on that though

arschles:
Yup, planning to use Rust for the operator. I'm finding that the code is probably going to be more sustainable in the long term for a few reasons: (1) there's less of the boilerplate code because a macro creates the necessary machinery for a CRD behind the scenes), (2) nothing needs to be generated prior to build time and (3) the reconciliation code is ending up to be shorter, partially a result of (1) and (2)

Rust is a more complex language on a few metrics so I'm flexible and if you both would rather the operator be written in Go. Not much of the operator is built yet so I shouldn't have much of an issue transcoding what's already there. Let me know.

tomkerkhove:
For sake of project alignment I would go with Go to use one stack but I don't do active coding so will leave it up to @zroubalik

zroubalik:
Yeah Rust is great language, but honestly I'd prefer for this to be written the similar way KEDA Operator is being done, thus Go + Operator-SDK (Kubebuilder). There's not that much boilerplate in the recent versions, the same for the CRD generation and it's (Go for operators) already well proven and used by many other projects, I don't see that much operators written in Rust.

arschles:
@zroubalik ok, sounds good. I started a Go operator yesterday (https://github.com/kedacore/http-add-on/pull/2/commits/a89454fa6dbd2d1aac23b1ac7e53e799b8e6d38b) so I'll continue down that path :+1:

tomkerkhove:
Thank you for your flexibility @arschles !

zroubalik:
@arschles thanks!

arschles:
My pleasure. I'm keen to write some Rust for Kubernetes, as some colleagues found some [good success](https://github.com/deislabs/krustlet) with it. I'll keep it in my mind and if it makes sense for something else later, I'll write a proposal for it.

khaosdoctor:
Hey folks! New here, I'm Lucas, I've been working with Aaron on this project, nice to meet you all and I'm super excited to make this project work 🥳 

I removed the old rust-related code and all the `cargo` files for now, it's still accessible in git's history so we haven't lost that code if we need it in the future. Since it's @arschles' repo that's being the source here, I forked his repo and I'm opening PRs over there too to merge in this branch so things should upstream automatically (https://github.com/arschles/http-add-on/pull/8).

tomkerkhove:
Thanks for your help @khaosdoctor !

PR #15, opened 2021-01-27
Improve status spec

### LINKED ISSUE #9: Improve the status field of HTTPScaledObjects
Currently, the `Status` field is not updated

## Expected Behavior

- The `status` field needs to be updated properly after an `HTTPScaledObject` gets created
- The `status` field should have a `conditions` list, similar to the one in `ScaledObject`s. That way, there is an auditable log of mutations of the `HTTPScaledObject`

## Actual Behavior

The status field remains blank

  1. `kubectl create -f examples/httpscaledobject.yaml`
  2. Observe application components like `Service` and `Deployment` get created
  3. `kubectl get -n kedahttp xkcd -o yaml`
  4. Observe that there is no `status` field

## Specifications

N/A

  - **KEDA Version:** *Please elaborate*
  - **Platform & Version:** *Please elaborate*
  - **Kubernetes Version:** *Please elaborate*
  - **Scaler(s):** *Please elaborate*

khaosdoctor:
Fixes #9


arschles:
@khaosdoctor let me know when you're ready for a review on this. I'll hold off for now since it's a draft.

khaosdoctor:
@arschles Just testing it now, but the code is ready. Will update the status here.

What I did was to create a `condition_provider.go` file that has a more fluent interface on creating new conditions to the `httpso` interface (based the types on [Keda's original typing](https://github.com/kedacore/keda/blob/f11984582ca412d0de526b0e4022b1d268e02d56/api/v1alpha1/condition_types.go#L20). Now the conditions can be created as follows:

```go	
condition:= *v1alpha1.CreateCondition(v1alpha1.Error,v1.ConditionFalse,v1alpha1.AppServiceTerminationError).SetMessage(err.Error())
httpso.AddCondition(condition).SaveStatus(logger, rec.Client)
```

khaosdoctor:
Updated all commits due to DCO check, I think it needs to be re-run

operator/api/v1alpha1/httpscaledobject_types.go
```diff
@@ -41,6 +77,22 @@ const (
 	Unknown HTTPScaledObjectCreationStatus = "Unknown"
 )
 
+// Condition to store the condition state
+type HTTPScaledObjectCondition struct {
```
zroubalik, line 81: I am curious, do you think that it is doable to use core KEDA Condition Type for this: https://github.com/kedacore/keda/blob/main/api/v1alpha1/condition_types.go#L20 ?

khaosdoctor: Do you mean using as a module instead of copying?

arschles: @zroubalik I think this was autogenerated like the KEDA core condition was. Better to stick with this?

operator/api/v1alpha1/condition_provider.go
```diff
@@ -0,0 +1,54 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// SaveStatus will trigger an object update to save the current status
+// conditions
+func (httpso *HTTPScaledObject) SaveStatus(
+	logger logr.Logger,
+	cl client.Client,
+) {
+	ctx := context.TODO()
```
arschles, line 18: @khaosdoctor can you have `SaveStatus` take in a context instead of this?

operator/controllers/external_scaler.go
```diff
@@ -9,6 +9,7 @@ import (
 	"github.com/kedacore/http-add-on/pkg/k8s"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
arschles, line 12: can you rename this import to `metav1`? doing so helps distinguish from `corev1`

operator/controllers/httpscaledobject_controller.go
```diff
@@ -117,57 +114,33 @@ func (rec *HTTPScaledObjectReconciler) Reconcile(req ctrl.Request) (ctrl.Result,
...
 		}
 
 		return ctrl.Result{}, err
 	}
 
-	// If all goes well, set the ready status to true
-	if allReady(httpso) {
-		httpso.Status.Ready = true
-	}
-	updateStatus(ctx, logger, rec.Client, httpso)
```
arschles, line 163: where do we update the status to complete?

operator/controllers/httpscaledobject_controller.go
```diff
@@ -117,57 +114,33 @@ func (rec *HTTPScaledObjectReconciler) Reconcile(req ctrl.Request) (ctrl.Result,
...
 	// success reconciling
 	logger.Info("Reconcile success")
 	finalHTTPSo := &httpv1alpha1.HTTPScaledObject{}
 	rec.Client.Get(ctx, types.NamespacedName{
-		Namespace: "kedahttp",
-		Name:      "myapp",
+		Namespace: httpso.Namespace,
+		Name:      httpso.Name,
 	}, finalHTTPSo)
 	logger.Info("Final HTTPScaledObject post-reconcile:", "object", *finalHTTPSo)
```
arschles, line 145: We can remove this debug message and the previous `rec.Client.Get` call

operator/controllers/httpscaledobject_logic.go
```diff
@@ -7,25 +7,19 @@ import (
 	"github.com/kedacore/http-add-on/pkg/k8s"
 	apierrs "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
arschles, line 10: same here regarding `metav1`

arschles (changes requested):

operator/api/v1alpha1/condition_provider.go
```diff
@@ -0,0 +1,65 @@
...
+		Name:      httpso.Name,
+	}, tmpHTTPSo); err != nil {
+		logger.Error(err, "HTTPScaledObject not found on test check")
+	} else {
+		logger.Info("Found HTTPScaledObject on test check", "scaled object", *tmpHTTPSo)
+	}
+
+	tmpHTTPSo.Status = httpso.Status
+	err := cl.Status().Update(ctx, tmpHTTPSo)
+	httpso = tmpHTTPSo
```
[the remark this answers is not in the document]
khaosdoctor: Done

operator/api/v1alpha1/condition_provider.go
khaosdoctor: Yes it's possible, I'll do it

operator/controllers/external_scaler.go
khaosdoctor: Done

operator/controllers/httpscaledobject_controller.go
khaosdoctor: We don't have a condition to set it as complete, I'll post a comment with an idea here, see what you think of it

khaosdoctor:
@zroubalik @arschles Regarding the conditions, I think we should follow a different approach on this, please tell me what you think of it.

Instead of using conditions and status as we're doing now (almost as a log) I suggest we use the Kubernetes Event API to post events as logs and have only two conditions:

- HTTPScaledObjectReady: Will be added when all the resources have been created
- HTTPScaledObjectNotReady: Will be added when CRD is created or when an update is in progress, which will be distinguished in the `reason` field and then transitioned to the ready condition again

In the future, we can use the status field to add the ingress host and/or service IP for the user app.

All other logs such as: Creating Deployment, Creating Scaler and so on would be logged as events using the `EventRecorder` object. Does that make sense?

zroubalik:
@khaosdoctor bear in mind, that Events are removed from k8s etcd after some period of time and checking operator log could be pretty anoying as the log could be quite huge :) On the other hand checking Status on some resource is quite easy and the information there is pernament 🤷‍♂️

operator/api/v1alpha1/httpscaledobject_types.go
zroubalik: Yeah I meant using a module. But I missed that you are using the autogeneration for this, in the KEDA core we added conditions when there wasn't autogeneration capabilities for this.

So I guess that you can keep the autogenerated stuff, if it is more convenient for you:)

khaosdoctor:
> @khaosdoctor bear in mind, that Events are removed from k8s etcd after some period of time and checking operator log could be pretty anoying as the log could be quite huge :) On the other hand checking Status on some resource is quite easy and the information there is pernament 🤷‍♂️

Seeing from this perspective you're right, let's stick with the statuses 😅

khaosdoctor:
**Notice: a74babb upgrades controller version to 0.8.4 so #21 can be closed**

khaosdoctor:
Ok, PR is done, everything working as expected:

image

operator/controllers/httpscaledobject_controller.go
khaosdoctor: Added on 1c83ed0

khaosdoctor:
Hey everyone. I'll close this PR because I'll open it in another branch so it does not make the development of new features blocked by it.

PR #56, opened 2021-02-09
Add magefiles

### LINKED ISSUE #42: Improve contrib experience with magefiles
Today there are a lot of makefiles to run scripts in the project, they're fine but take a while. The `docker-push-all` task runs all the builds sequentially while they could be running in parallel, this is just an example.

I propose to replace makefiles for [magefiles](https://magefile.org/) as @arschles pointed out in one of our discussions and port the scripts to these files. This will allow us to have a better handling of files and allow us to do more complex tasks in the dev environment.

> Plus, to improve contribex we can also version and ship the mage binary along with the repository to make the usage easier.

khaosdoctor:
_Adds magefiles to replace makefiles_

### Checklist

Fixes #42


magefile.go
```diff
@@ -0,0 +1,287 @@
...
+// interceptor and scaler images on the same SHA
+func UpgradeOperator(namespace string, imageRepository string) error {
+	if namespace == "" {
+		namespace = DEFAULT_NAMESPACE
+	}
+
+	if err := sh.RunV(
+		"helm",
+		"upgrade",
+		DEFAULT_NAMESPACE,
```
arschles, line 198: shouldn't this be the helm chart name instead?

arschles (changes requested):
@khaosdoctor a few comments. can you also keep the `Makefile` for now? we can phase it out in a follow-up after things seem to be working ok

magefile.go
khaosdoctor: Yes, the default namespace and the chart name are the same, I must have changed it as a bulk edit. Will correct it

khaosdoctor:
> @khaosdoctor a few comments. can you also keep the `Makefile` for now? we can phase it out in a follow-up after things seem to be working ok

Since I already deleted it from here, can we remove those Makefiles and I create a mage binary so we don't need to have mage installed?

magefile.go
khaosdoctor: Changed on 6e9629d

khaosdoctor: Updated on 6e9629d

khaosdoctor:
The docs are the only thing missing now

khaosdoctor:
Added docs, these are good to go

arschles:
>Since I already deleted it from here, can we remove those Makefiles and I create a mage binary so we don't need to have mage installed?

@khaosdoctor I'd like to leave the Makefile in completely until we all get comfortable using mage (or the binary). at that point, I'm perfectly fine removing

docs/contributing.md
```diff
@@ -0,0 +1,80 @@
...
+- `./build -h` shows the general help
+
+Most of the commands are simple, and we have a few commands that chain other commands together, for reference on chains,
+check the [Magefile](../magefile.go) source code. Below is a list of the most common build commands
+
+> All commands are case insensitive, so `buildAll` and `buildall` are the same.
+
+In the root directory:
+
+- `./build buildAll`: Builds all the binaries for local testing.
```
arschles, line 61: could you call this command `all`, so it reads `./build all`?

docs/contributing.md
```diff
@@ -0,0 +1,80 @@
...
+- `./build upgradeOperator [namespace] <image>`: Will install the add-on in the given `namespace` if not installed, or
+  update it using the provided `image`.
+
+> The default values for the `namespace` if not provided (when passed as `""`, like `./build upgradeOperator "" image`) is `kedahttp`
+
+In the operator directory:
+
+- `./operator/build Manifests`: Builds all the manifest files for Kubernetes, it's important to build after every change
+  to a Kustomize annotation.
+- `./operator/build All`: Generates the operator.
```
arschles, line 80: could you put these commands in the top-level `build` command, so we have one place to do everything?

magefile.go
```diff
@@ -0,0 +1,287 @@
...
+
+const (
+	SCALER      ModuleName = "scaler"
+	INTERCEPTOR ModuleName = "interceptor"
+	OPERATOR    ModuleName = "operator"
+)
+
+// --- Utils --- //
+
+func getGitSHA() (string) {
```
arschles, line 30: no need for the parentheses around the `string` return type

operator/scripts/download-controller-gen.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
+CONTROLLER_GEN_TMP_DIR=$(mktemp -d)
+cd $CONTROLLER_GEN_TMP_DIR
+
+go mod init tmp
+echo "Downloading controller gen"
+go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.4.1
```
arschles, line 8: I think a `go install` would work better here. in 1.16, an `install` won't modify the project's `go.mod` or `go.sum` file, which is nice. what do you think?

arschles (changes requested):
@khaosdoctor this is great. one request: can we not check in the `build` binaries? it's inevitable that we'll forget to update the binary when we update the `Magefile` sometime in the future, and I think it's a fair tradeoff to ask developers to download `mage` before beginning

operator/scripts/download-controller-gen.sh
khaosdoctor: Yeah it'd be better, I'll change it

docs/contributing.md
khaosdoctor: right away

khaosdoctor: Done

docs/contributing.md
khaosdoctor: Done!

khaosdoctor:
@arschles done with the previous review

arschles (changes requested):
@khaosdoctor this looks great except for one more piece - can we not have a `./build` binary? Instead, can we ask people to download mage and run `mage <command>`?

khaosdoctor:
> @khaosdoctor this looks great except for one more piece - can we not have a `./build` binary? Instead, can we ask people to download mage and run `mage <command>`?

Done on 4a93697

khaosdoctor:
Closing this to try to fix DCO

Issue #12, opened 2021-01-27
Use an external DNS service to automatically configure subdomains or custom domains


arschles:
### Use-Case

Currently, you get an external IP address for a new app you create. That means you have to manually configure DNS to complete the app creation process.

### Specification

It would be nice to remove that extra manual step. Possibly use https://github.com/kubernetes-sigs/external-dns to automate some/all of that process

khaosdoctor:
Needs #36 and #10

khaosdoctor:
I'll get on to it after #7

arschles:
Out of scope. See #82

PR #73, opened 2021-02-24
Creating ingress resources when apps are created

### LINKED ISSUE #10: Use an ingress controller
Instead of creating a `Service` with `type: LoadBalancer` on it for the interceptors, use a single ingress controller (it should be installed by the helm chart) and create an `Ingress` for each new app.

### Use-Case

Users should be able to use their existing ingress controller and have the operator integrate with it. Three use cases:

- A user has a brand new, vanilla k8s cluster and wants to install an ingress controller. they `helm install` the chart and the operator spits out `Ingress` resources
- A user has already installed the chart on their cluster and wants to reuse their existing ingress controller, so they `helm install` the chart and turn off the ingress controller option. the operator still spits out `Ingress` resources and the existing ingress controller picks them up properly
- A user has a new k8s cluster on their cloud provider, which has a built-in ingress controller, so they `helm install` the chart and the operator spits out `Ingress` objects and the cloud provider ingress picks them up properly

### Specification

- The helm chart should deploy an ingress controller
    - Have the helm chart take a dependency on the nginx ingress controller chart
    - Allow users to override that dependency, or turn it off completely if they already have one
- The operator should deploy `Ingress` objects

arschles:
When an `HTTPScaledObject` is created anew, it now has an optional `IngressHost` field which, if specified, will trigger the operator to create an `Ingress` object to route requests from the specified host to the backend interceptor. This relates to #12 because, when that is integrated, the host may not need to be specified

### Checklist

- [x] Update and generate the schema to include the `useIngress` field
    - [x] Also move the generated CRD to the helm chart
- [x] ~Add a dependency in the helm chart for an ingress controller~ - we're going to not do this and require users to have created their own ingress controller. This is because there are too many possible ingress controllers and configurations for each to reasonably support all the possible options
- [x] Add docs on setting up an ingress controller

Fixes #10


tomkerkhove:
Docs look OK to me

arschles:
after the March 2, 2021 community standup, I'm not sure if we want this anymore? In that meeting, we discussed removing functionality from the operator that creates application `Deployment`s and `Service`s (#78) but we also discussed restricting the scope more generally as well. In #82, I wrote a more general document about the scope of this project and proposed that we avoid anything outside of what's strictly necessary to "respond to new, modified, or deleted `HTTPScaledObject`s, and when one is created, the add on only creates the infrastructure needed specifically to accomplish autoscaling" (this is from what I proposed in the [#82](https://github.com/kedacore/http-add-on/pull/82/files#diff-bd94d4b03da9ab69e636bc807ca892e53b532aec635152e94072322577bf5b20R9))

@tomkerkhove @zroubalik @ahmelsayed WDYT?

tomkerkhove:
I think that's the sweet spot indeed, all the abstractions on top of that could be another tool, but not perse KEDA I think - Sorry!

arschles:
closing, out of scope

PR #78, opened 2021-03-01
Not creating app deployments 

### LINKED ISSUE #35: Allow users to scale an existing Deployment
### Use-Case

Currently you can't start scaling an existing `Deployment`, you can only create a new one by submitting a new `HTTPScaledObject`. It would be helpful to allow someone to submit a new `HTTPScaledObject` to begin scaling an already existing `Deployment`.

This idea came from #31 in [this comment](https://github.com/kedacore/http-add-on/pull/31/files#r568569492)

### Specification

- [ ] User should be able to specify an existing `Deployment` in an `HTTPScaledObject`
    - They will also need to specify the `Service` name and port to which traffic should be forwarded to their app
- [ ] KEDA and the HTTP add on should begin routing traffic to their app and scaling based on HTTP traffic

### LINKED ISSUE #74: Add documentation for the HTTPScaledObject schema
We only have the code and example YAML at the moment. We should document this much better in the repo

arschles:
Fixes https://github.com/kedacore/http-add-on/issues/35

Previous to this PR, the operator created an application `Deployment` and `Service`. This pull request removes that functionality and requires that a user creates those things themself. In doing so, we're focusing the HTTP add on directly on:

- Autoscaling
- The minimum amount of traffic routing and distribution functionality needed to achieve autoscaling

This PR also changes the schema of the CRD. The new CRD needs to allow the user to specify what to scale and how to route. To achieve that, I've added a `scaleTargetRef` field to match the pattern that KEDA core has. It looks similar to this:

```yaml
    scaleTargetRef:
        deployment: xkcd
        service: xkcd
        port: 8080
```

The above would instruct the operator to scale a `Deployment` called `xkcd` and route traffic through a `Service` called `xkcd` on port 8080.

### Checklist

- [x] Regenerate the CRD and deepcopy code
- [x] Remove the option to create a new deployment/service - the http-add-on should only scale existing deployments

Fixes #35
Fixes #74


tomkerkhove (commented):
Can we add some docs please?

arschles:
@tomkerkhove yup, documentation added here. I also removed the functionality that creates application `Deployment`s and `Service`s, as discussed in the March 2, 2021 community meeting.

docs/ref/http_scaled_object.md
```diff
@@ -0,0 +1,33 @@
+# The `HTTPScaledObject`
+
+Each `HTTPScaledObject` looks approximately like the below:
+
+```yaml
+kind: HTTPScaledObject
+apiVersion: http.keda.sh/v1alpha1
+metadata:
+    name: xkcd
+spec:
```
tomkerkhove, line 10: As a user, I'd be wondering how I configure when it should scale?

docs/walkthrough.md
```diff
@@ -1,17 +1,43 @@
...
+```shell
+k port-forward svc/xkcd-interceptor-proxy -n ${NAMESPACE} 8080:80 
+```
+
+### Routing to the Right `Service`
+
+As said above, you need to route your HTTP traffic to the `Service` that the add on has created. If you have existing systems - like an ingress controller - you'll need to anticipate the name of these created `Service`s. Each one will be named consistently like so, in the same namespace as the `HTTPScaledObject` and your application (i.e. `$NAMESPACE`):
+
+```shell
+<deployment name>-interceptor-proxy
```
tomkerkhove, line 40: Let's also make the interceptor service configurable?

```yaml
kind: HTTPScaledObject
metadata:
    name: xkcd
spec:
    scaleTargetRef:
        deployment: xkcd
        service: xkcd
        port: 8080
    service:
        name: my-interceptor-proxy
        port: 9999
        type: LoadBalancer
```

Or just surface it like this:

```yaml
kind: HTTPScaledObject
metadata:
    name: xkcd
spec:
    scaleTargetRef:
        deployment: xkcd
        service: xkcd
        port: 8080
    service:
        name: my-interceptor-proxy
        spec:
            selector:
              app: MyApp
            ports:
            - protocol: TCP
		      port: 80
		      targetPort: 9376
```

Happy to open an issue if you want.

tomkerkhove (commented):
This is looking good, thanks for making the change!

Added a few suggestions and was wondering why we are adding the xkcd Helm chart? Maybe we should leave that one out?

docs/walkthrough.md
arschles: @tomkerkhove yea, can you create an issue for that? One thing I wonder is, should we not have it create a `Service` at all? By not doing so, we won't impose limitations on how they're configured. 

We could then add to the documentation that users can create a `Service` with a specific label selector if they want.

docs/ref/http_scaled_object.md
arschles: Not sure what you mean. Is there another time it should scale besides when HTTP requests come into (or stop coming into) the system?

docs/walkthrough.md
tomkerkhove: What do you think @zroubalik?

I think having it part of the spec might be useful? Users and reading docs doesn't always go that well 😁

arschles: Very good point :laughing:. Would it be possible to merge this and then either add the `service` field or remove `Service` creation altogether in a follow up (depending on what you and @zroubalik decide)?

Also, if this helps:

In the [scope doc](https://github.com/kedacore/http-add-on/blob/main/docs/scope.md#the-scope-of-the-keda-http-add-on), it says the following is in scope: `Route HTTP traffic from a given source to an arbitrary HTTP server, as far as we need to efficiently accomplish (1)`. That could help the decision. I'd like to prefer having the operator create less things, but I'll defer to what you and @zroubalik think.

docs/ref/http_scaled_object.md
arschles: I see what you mean now. We do 0->n scaling by default and you're right, there's no way to change that at the moment. Would you mind creating an issue to add that configurability?

tomkerkhove: I would most definately add some more clear docs around this then, because I had the impression we supported `0->n` scaling and was confused about defining when to scale.

tomkerkhove: https://github.com/kedacore/http-add-on/issues/101

docs/walkthrough.md
tomkerkhove: Moving to another PR is fine for me, but this feels like something we need to have to be able to work imo since this is for the interceptor.

If they have to create the service for KEDA, then it's too much infrastructure imo.

It's all about how you approach it. When I create an HTTPScaledObject you could add a field that surfaces the port + IP (if need be) so that it's all in that CRD when doing `k get HTTPScaledObject` and users don't even know about the service that was created. (not saying we should but a possibility).

arschles: Ok, yea I tend to agree with that @tomkerkhove . I'm happy to add the fields that surface port, IP, and maybe other fields. This PR is blocking some other work (since it makes such big fundamental changes to the operator). If there's nothing else you'd like to change in it, I'd like us to merge it and I will concurrently open a follow-up PR that adds those fields. What do you think @tomkerkhove @zroubalik ?

tomkerkhove: Fine by me!

tomkerkhove (approved):

operator/api/v1alpha1/httpscaledobject_types.go
```diff
@@ -113,18 +113,23 @@ type ReplicaStruct struct {
...
 	// (optional) Replica information
 	//+optional
 	Replicas ReplicaStruct `json:"replicas,omitempty"`
 }
 
-// TODO: Add ingress configurations
+// ScaleTargetRef contains all the details about an HTTP application to scale and route to
+type ScaleTargetRef struct {
+	// The name of the deployment to scale according to HTTP traffic
+	Deployment string `json:"deployment"`
```
zroubalik, line 127: Just a nit. Do you want to target just `Deployments`, what about `StatefulSet` or `CustomResources` with `/scale` subresoure? Might be worth checking how is this handled in ScaledObject: https://github.com/kedacore/keda/blob/6dc4c3524ba6278276cd1338cc7e4e9a7dd88963/api/v1alpha1/scaledobject_types.go#L75-L81

zroubalik (approved):
Looking good

operator/api/v1alpha1/httpscaledobject_types.go
arschles: I would like to target `Deployment`s for now, and open it up to other HTTP workloads that folks might want to scale in the future. Does that sound ok to you @zroubalik ?

docs/walkthrough.md
arschles: https://github.com/kedacore/http-add-on/issues/107 is the follow-up issue

operator/api/v1alpha1/httpscaledobject_types.go
arschles: @zroubalik I am going to merge this PR now, and we can create a follow-up if you think we need to support `StatefulSet` or other resources with a `/scale` subresource right now.

PR #85, opened 2021-03-02
Add min and max replicas

### LINKED ISSUE #62: Expose min and max replicas on the HTTPScaledObject
### Use-Case

Users should be able to specify the minimum (0 or more) and maximum (1 or above) replicas that their app should have. These values map to the `ScaledObject` created behind the scenes.

khaosdoctor:
Adds new API to provide min and max replicas to scaledobjects

### Checklist

Fixes #62


operator/controllers/config/app_info.go
```diff
@@ -6,6 +6,8 @@ type AppInfo struct {
 	Name                 string
 	Port                 int32
 	Image                string
+	MinReplicas          int32
```
arschles, line 9: @khaosdoctor can we leave these out of the app config and use `httpso.Spec.MinReplicas`/`MaxReplicas` instead?

arschles (changes requested):
looks good @khaosdoctor, one request only

examples/httpscaledobject.yaml
```diff
@@ -6,3 +6,5 @@ spec:
     app_name: xkcd
     app_image: arschles/xkcd
     port: 8080
+    min_replicas: 5
+    max_replicas: 10
```
tomkerkhove, line 10: How about this?
```suggestion
    replicas:
      min: 5
      max: 10
```

khaosdoctor: I like it, will update the API to make this the default

khaosdoctor: Resolved in last commit

operator/controllers/config/app_info.go
khaosdoctor: Also resolved in last commit

charts/keda-http-operator/crds/httpscaledobjects.http.keda.sh.yaml
```diff
@@ -4,7 +4,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
```
tomkerkhove, line 4: We'll need to update this one

tomkerkhove (approved):

arschles (approved):
Looks great, thanks @khaosdoctor !

PR #93, opened 2021-03-10
Updating arch diagram according to new scope

arschles:
The scope of the project has changed and it does not modify ingresses, services, or deployments related to the target application. This PR reflects that change in the architecture diagram

### Checklist


tomkerkhove (approved):
LGTM but I think having some color-coding indicating who owns what would help

arschles:
Thanks @tomkerkhove. I added #99 to track that work.

PR #94, opened 2021-03-11
Remove unstructured objects

### LINKED ISSUE #72: Use kedacore/keda ScaledObject types
Instead of using `unstructured.Unstructured` in the functionality where we [create `ScaledObject`s](https://github.com/kedacore/http-add-on/blob/main/pkg/k8s/scaledobject.go#L30), we should use the actual [`ScaledObject` type](https://pkg.go.dev/github.com/kedacore/keda/v2@v2.1.0/api/v1alpha1#ScaledObject) in KEDA core itself. From experience, we'll probably need to do #64 first.

khaosdoctor:
Remove unstructured type and use keda typings

### Checklist

Fixes #72
Signed-off-by: Lucas Santos <lhs.santoss@gmail.com>


khaosdoctor:
Built from wrong branch

Issue #97, opened 2021-03-15
Use the external push gRPC protocol for external scaler => interceptor communication


arschles:
We currently use a simple HTTP/JSON protocol for the scaler to communicate with the interceptor. Because this protocol is just request/response, it has limitations on latency that can cause very slow scale up times, which are especially noticeable when scaling up from 0 replicas.

It would be helpful to have some kind of two way protocol rather than what we have now. The [external push](https://keda.sh/docs/2.1/scalers/external-push/) scaler provides just this. We should take this and use it in the scaler.

### Use-Case

As mentioned above, this would be useful to speed up scale-up (and down) latencies. This may help with #60 as well, but that's not the primary goal here.

benjaminhuo:
Is there any plan to implement the external push interface?

arschles:
@benjaminhuo we would likely implement the external push interface or something similar but simpler (because my guess is we won't need all the complexity of the current external push interface)

benjaminhuo:
Looking forward to seeing this added!

arschles:
Same here @benjaminhuo. @ajanth97 and I are going to come up with a design in the near future and the current plan is he's going to implement it. This will probably all start in earnest after the big #206 is merged.

Let me know if you'd like to get involved - you're certainly welcome to!

benjaminhuo:
@arschles Yes sure, I'm happy to get involved!

ajanth97:
Awesome

benjaminhuo:
> ok. let's start by collaborating on a design. I've started a document [here](https://hackmd.io/ttpSY5KQQtGIqbAzyKex3g). feel free to leave comments etc...
> 
> Note: If you haven't used hackmd before, you have to login with your GitHub (or it supports some other SSO types I think) account before you can edit or comment.

That's a great place to start

arschles:
This is ongoing. @ajanth97 is working on it

tomkerkhove:
Any update on this?

ajanth97:
Hi Sorry, I will work on this soon

arschles:
>Is it still relevant?

FWIW, I think it still is 😄

sanjay7178:
I took a look at the codebase and the scaler is still polling each interceptor pod over HTTP on a timer to grab queue counts. `StreamIsActive` also works on a polling loop rather than reacting to actual changes, so the latency issue described here is still present.

A few things I'd love clarity on before diving in:
[...]

PR #105, opened 2021-03-17
Refactoring the interceptor

### LINKED ISSUE #61: Interceptor doesn't always "hold" request when there are no replicas available
When there's no app available, we shouldn't get a connection refused. The request should hang until the app scales up from 0 to 1. Sometimes the interceptor doesn't do this.

We should also expose a timeout to the user via the `HTTPScaledObject` so that they can configure how long they'd like the interceptor to hold the request

arschles:
This PR fixes a bug in the interceptor in which it would not wait for a connection to be established if there was no server already waiting to accept conns, by adding a retry mechanism for the `DialContext` function in the HTTP proxy.

It does a few more things:

- Adding more testing to the proxy
- Refactoring the proxy to use standard net/http (gives it more control over the response lifecycle)
- Using the standard library test framework, because [stretchr/testify/suite](https://godoc.org/github.com/stretchr/testify/suite) was not adding much ([stretchr/testify/require](https://godoc.org/github.com/stretchr/testify/require) is still in use though)
- Add a check in the proxy for the number of replicas in the deployment. If there are 0 replicas, it holds the request without trying to do any network operations until the replica count goes up
    - This is done in preparation for #97 - we want the interceptor to be able to use this knowledge to push a prompt to KEDA to start the scaling operation

### Checklist

- [x] Have the interceptor watch the backing deployment and only send requests when they're at 1 or more replicas (when we get to #97, we can push to the scaler when we see a request come in and 0 replicas on the deployment)
- [x] Fix tests
- [x] Add more tests for the proxy handler
- [x] Make the operator start up the interceptor with the proper new env vars
- [x] Figure out the connection timeout tests (the ones that are failing right now)
- [x] Make the interceptor time out on the `waitFunc()` (see `newForwardingHandler`)

Fixes #61


arschles:
@tomkerkhove @khaosdoctor @zroubalik this was a bigger change and one of the big blockers for releasing alpha 2. Can you PTAL?

khaosdoctor (approved):

khaosdoctor:
I think that's fine, there's a test failing though.

zroubalik (approved):
LGTM, there's failing test though

arschles:
Ok, tests fixed. Thanks for the reviews @khaosdoctor @zroubalik !

PR #106, opened 2021-03-18
Use embedded scaledobject templates

### LINKED ISSUE #72: Use kedacore/keda ScaledObject types

khaosdoctor:
Removing the unstructured code object in the scaled object creation and started using go embedded templates as per what @coderanger suggested.

### Checklist

Fixes #72


khaosdoctor:
@arschles can you please give it a review?

coderanger:
Nifty, much simpler than my old thing using vfsgen.

zroubalik:
I am curious, is there a specific reason, why don't you want to import the type from keda? You don't want to mess with dependencies? Or what is the benefit?

coderanger:
I explained it a bit on Slack but the main reason is you cannot use concrete structs with Server Side Apply and if this controller isn't using that already, it should soon. So if you have to use an Unstructured anyway, might as well make it pretty. As a secondary issue, yes dependency hell is nice to avoid.

That first issue might some day go away when controller-gen grows support for the typed apply builder syntax but that's a long way off right now (core support is going in to client-go for 1.21 as a "no really we mean it" alpha).

zroubalik:
Thanks, that make sense.

> That first issue might some day go away when controller-gen grows support for the typed apply builder syntax but that's a long way off right now (core support is going in to client-go for 1.21 as a "no really we mean it" alpha).

Is this being tracked in some kubebuilder (controller-runtime) issue? I wasn't able to find anything.

coderanger:
Not yet, not that far along :) The core support in client-go was only merged a few days ago.

arschles:
>You don't want to mess with dependencies?

@zroubalik this certainly is a nice thing, though

pkg/k8s/scaledobject.go
```diff
@@ -52,33 +61,30 @@ func NewScaledObject(
...
 		labels[k] = vIface
 	}
+
+	tpl, err := template.ParseFS(scaledObjectTemplateFS, "templates/scaledobject.yaml")
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	var scaledObjectTemplateBuffer bytes.Buffer
+	tpl.Execute(&scaledObjectTemplateBuffer, map[string]interface{}{
```
arschles, line 71: @khaosdoctor can you check the error here please?

pkg/k8s/scaledobject.go
```diff
@@ -52,33 +61,30 @@ func NewScaledObject(
...
+		"MinReplicas": minReplicas,
+		"MaxReplicas": maxReplicas,
+		"DeploymentName": deploymentName,
+		"ScalerAddress": scalerAddress,
+	})
+
+	var decodedYaml map[string]interface{}
+	decodeErr := yaml.Unmarshal(scaledObjectTemplateBuffer.Bytes(), &decodedYaml)
+	if decodeErr != nil {
+		log.Fatal(err)
```
arschles, line 84: @khaosdoctor instead of `log.Fatal`, can you please return the error?

arschles (changes requested):
@khaosdoctor a few requests

khaosdoctor:
Will do :)!

khaosdoctor:
@arschles done!

arschles (approved):
looks great @khaosdoctor ✌️ 🚀

PR #112, opened 2021-03-23
Expose service configuration

### LINKED ISSUE #107: Add configuration parameters for operator-created services
After https://github.com/kedacore/http-add-on/pull/78#pullrequestreview-615892955, the operator will create the interceptor and external scaler, but not target application resources.

### Use-Case

As part of that PR, the operator will create the `Service` that routes to the interceptor, and users will need to point traffic to that service. Since the `Service` is part of the critical request path and it's where users will primarily interact with the system, it would be nice to be able to customize it. 

### Specification

- [ ] Users should be able to say at least the `type` of the `Service` (i.e. `LoadBalancer`, `ClusterIP`)

khaosdoctor:
Adding exposed configuration, for now the only configs that can be changed are the service exposed port and type because labels are defined by the add-on.

### Checklist

Fixes #107


khaosdoctor:
@arschles @tomkerkhove can you PTAL? Just finished

arschles (changes requested):
LGTM @khaosdoctor. Can you add documentation to [the reference](https://github.com/kedacore/http-add-on/blob/main/docs/ref/http_scaled_object.md)?

khaosdoctor:
Will do!

arschles (changes requested):
@khaosdoctor one more thing - can you add a test for this too?

khaosdoctor:
Commited the requested changes, but haven't done the test yet because I'm having some issues running Go in my PC, also I think we'll need to run `mage manifests` again to rebuild the RBAC config and CRD struct

docs/ref/http_scaled_object.md
```diff
@@ -24,10 +24,30 @@ This is the primary and most important part of the `spec` because it describes (
 
 This is the name of the `Deployment` to scale. It must exist in the same namespace as this `HTTPScaledObject` and shouldn't be managed by any other autoscaling system. This means that there should not be any `ScaledObject` already created for this `Deployment`. The HTTP add on will manage a `ScaledObject` internally.
 
-### `service`
+## Replicas
```
arschles, line 27: @khaosdoctor can you add this to the example yaml above?

arschles (changes requested):
@khaosdoctor will this mean we need to make changes to https://github.com/kedacore/charts/pull/137 as well?

khaosdoctor:
> @khaosdoctor will this mean we need to make changes to [kedacore/charts#137](https://github.com/kedacore/charts/pull/137) as well?

Most likely yes, since the CRD will change (we may need to think in an automated way to do this)

arschles:
@khaosdoctor in that case, do you think we need to bump the version of the CRD?

khaosdoctor:
> @khaosdoctor in that case, do you think we need to bump the version of the CRD?

This might be a good idea, since this is a breaking change I propose a major version increase

arschles:
>This might be a good idea, since this is a breaking change I propose a major version increase

would it, though? if you're just adding a new field to the spec, it won't break old `HTTPScaledObject`s. we could bump from `v1alpha1` -> `v1alpha2`

khaosdoctor:
> >This might be a good idea, since this is a breaking change I propose a major version increase
> 
> 
> 
> would it, though? if you're just adding a new field to the spec, it won't break old `HTTPScaledObject`s. we could bump from `v1alpha1` -> `v1alpha2`

Agreed, we can keep the version and change the release :)

arschles:
@khaosdoctor this looks great. I want to get the 0.1.0 release out before we merge this though. stand by :)

khaosdoctor:
Closing it as it's gonna be superseeded by #206

PR #114, opened 2021-03-29
changing default image names in Makefile to GH container registry

arschles:
The `Makefile` still had `arschles/keda-http-{scaler, operator, interceptor}` image names in it. Now that we've moved the official images to GH container registry, this patch changes the image names to `ghcr.io/kedacore/http-add-on-{scaler, operator, interceptor}` image names

### Checklist

- [x] (N/A) A PR is opened to update the documentation on [our docs repo](https://github.com/kedacore/keda-docs)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 GIT_TAG?=$(shell git rev-parse --short HEAD)
-SCALER_DOCKER_IMG?=arschles/keda-http-scaler:${GIT_TAG}
-INTERCEPTOR_DOCKER_IMG?=arschles/keda-http-interceptor:${GIT_TAG}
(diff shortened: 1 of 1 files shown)


arschles:
cc/ @khaosdoctor

tomkerkhove (approved):

Makefile
```diff
@@ -1,7 +1,7 @@
 GIT_TAG?=$(shell git rev-parse --short HEAD)
```
tomkerkhove, line 1: Is this for experimental and/or official releases?

Because for official we should use version instead

arschles: @tomkerkhove this is for experimental / dev releases. If someone wants to install an official release, we'd have them do a `helm install` directly.

PR #129, opened 2021-04-12
Using proper go versions in dockerfiles

arschles:
The dockerfiles for scaler, interceptor and operator all defaulted to Go version 1.14. Since we're now using the [`embed`](https://godoc.org/embed) package, which is available in 1.16 and above, we need to change that.

### Checklist

- [X] Commits are signed with Developer Certificate of Origin (DCO)


arschles:
cc/ @tomkerkhove

arschles:
Superseded by #133. Closing

PR #125, opened 2021-04-06
Move charts to keda/charts

khaosdoctor:
Moving charts from this repo to the official [kedacore/charts](https://github.com/kedacore/charts) repo
### Checklist

- [x] Commits are signed with Developer Certificate of Origin (DCO)
- [x] A PR is opened to update the documentation on [our docs repo](https://github.com/kedacore/keda-docs)

Fixes #104

PR #131, opened 2021-04-12
adding required parameters to the xkcd chart values file

arschles:
Doing a `helm install` on the `charts/xkcd` chart failed due to several default values being missing in the `values.yaml` file. This PR adds those default values

### Checklist

cc/ @tomkerkhove


tomkerkhove (approved):
LGTM, out of curiosity - Why do we still have this?

Are we using it as an app-chart? Might be good to move that to /examples maybe?

arschles:
thanks @tomkerkhove. you're right, we're going to move the `xkcd` chart to `examples/` in https://github.com/kedacore/http-add-on/pull/125#pullrequestreview-635874933

PR #154, opened 2021-04-24
Update faq.md

embano1:
Remove requirement of Service Mesh in Knative

### Checklist

  - [`README.md`](/README.md)
  - [`docs/`](./docs)

Fixes #

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -25,7 +25,7 @@ KNative serving and KEDA-HTTP both have core support for autoscaling, including 
 - Complex deployment strategies like [blue-green](https://knative.dev/docs/serving/samples/blue-green-deployment/)
 - Supporting other autoscaling mechanisms beyond the built-in [HPA](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/), such as the [KNative Pod Autoscaler (KPA)](https://knative.dev/docs/serving/autoscaling/autoscaling-concepts/#knative-pod-autoscaler-kpa)
 
-Additionally, KNative requires a service mesh, while KEDA-HTTP does not out of the box (support for that is forthcoming).
+Additionally, KNative supports a service mesh, while KEDA-HTTP does not out of the box (support for that is forthcoming).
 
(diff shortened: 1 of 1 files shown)


docs/faq.md
```diff
@@ -25,7 +25,7 @@ KNative serving and KEDA-HTTP both have core support for autoscaling, including
 - Complex deployment strategies like [blue-green](https://knative.dev/docs/serving/samples/blue-green-deployment/)
 - Supporting other autoscaling mechanisms beyond the built-in [HPA](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/), such as the [KNative Pod Autoscaler (KPA)](https://knative.dev/docs/serving/autoscaling/autoscaling-concepts/#knative-pod-autoscaler-kpa)
 
-Additionally, KNative requires a service mesh, while KEDA-HTTP does not out of the box (support for that is forthcoming).
+Additionally, KNative supports a service mesh, while KEDA-HTTP does not out of the box (support for that is forthcoming).
```
arschles, line 28: @embano1 to be specific, is it eventing that supports but does not require a mesh? the [install instructions for serving](https://knative.dev/docs/install/install-serving-with-yaml/#install-a-networking-layer) seem to require what they call a networking layer

arschles (changes requested):
thanks so much for this @embano1 - just left a question

docs/faq.md
arschles: also, I think this section in the FAQ should probably be specific to serving. thoughts on that?

embano1: Initially, Knative required Istio but that has changed and one can run Knative (Serving) with a ingress controller as documented in the README. That's why I changed it to "supports" bc one can still use a service mesh with Knative Serving, but it's not a requirement.

> also, I think this section in the FAQ should probably be specific to serving. thoughts on that?

Yeah, you're right. I think we should drop the eventing related comparisons as they are a better fit in the KEDA core repo.

arschles: @embano1 got ya. I'll merge this PR and then would you be up for submitting a follow-up that specifies serving in this FAQ?

arschles (approved):

PR #155, opened 2021-04-27
Make imagePullPolicy configurable

### LINKED ISSUE #115: Make image pull policy configurable on helm chart
Right now, it's hard-coded to `Always`. We should make that configurable.

### Use-Case

Some folks might rather not always re-pull images, especially for the scaler and interceptor, which might be spinning up and down repeatedly, if they add more apps or, in the case of the interceptor, traffic increases happen.

### Specification

- [ ] A new value is in `values.yaml` called `operator.pullPolicy` that parameterizes the operator's pull policy
- [ ] The operator has a new env var called `INTERCEPTOR_PULL_POLICY` that controls the interceptor's pull policy when the operator creates it
- [ ] Similarly, the operator has a `SCALER_PULL_POLICY` that controls the scaler's pull policy

khaosdoctor:
Makes pull policies configurable

### Checklist

  - [`README.md`](/README.md)
  - [`docs/`](./docs)

Fixes #115


operator/controllers/external_scaler.go
```diff
@@ -47,6 +48,7 @@ func createExternalScaler(
 			},
 		},
 		k8s.Labels(appInfo.ExternalScalerDeploymentName()),
+		env.GetOr("SCALER_PULL_POLICY", "Always"),
```
arschles, line 51: @khaosdoctor can we pass this and the other `*_PULL_POLICY` values down from the controller's `main.go`. It's more work, but it's also helpful to have most/all of the config centralized

operator/controllers/interceptor.go
```diff
@@ -62,6 +63,7 @@ func createInterceptor(
 		},
 		interceptorEnvs,
 		k8s.Labels(appInfo.InterceptorDeploymentName()),
+		env.GetOr("INTERCEPTOR_PULL_POLICY", "Always"),
```
arschles, line 66: same here regarding centralizing config

arschles (changes requested):

operator/controllers/config/config.go
```diff
@@ -23,17 +24,20 @@ func NewInterceptorFromEnv() (*Interceptor, error) {
 	}
 	adminPort := env.GetInt32Or("KEDAHTTP_OPERATOR_INTERCEPTOR_ADMIN_PORT", 8090)
 	proxyPort := env.GetInt32Or("KEDAHTTP_OPERATOR_INTERCEPTOR_PROXY_PORT", 8091)
+	pullPolicy := env.GetOr("INTERCEPTOR_PULL_POLICY", "Always")
```
arschles, line 27: @khaosdoctor what are your thoughts on checking this for a valid value (i.e. `Always`, etc...)?

arschles (approved):
Just one question that we can do in a follow-up, otherwise 👍

PR #168, opened 2021-05-18
Added targetMetric field in CRD

ajanth97:
I added the target metric field in the CRD.
### Checklist

  - [`README.md`](/README.md)
  - [`docs/`](./docs)

Fixes #


go.mod
```diff
@@ -3,8 +3,14 @@ module github.com/kedacore/http-add-on
 go 1.16
 
 require (
+	github.com/armon/consul-api v0.0.0-20180202201655-eb2c6b5be1b6 // indirect
```
arschles, line 6: @ajanth97 I'm not seeing why all these new dependencies need to be added. could you try running `go mod tidy`? it _should_ get rid of these

scaler/handlers.go
```diff
@@ -60,15 +63,30 @@ func (e *impl) StreamIsActive(
 	}
 }
 
+func getTargetMetric() int64 {
+	targetMetricStr, err := env.Get("KEDA_HTTP_SCALER_TARGET_METRIC")
```
arschles, line 67: @ajanth97 can you read this environment variable in the `main` function [near where all the other configs are read](https://github.com/kedacore/http-add-on/blob/main/scaler/main.go#L22-L40), and then pass the value into this function?

arschles (changes requested):
@ajanth97 thanks for this! two requests

tomkerkhove:
Don't forget to update the Helm chart as well please on kedacore/charts

go.mod
ajanth97: Yes, that fixed it

scaler/handlers.go
ajanth97: I made new changes, please check it

go.mod
```diff
@@ -17,6 +17,7 @@ require (
 	google.golang.org/grpc v1.33.2
 	google.golang.org/protobuf v1.26.0
 	k8s.io/api v0.20.4
+	k8s.io/apiextensions-apiserver v0.20.2 // indirect
```
arschles, line 20: @ajanth97 does this still show up in the `go.mod` after you run this command:

```shell
go mod tidy
```

?

operator/config/rbac/role.yaml
```diff
@@ -4,7 +4,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   creationTimestamp: null
-  name: keda-http-addon-manager-role
+  name: keda-http-manager-role
```
arschles, line 7: @ajanth97 did the `mage generate` command change this?

arschles (changes requested):
@ajanth97 can you remove the `mage` binary from this PR? As a sidenote, you might want to consider putting it somewhere in your `PATH` so you can use it anywhere

go.mod
ajanth97: Yes I removed the mage binary and added it to my PATH. And yes the go.mod file still shows those depenedencies

operator/config/rbac/role.yaml
ajanth97: I used mage dockerBuildAll, I am not really sure what caused this change :(

arschles:
Corresponding chart change in https://github.com/kedacore/charts/pull/153

arschles:
@ajanth97 this looks good! I want to hold on this until the 0.1.0 release, which should be very soon. stand by!

arschles:
feel free to resolve conflicts in the meantime though

ajanth97:
Sure

arschles:
@ajanth97 this is ready to merge. can you resolve conflicts?

ajanth97:
> @ajanth97 this is ready to merge. can you resolve conflicts?

Thanks, Sure

charts/keda-http-operator/crds/http.keda.sh_httpscaledobjects.yaml
```diff
@@ -85,6 +93,10 @@ spec:
                 - port
                 - service
                 type: object
+              targetMetric:
+                description: (optional) Target metric value
```
khaosdoctor, line 97: Could you add a bit more explanation here? I think people will get lost on what this is for

khaosdoctor (changes requested):
Amazing work. I just got confused on the name `targetMetric`, maybe this will generate some confusion as it is a bit generic. 

Maybe `scaleRequestAmount` (I'm terrible at naming)

charts/keda-http-operator/crds/http.keda.sh_httpscaledobjects.yaml
ajanth97: Thank you, Yeah sure I will add some explanation. As for the naming yeah lets change it if its causing confusion. However I feel `scaleRequestAmount` is a bit counterintuitive since the lower this value the faster it scales. Let me know your thoughts

khaosdoctor: What about `queueMinScaleSize`?

ajanth97: That makes more sense but is a bit too long, How about `targetQueueSize` ?

khaosdoctor: The thing about the `target` is that it seems that it'll only scale once it hits this exact number. What about `scaleReqCount`?

ajanth97: The reason I used `target` is because I though its relatable to other scaling methods such as CPU for example where the ratio of `current/target ` is considered. I don't understand why we need to have `scale` in the name for this metric. Maybe we could have a poll ? Lets use whatever most people find most intuitive ?

khaosdoctor: I was talking to @arschles today, we came up with `maxParallelRequests`. As this metric is going to be the amount of requests per each app. What do you think?

ajanth97: Yes that makes more sense as it is the amount of requests. Let's go with that then.

tomkerkhove: What do you think @zroubalik? I find it a bit technical and was thinking more along the lines of `requestsPerSecond` since, in the future, that is going to be the trigger for SMI (for example)

khaosdoctor: That is another point too, we need to make it simple for end users. I agree with this naming

ajanth97: In my opinion `targetQueueSize` feels the most intuitive since the HPA is comparing the `currentQueueSize ` vs `targetQueueSize`

arschles: I'm +1 to @ajanth97 (he and I are talking about this now). 

Also, this isn't necessarily the number of request per second, or the target number of requests per second. The time window will depend on polling intervals and the semantics of this value are more nuanced. Also, in the SMI world, we might not be using requests per polling interval, so I think we should assume we'll use a different field in the `HTTPScaledObject` for that.

khaosdoctor: Let's go with that :)

tomkerkhove: Frankly, if you do not know how it works using `queue` is very confusing as you are using HTTP traffic and not a queue.

When scaling HTTP, they don't care about the inner workings - They just want it to scale to meet the demand.

tomkerkhove: But if @zroubalik likes the name as well then that's fine for me.

arschles: @tomkerkhove how about `targetPendingRequests` then? I agree with you RE the `queue` nomenclature, but I think if a user wants to tweak this, then they will be interested enough in inner workings to learn about how the Addon treats pending requests.

arschles (commented):
This is good with me, but I'd like to hear from you @tomkerkhove about the name. this PR currently is using `targetPendingRequests`

charts/keda-http-operator/crds/http.keda.sh_httpscaledobjects.yaml
tomkerkhove: > @tomkerkhove how about `targetPendingRequests` then? I agree with you RE the `queue` nomenclature, but I think if a user wants to tweak this, then they will be interested enough in inner workings to learn about how the Addon treats pending requests.

I think you might be surprised about how many people will (want) to define their own number, without fully understanding/caring how it works internally.

Name is fine for me.

RELEASE-PROCESS 2.md
```diff
@@ -0,0 +1,43 @@
+# Release Process
```
arschles, line 1: @ajanth97 can you remove this document? it appears to be a duplicate of the release process document

docs/developing 2.md
```diff
@@ -0,0 +1,163 @@
+# Developing
```
arschles, line 1: @ajanth97 same here regarding the duplicate document

arschles (changes requested):

charts/keda-http-operator/crds/http.keda.sh_httpscaledobjects.yaml
arschles: >I think you might be surprised about how many people will (want) to define their own number, without fully understanding/caring how it works internally.

if they do want to do so, and don't understand/care how it works internally, we'll have good documentation (which I'll submit in a follow-up to this PR) to explain it

operator/api/v1alpha1/httpscaledobject_types.go
```diff
@@ -115,6 +115,8 @@ type HTTPScaledObjectSpec struct {
 	// (optional) Replica information
 	//+optional
 	Replicas ReplicaStruct `json:"replicas,omitempty"`
+	//(optional) Target metric value
+	TargetMetric int32 `json:"targetMetric,omitempty" description:"The target metric value for the HPA (Default 100)"`
```
arschles, line 119: @ajanth97 you'll need to change this to `TargetPendingRequests` as well, and then re-generate the CRD

arschles (changes requested):

arschles (approved):

ajanth97:
Thank you @arschles for helping out with this PR

PR #182, opened 2021-06-15
Using structured logging in the interceptor

arschles:
Instead of using the standard library's `log` package in the interceptor, this PR uses [klogr](https://pkg.go.dev/k8s.io/klog/v2/klogr), a [logr](https://pkg.go.dev/github.com/go-logr/logr) implementation for the `klogr` library that we already depend on in this repository.

The advantages of using this library -- and structured logging in general -- over the std lib `log` package are:

- we pass a logger around for functions to use, which can "subclass" them (using `.WithName`, `WithValues`) to their needs
- since we're using [logr](https://pkg.go.dev/github.com/go-logr/logr), we can swap out the underlying implementation to  adapters for other underlying log libraries if we want
- we can log different lines at different levels
- we log variable data in a structured format, rather than in format strings
- we can log data to multiple locations using klog

### Checklist

  - [`README.md`](/README.md)
  - [`docs/`](./docs)

Fixes #


interceptor/forward_wait_func.go
```diff
@@ -1,26 +1,36 @@
...
 	deployName string,
 	totalWait time.Duration,
 ) forwardWaitFunc {
+	logger = logger.WithName("newDeployReplicasForwardWaitFunc")
 	return func() error {
 		deployment, err := deployCache.Get(deployName)
 		if err != nil {
 			// if we didn't get the initial deployment state, bail out
-			return fmt.Errorf("Error getting state for deployment %s (%s)", deployName, err)
+			logger.Error(
```
arschles, line 27: NOTE: this block is significantly changed in https://github.com/kedacore/http-add-on/pull/181/files. The conflict should be resolved favoring the other PR

arschles:
This will be superseded by #206. Closing

PR #198, opened 2021-06-24
setting proper tag version in release script

arschles:
Before this PR, the GitHub action for releasing a tag version was failing as such: 

```
error: invalid tag "ghcr.io/kedacore/http-add-on-operator:${GITHUB_REF#refs/tags/v}": invalid reference format
107
Error: buildx call failed with: error: invalid tag "ghcr.io/kedacore/http-add-on-operator:${GITHUB_REF#refs/tags/v}": invalid reference format
```

(see https://github.com/kedacore/http-add-on/runs/2906000365?check_suite_focus=true#step:6:107 for this log line)

this PR hopefully fixes that. since there is no way to reliably test GH actions locally, we have to wait and see.

### Checklist

  - [`README.md`](/README.md)

Fixes #

### DIFF
--- a/.github/workflows/build_release.yml
+++ b/.github/workflows/build_release.yml
@@ -21,6 +21,11 @@ jobs:
         id: prep
         run: |
           echo ::set-output name=created::$(date -u +'%Y-%m-%dT%H:%M:%SZ')
+      
+      - name: Get the version
+        id: get_version
+        run: |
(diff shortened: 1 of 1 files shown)


arschles:
Since this is necessary to execute the GitHub action for a release, and it cannot be tested without merge to `main`, I am going to self-merge it now.

PR #206, opened 2021-07-01
Multi-tenant interceptor and scaler

arschles:
**This is a large PR that makes the interceptor and external scaler multi-tenant**.

See [below](https://github.com/kedacore/http-add-on/pull/206#issuecomment-900421289) for testing instructions.

Instead of starting up a single interceptor/scaler automatically _per application_, a fleet of interceptors/scalers run and can operate on any application in the same namespace. Interceptors can dynamically proxy requests based on the incoming request, scalers can dynamically report metrics on all applications, and the operator can provide routing information to the interceptor fleet. See https://hackmd.io/@arschles/mutitenant-keda-http-addon for more detailed design information.

In this pull request:

- The operator will not install any `Pod`s into the cluster when any `HTTPScaledObject` is created
- Any given interceptor pod can route a request to any installed application (in the same namespace)
- Interceptors are (still) horizontally scalable
- The operator maintains a "routing table" - a lookup table from hostname to the `Service`, Port`, and `Deployment` of a backing application
- Interceptor pods periodically request the updated routing table from the operator, and update their internal copy
- When the routing table is updated -- for example, when a new `HTTPScaledObject` is created -- the operator pings all interceptor pods to refresh their copy

There are several follow-ups to this PR, including but not limited to:

- https://github.com/kedacore/http-add-on/pull/247
- https://github.com/kedacore/http-add-on/pull/246
- #242

### Checklist

- [x] Update `design.md` and `walkthrough.md` in `docs/`directory
- [x] Commits are signed with Developer Certificate of Origin (DCO)
- [x] Implement the routing table in the operator
- [x] Implement the routing table in the interceptor
- [x] Implement interceptor <--> operator RPC for the routing table
- [x] Remove code from operator that creates new resources
- [x] Make the routing table fetch path configurable in the operator
- [x] Expose metrics on the scaler so that KEDA can scale the interceptors as well
- [x] Clean up status values on the HTTPScaledObject
- [x] Make the external scaler respond based on the incoming `HTTPScaledObject`
    - requires that the queue knows about _all_ hosts
- [x] Fix & add tests:
    - [x] Add tests for the routing table in pkg/routing
    - [x] Add tests for the routing table pinger in the interceptor
    - [x] Add tests for the operator's routing table construction logic
    - [x] round trip tests for the routing table
    - [x] round trip tests for the HTTP request queue
    - [x] Fix broken tests
- [x] Make applicable updates to the helm chart in `kedacore/charts` (https://github.com/kedacore/charts/pull/169)
[...]

PR #246, opened 2021-08-30
Adding endpoint on interceptors to reveal the routing table version history

### LINKED ISSUE #225: Have interceptors record what routing table version(s) they are aware of
In #206, interceptors have to stay up to date on a namespace-global routing table. It would be helpful to provide a way for interceptors to confirm that they have gotten a specific version of the routing table.

### Use-Case

I imagine the most common use case for this functionality would be to ensure that all interceptors are aware of the most up to date routing table.

### Specification

The implementation can vary, but I can think of a few possible and feasible user interfaces:

- Have all interceptors write to some Kubernetes object indicating what version of the routing table they are currently aware of. We may find value in having them append to the object so that some observer can see history of each observer. 
    - One possible negative to this approach is that there may be a significant amount of garbage generated as changes to the routing table mount, and as interceptors come and go
- Expose an RPC endpoint on all interceptors that indicates what version they are aware of (and possibly a history of all versions they've been aware of, again for debuggability).
    - To fetch the currently-known versions of all interceptors, an observer would fetch the endpoints for the interceptor's admin `Service`, then issue the RPC to each interceptor. The scaler currently does something similar with `/queue` endpoints.

Note that since the routing table is stored in a `ConfigMap`, the aforementioned version numbers can be the same thing as the Kubernetes resource version.

cc/ @yaron2 since this is related to #206 and #183

arschles:
This PR adds in-memory storage to the interceptors for the routing table version (the same identifier as the Kube `ConfigMap` resource version), and exposes it via a REST API endpoint.

>~This PR is submitted as a draft and intended to be reviewed and merged after #206 is complete~
>
>**Update:** #206 is merged, and this PR is taken out of draft status. It is ready for review.

### Checklist

- [x] Fix build breakage
  - [`README.md`](/README.md)
- [x] Split the routing table into reader, writer, and versioner interfaces for use in interceptor vs. operator

Fixes #225


arschles:
Closing since #225 was marked stale and closed

PR #269, opened 2021-09-21
Allowing cluster-global operation

### LINKED ISSUE #240: Make interceptor Cluster bound instead of namespace bound
Interceptor should be able to run in an admin namespace (eg: keda-http) and should be able to forward traffic to multiple services in multiple namespaces

### Use-Case
In a scenario where, there are multiple services in multiple namespaces in a given cluster, sometimes it could be costly to add a keda http interceptor per namespace. Rather, just like how ingress controller works, it would be best if the interceptor can reside in an admin namespace and forward traffic to multiple services in multiple namespaces.

### LINKED ISSUE #257: Refactor the routing table config map updater
Currently, `(pkg.routing).StartConfigMapRoutingTableUpdater` needs access to  a `(pkg.queue).Counter` so that it can ensure that the queue data contains _every_ host in the routing table, even if there are no pending requests for it. We implemented the queue this way so that the scaler can receive it from an interceptor and be sure it can report all hosts to KEDA, and KEDA will appropriately scale to zero.

We should change how we do this, though:

- Interceptor should return queue sizes for _only_ hosts that have a nonzero queue
- Scaler should merge the interceptor's queue with the routing table, to ensure that it reports a complete picture to KEDA

>NOTE: This needs to be done after #242, because we introduce the routing table into the scaler in that PR.

arschles:
This patch allows the KEDA HTTP Addon operator, interceptor, and scaler to operate in either namespace-local or cluster-global mode.

In the former mode, the addon only operates on new HTTPScaledObjects that are created in a given namespace. In the latter, it operates on all HTTPScaledObjects in the cluster. **Importantly, only one cluster-
global installation can be run, with no other cluster-global or namespace-local installations.**

>This should be merged after the `v0.2.0` release is finished

### Design

The design of this change is detailed at https://hackmd.io/@arschles/kedahttpclusterglobal

### Checklist

- [x] Stop merging routing table into queue counts in the interceptor. Instead, do this merge on demand (i.e. when KEDA requests counts) in the scaler.
- [x] Add documentation 
- [x] Add configuration to the chart (https://github.com/kedacore/charts/pull/225)
- [x] Manual tests
- [x] Merge memory and fake deployment caches
- [x] enable global watch on interceptor and scaler
- [x] On interceptor startup, ensure all configured durations make sense. For example, polling intervals must be less than timeouts
- [x] On interceptor startup, log non-sensitive config
  - [`README.md`](/README.md)

Fixes #240
Fixes #257 
Closes #259


bketelsen (approved):
LGTM

PR #291, opened 2021-10-07
Refactoring concurrency in scaler's pending HTTP queue fetcher logic

### LINKED ISSUE #139: Add tests to the scaler
The scaler has very little testing (it will have slightly more after #116)

### Use-Case

More test coverage for the scaler

### Specification

- [ ] Tests for the gRPC handler
- [ ] Tests for the queue pinger

### LINKED ISSUE #286: Fix concurrency in scaler requestCounts function
The `requestCounts` function in the scaler has "funky" concurrency:

- It starts up one goroutine per interceptor endpoint inside an `errgroup`. Each of these tries to send that interceptor's counts on a channel called `countsCh`
- It starts up one final goroutine that `range`s over the same `countsCh` channel
- `countsCh` is closed in a `defer`

This design means that, when `requestCounts` returns, none of the updates are processed because everything is waiting for the `defer close(countsCh)`. The updates will happen after it returns. This behavior means that updates to an interceptor might not be viewable immediately after `requestCounts` completes, and that's confusing.

### Use-Case

When code calls `requestCounts`, it should be able to observe updates to all counts within the queue pinger immediately after that function returns.

### Specification

- Refactor `requestCounts` to update all counts before it returns
- Write tests that call `requestCounts` directly

Tell us in detail how this feature should work

arschles:
This patch refactors the `queuePinger`'s `requestCounts` function in the scaler. Highlights:

- Renames `requestCounts` to `fetchAndSaveCounts`
- Creates a new `fetchCounts` function (not method on the `queuePinger`) that does all the heavy lifting
- Changes `fetchAndSaveCounts` to simply call `fetchCounts` and save the results internally to the `queuePinger`
- Adds tests for `fetchCounts`
- Ensures that all counts are fetched when `fetchCounts` returns
- The queue tick duration is now configurable

### Checklist

  - [`README.md`](/README.md)
- [x] Refactor requestCounts to update all counts before it returns
- [x] Write tests that call requestCounts directly
- [x] The queue pinger should not start the background goroutine in its constructor. A separate function call should be required for that.

Fixes #286 
Fixes #139

### DIFF
--- a/docs/developing.md
+++ b/docs/developing.md
@@ -74,11 +74,13 @@ Some of the above commands require several environment variables to be set. You 
 
 >Suffix any `*_IMAGE` variable with `<keda-git-sha>` and the build system will automatically replace it with `sha-$(git rev-parse --short HEAD)`
 
-## Helpful Tips
+## Debugging and Observing Components
 
-The below tips assist with debugging, introspecting, or observing the current state of a running HTTP addon installation. They involve making network requests to cluster-internal (i.e. `ClusterIP` `Service`s). 
+The below tips assist with debugging, introspecting, or observing the current state of a running HTTP addon installation. They involve making network requests to cluster-internal (i.e. `ClusterIP` `Service`s).
 
-There are generally two ways to communicate with these services.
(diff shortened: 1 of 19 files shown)


arschles:
cc/ @asw101

PR #322, opened 2021-11-12
Adding prometheus metrics to the interceptor

### LINKED ISSUE #321: Add prometheus-compatible instrumentation to the interceptor
The interceptor already has several [debugging/observation endpoints](https://github.com/kedacore/http-add-on/blob/main/docs/developing.md#debugging-and-observing-components) that expose several pieces of useful information to be used primarily for debugging purposes. For example, you can fetch a point-in-time copy of the pending queue or the routing table, all via a simple REST API that you can `curl`. These endpoints are accessible on an HTTP server that runs on a dedicated port, separate from the main interceptor (proxy) server. The target client for these endpoints is more than likely a human.

Since the interceptor is in the critical path of end-user HTTP requests, there are many more metrics that _could_ be captured and most likely consumed by a machine.

### Use-Case

I'd like to have the interceptor run an endpoint on which folks can fetch prometheus metrics that describe the current state of the system. For example, it could serve histograms for response codes, counters for total number of requests or timeouts, and so forth.

### Specification

- The interceptor is instrumented with prometheus-compatible metrics
- The interceptor runs a prometheus endpoint
- Documentation is added (probably to `development.md` or a new document) regarding these metrics

arschles:
Adds prometheus instrumentation and a fetch endpoint for those instrumentation

### Checklist

  - [`README.md`](/README.md)

Fixes #321

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -15,6 +15,6 @@ jobs:
     - name: Install Go
       uses: actions/setup-go@v2
       with:
-        go-version: 1.16.x
+        go-version: 1.17.x
     - name: Test
       run: go test ./...
--- a/go.mod
+++ b/go.mod
@@ -21,6 +21,7 @@ require (
 	k8s.io/apimachinery v0.22.2
 	k8s.io/client-go v0.22.2
 	sigs.k8s.io/controller-runtime v0.10.1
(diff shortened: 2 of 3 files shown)


abatilo (approved):

tomkerkhove:
@arschles Should we abandon this PR or not?

JorTurFer:
This is already implemented

PR #326, opened 2021-11-17
Add configMap informer

### LINKED ISSUE #287: Use client-go's informer to replace the custom routing table config map updater
There are quite a few places where we have to consume a kubernetes resource, watch it for changes, and keep a copy of it (or some data derived from a copy of it) in memory and up-to-date. So far, we have written code from scratch that uses a `watch.Interface` to watch for changes, and uses a Kubernetes client to periodically do a full fetch of the Kubernetes resource(s) in question.

In #269, we're beginning to remove some of this custom code in favor of the [informer](https://pkg.go.dev/k8s.io/client-go@v0.22.2/informers) framework for the deployment cache. This issue is to replace all of the other appropriate instances of such code:

- Watching the routing table `ConfigMap` in the scaler
- Watching the routing table `ConfigMap` in the interceptor

### Use-Case

We will be able to delete a nontrivial amount of code by doing this work.

### Specification

- Refactor the code in `https://github.com/kedacore/http-add-on/blob/main/pkg/routing/config_map_updater.go` to use an informer rather than a custom-built `for` loop

tpiperatgod:
Use client-go's informer to replace the custom routing table config map updater.

### Checklist

  - [`README.md`](/README.md)

Fixes #287 

### Additional information

When the configmap named "keda-http-routing-table" changes (e.g., a new httpscaledobjects is created), the external-scaler service will request keda to access the IsActive rpc of the http-add-on in the external-push mode. 

However, since the [minimum resynchronization period](https://github.com/kubernetes/client-go/blob/v0.22.2/tools/cache/shared_informer.go#L475) of the informer is 1s, the IsActive rpc request launched by keda will fail in the first 1s (this has no effect on the subsequent functionality, but there will be error log output).

Should we make StreamIsActive run after a 1 second delay? or just keep it.

And according to the optimization for the external-scaler(and interceptor) mentioned at https://hackmd.io/ttpSY5KQQtGIqbAzyKex3g, it will also solve this problem I think.

Signed-off-by: laminar <fangtian@kubesphere.io>


ajanth97:
@tpiperatgod thanks for this PR. Could you please share the error logs that you came across ?

tpiperatgod:
Sure, the error log is as follows:

```
{"level":"error","ts":1637142750.0702565,"logger":"IsActive","caller":"scaler/handlers.go:72","msg":"Given host was not found in queue count map","host":"myhost.com","allCounts":{},"error":"host 'myhost.com' not found in counts","stacktrace":"main.(*impl).IsActive\n\t/go/src/github.com/kedacore/http-add-on/scaler/handlers.go:72\ngithub.com/kedacore/http-add-on/proto._ExternalScaler_IsActive_Handler\n\t/go/src/github.com/kedacore/http-add-on/proto/scaler_grpc.pb.go:139\ngoogle.golang.org/grpc.(*Server).processUnaryRPC\n\t/go/pkg/mod/google.golang.org/grpc@v1.41.0/server.go:1279\ngoogle.golang.org/grpc.(*Server).handleStream\n\t/go/pkg/mod/google.golang.org/grpc@v1.41.0/server.go:1608\ngoogle.golang.org/grpc.(*Server).serveStreams.func1.2\n\t/go/pkg/mod/google.golang.org/grpc@v1.41.0/server.go:923"}
{"level":"error","ts":1637142750.084912,"caller":"scaler/handlers.go:97","msg":"error getting active status in stream, continuing","error":"host 'myhost.com' not found in counts","stacktrace":"main.(*impl).StreamIsActive\n\t/go/src/github.com/kedacore/http-add-on/scaler/handlers.go:97\ngithub.com/kedacore/http-add-on/proto._ExternalScaler_StreamIsActive_Handler\n\t/go/src/github.com/kedacore/http-add-on/proto/scaler_grpc.pb.go:156\ngoogle.golang.org/grpc.(*Server).processStreamingRPC\n\t/go/pkg/mod/google.golang.org/grpc@v1.41.0/server.go:1532\ngoogle.golang.org/grpc.(*Server).handleStream\n\t/go/pkg/mod/google.golang.org/grpc@v1.41.0/server.go:1612\ngoogle.golang.org/grpc.(*Server).serveStreams.func1.2\n\t/go/pkg/mod/google.golang.org/grpc@v1.41.0/server.go:923"}
```

ajanth97:
@tpiperatgod thank you. I think maybe for now we could reduce the `minimum resynchronization period`. Ideally when the routing table gets updated it should trigger the resync, although this isn't too much of a concern right now. @arschles What do you think ?

interceptor/main.go
```diff
@@ -101,6 +110,13 @@ func main() {
 		return err
 	})
 
+	errGrp.Go(func() error {
+		defer ctxDone()
+		err := configMapInformer.Start(ctx)
+		lggr.Error(err, "config map watcher failed")
+		return err
+	})
+
```
arschles, line 119: I think it would make sense to start the `configMapInformer` from inside of the `StartConfigMapRoutingTableUpdater` so that all of the routing table-related functionality is in one place. What do you think?

arschles (changes requested):
@tpiperatgod this looks great! I only have one suggestion below.

Also, regarding the resynchronization period, we can fetch an initial version of the routing table on startup before we start the informer loop? That way, we'll be able to immediately report `true` for the `IsActive` call, and can watch for changes with the informer from that point onward.

Thoughts?

interceptor/main.go
tpiperatgod: You're right. I will move the `configMapInformer.Start()` into `StartConfigMapRoutingTableUpdater`.

tpiperatgod:
I created a channel to receive the ConfigMap object from eventHandler(addEvtHandler, updateEvtHandler, deleteEvtHandler). When eventhandler get an event, it will pass the ConfigMap object to the StartConfigMapRoutingTableUpdater. This process is now independent with the resynchronization of informer's local cache.

But it still throws the same exception. I will continue to debug it.

tpiperatgod:
Hi folks, here is the flowchart of keda-http-add-on as I understand it, is it correct?

If this flowchart is correct, I think it is possible that `SteamIsActive()` was started too early causing a brief exception in the scaler service (extending the `SteamIsActive()` interval from 5ms to 200ms could circumvent the exception)

I also calculated the time consumption in `StartConfigMapRoutingTableUpdater()`, from the time the latest configmap object is received to the time `updateQueueFromTable()` is executed, it takes 10~60 microseconds, while the original time.ticker is 100 milliseconds, it confused me a bit.:thinking:

arschles:
> here is the flowchart of keda-http-add-on as I understand it, is it correct?

@tpiperatgod yes, I think that flowchart is correct, if I'm reading it properly.

>I think it is possible that SteamIsActive() was started too early causing a brief exception in the scaler service (extending the SteamIsActive() interval from 5ms to 200ms could circumvent the exception)

Yes, I think you're right. The design of `StreamIsActive` is going to be updated as part of https://github.com/kedacore/http-add-on/issues/97 (and the design document you mentioned) so this change would be temporary. 

>I also calculated the time consumption in StartConfigMapRoutingTableUpdater(), from the time the latest configmap object is received to the time updateQueueFromTable() is executed, it takes 10~60 microseconds

This time range is excellent. 

> while the original time.ticker is 100 milliseconds, it confused me a bit.🤔

The original duration was 500ms if I remember correctly. After testing a bit, we lowered it to 100 because it reduced scale-up latencies a bit but didn't put too much more load on the Kubernetes server. In reality, we could reduce it more because there won't be more than 1 scaler replica running at once. In the success case, however, this update duration doesn't matter much because changes should be applied via watch stream events. In other words, this duration is in place so that we can ensure that even if the watch stream is broken, the scaler will have an up-to-date routing table within 100ms (or whatever is configured in the `KEDA_HTTP_SCALER_ROUTING_TABLE_UPDATE_DUR` environment variable).

To overcome this start-up problem, what if you fetched the routing table before reporting that the scaler is healthy via the health/liveness check?

pkg/k8s/config_map_cache_informer.go
```diff
@@ -0,0 +1,132 @@
...
+			fmt.Errorf("informer expected configMap, got %v", obj),
+			"not forwarding event",
+		)
+		return
+	}
+	i.cmObjCh <- cm
+	i.bcaster.Action(watch.Deleted, cm)
+}
+
+func (i *InformerConfigMap) GetCMObjectChannel() chan *corev1.ConfigMap {
```
arschles, line 106: Would it be sufficient to return a read-only channel? i.e. `<-chan *corev1.ConfigMap` Also, can we rename this the full name, rather than `CM`?

```suggestion
func (i *InformerConfigMap) GetConfigMapObjectChannel() <-chan *corev1.ConfigMap {
```

pkg/routing/config_map_updater.go
```diff
@@ -25,47 +21,29 @@ import (
...
+	go func() {
+		defer ctx.Done()
+		err := cmInformer.Start(ctx)
+		if err != nil {
+			informerMsg <- err
+		}
+	}()
+
+	ch := cmInformer.GetCMObjectChannel()
+	defer close(ch)
```
arschles, line 41: don't you need to stop the informer rather than closing this channel?

pkg/routing/config_map_updater_test.go
```diff
@@ -129,6 +127,8 @@ func TestStartUpdateLoop(t *testing.T) {
 	const waitDur = interval * 5
 	time.Sleep(waitDur)
 
+	_, err := fakeGetter.CoreV1().ConfigMaps(ns).Get(ctx, ConfigMapRoutingTableName, metav1.GetOptions{})
```
arschles, line 130:
```suggestion
	_, err := fakeGetter.CoreV1().ConfigMaps(ns).Get(ctx, ConfigMapRoutingTableName, metav1.GetOptions{})
	r.NoError(err)
```

arschles (changes requested):

tpiperatgod:
I think I probably know what the problem is, it's because the value of **KEDA_HTTP_QUEUE_TICK_DURATION** is too big, I adjusted it from the default **500ms** to **450ms**, which can overcome the problem that `IsActive()` doesn't get the count of the host in the queuePinger every time the httpscaledobject is created.

So I considered setting the default value of **KEDA_HTTP_QUEUE_TICK_DURATION** to 450ms or less, but it looks like 450ms would be more reasonable?

Also, I think we can deprecate the **KEDA_HTTP_SCALER_ROUTING_TABLE_UPDATE_DUR** in the scaler service and the **KEDA_HTTP_ROUTING_TABLE_UPDATE_DURATION_MS** in the interceptor service since we now use the informer-watch mechanism instead of the original loops. In the meantime, is it necessary to introduce a new configuration for setting the local cache resynchronization time of the informer (which is independent of the eventHandler trigger)

arschles:
@tpiperatgod that's great to hear that you tracked down the exception. I will do another review of this PR on Monday.

pkg/k8s/config_map_cache_informer.go
```diff
@@ -0,0 +1,141 @@
...
+	cm, ok := newObj.(*corev1.ConfigMap)
+	if !ok {
+		i.lggr.Error(
+			fmt.Errorf("informer expected configMap, got %v", newObj),
+			"not forwarding event",
+		)
+		return
+	}
+	i.cmObjCh <- cm
+	i.bcaster.Action(watch.Modified, cm)
```
arschles, line 99: can we unify these two actions, and have the goroutine listening on this channel just listen on a broadcaster channel?

pkg/routing/config_map_updater.go
```diff
@@ -21,51 +17,35 @@ import (
 //	called ConfigMapRoutingTableName. On either of those events, decodes
 //	that ConfigMap into a routing table and stores the new table into table
 //	using table.Replace(newTable)
+//	- Execute the callback function, if exist
```
arschles, line 20:
```suggestion
//	- Execute the callback function, if one exists
```

pkg/routing/config_map_updater.go
```diff
@@ -21,51 +17,35 @@ import (
...
-	ticker := time.NewTicker(updateEvery)
-	defer ticker.Stop()
+	informerMsg := make(chan error)
+	go func() {
+		defer ctx.Done()
+		err := cmInformer.Start(ctx)
+		if err != nil {
+			informerMsg <- err
+		}
+	}()
```
arschles, line 40: do you think it would be worth using an [errgroup](https://pkg.go.dev/golang.org/x/sync/errgroup) for this

arschles (changes requested):
@tpiperatgod this is making really good progress. I only have a few smaller comments here

interceptor/config/serving.go
```diff
@@ -16,15 +18,9 @@ type Serving struct {
...
-	//
-	// The interceptor will also open a watch stream to the routing table
-	// ConfigMap and attempt to update the routing table on every update.
-	//
-	// Since it does full updates alongside watch stream updates, it can
-	// only process one at a time. Therefore, this is a best effort timeout
-	RoutingTableUpdateDurationMS int `envconfig:"KEDA_HTTP_ROUTING_TABLE_UPDATE_DURATION_MS" default:"500"`
+	// ConfigMapCacheRsyncPeriod is the time interval
+	// for the configmap informer to rsync the local cache.
+	ConfigMapCacheRsyncPeriod time.Duration `envconfig:"KEDA_HTTP_SCALER_CONFIG_MAP_INFORMER_RSYNC_PERIOD" default:"60m"`
```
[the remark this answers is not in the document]
tpiperatgod: I think the two environment variables are already fundamentally different in meaning, which could lead to some misunderstandings in their use. So I tend to use the new one.

pkg/k8s/config_map_cache_informer.go
tpiperatgod: As it should be. The reason why a cmObjCh channel was added was that the code I wrote was causing the broadcaster to block, now that I've fixed the broadcaster blocking problem, everything will communicate through the broadcaster.

pkg/routing/config_map_updater.go
tpiperatgod: I also think errgroup would be more appropriate

pkg/k8s/config_map_cache_informer.go
```diff
@@ -0,0 +1,132 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/informers"
+	infcorev1 "k8s.io/client-go/informers/core/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+)
+
+type InformerConfigMap struct {
```
arschles, line 20: can we call this `InformerConfigMapUpdater` or something similar with `Updater` in the name? I think such a name will help ensure its purpose is clear.

```suggestion
type InformerConfigMapUpdater struct {
```

pkg/k8s/config_map_cache_informer.go
```diff
@@ -0,0 +1,132 @@
...
+		i.lggr.Error(
+			fmt.Errorf("informer expected configMap, got %v", obj),
+			"not forwarding event",
+		)
+		return
+	}
+	i.bcaster.Action(watch.Deleted, cm)
+}
+
+func NewInformerConfigMap(
```
arschles, line 111:
```suggestion
func NewInformerConfigMapUpdater(
```

arschles (commented):
@tpiperatgod this looks great! I have two nit-pick comments, which we can address in a follow-up PR (in the interest of getting this merged) if you'd like. Let me know

(also, thank you for being patient on reviews. I was taking time off last week and am catching up this week)

interceptor/config/serving.go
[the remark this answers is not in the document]
arschles: ok, fine with me. can you create a (draft) PR in the kedacore/charts repository so that the http-add-on chart has the new environment variable, and removes the old ones?

arschles:
@tpiperatgod the diagram you created in https://github.com/kedacore/http-add-on/pull/326#issuecomment-972946943 would be very helpful to add to the project documentation. Is it ok with you if I add it? If so, I can do that.

pkg/k8s/config_map_cache_informer.go
tpiperatgod: I agree with this change.

tpiperatgod:
> @tpiperatgod the diagram you created in [#326 (comment)](https://github.com/kedacore/http-add-on/pull/326#issuecomment-972946943) would be very helpful to add to the project documentation. Is it ok with you if I add it? If so, I can do that.

Feel free to use it. :smile: And if you find any details that are inaccurately expressed, please let me know and I will fix it.

pkg/routing/config_map_updater.go
```diff
@@ -21,49 +19,40 @@ import (
...
-			evtType := evt.Type
-			obj := evt.Object
-			if evtType == watch.Added || evtType == watch.Modified {
-				cm, ok := obj.(*corev1.ConfigMap)
-				// by definition of watchIface, all returned objects should
-				// be assertable to a ConfigMap. In the likely impossible
-				// case that it isn't, just ignore and move on.
-				// This check is just to be defensive.
+	grp, ctx := errgroup.WithContext(ctx)
+	defer ctx.Done()
```
arschles, line 39: `defer ctx.Done()` is effectively a no-op here. I think a better way to go would be the below code. you'd also need to replace the other `defer ctx.Done()` lines with `defer done()`

```suggestion
        ctx, done := context.WithCancel(ctx)
        defer done()
	grp, ctx := errgroup.WithContext(ctx)
```

pkg/routing/config_map_updater.go
```diff
@@ -21,49 +19,40 @@ import (
...
+
+	grp.Go(func() error {
+		defer ctx.Done()
+		return cmInformer.Start(ctx)
+	})
+
+	grp.Go(func() error {
+		for {
```
arschles, line 47:
```suggestion
                defer done()
		for {
```

scaler/main.go
```diff
@@ -62,8 +62,27 @@ func main() {
...
 
+	// This callback function is used to fetch and save
+	// the current queue counts from the interceptor immediately
+	// after updating the routingTable information.
+	callbackWhenRoutingTableUpdate := func() error {
+		if err := pinger.fetchAndSaveCounts(ctx); err != nil {
+			return err
+		}
+		return nil
+	}
```
arschles, line 73: @tpiperatgod instead of fetching and saving counts, what do you think about simply adding the hostname to the scaler's in-memory map and letting the pinger fetch counts as it currently does in its own loop?

tpiperatgod: Can I understand your suggestion in this way: 
1. pass in the **pinger** to `StartConfigMapRoutingTableUpdater`
2. then add the element with key the **hostname** to `pinger.allCounts` in the processing

If so, then I used a generic function `cbFunc` instead of passing in **pinger** directly, considering that **pinger** only runs in the scaler service, while `StartConfigMapRoutingTableUpdater` is used by both the scaler and the interceptor.

As well `fetchAndSaveCounts` is an already existing method, and it doesn't need to pass additional arguments, which helps make `callbackWhenRoutingTableUpdate` a generic function, so maybe I should think it makes sense in this scenario?

Please correct me if I have misunderstood...

tpiperatgod: At the same time, the pinger will keep its loop.

pkg/routing/config_map_updater.go
tpiperatgod: Thanks! I've learned.

scaler/main.go
tpiperatgod: I think it is also workable to pass in **pinger**, and I will give priority to your suggestion.

arschles: >Can I understand your suggestion in this way:


@tpiperatgod that would be one way to do it. it looks like a larger refactor might be in order, which we can certainly do in a follow-up pull request. if this works now as-is, then I'm fine with it!

for the follow-up, one possible refactor might be to have `StartConfigMapRoutingTableUpdater` pass more information to the callback about what was updated, so that we can then immediately reflect the updates in the pinger. for example, we could pass the `watch.Event` to the callback. I think we can look into this in more depth after this is merged.

PR #384, opened 2022-01-25
Add TestStreamIsActive

### LINKED ISSUE #285: Add tests for StreamIsActive
The scaler's current `StreamIsActive` method does not have a test. It should. This should be done as part of, or after #97, although that's not required. Any tests written could be later retro-fitted to work with the improvement in #97  

### Use-Case

Adds reliability and stability long term to the scaler.

### Specification

- At least one unit test exists that calls and tests the output of `StreamIsActive`

asw101:
This PR adds a test for StreamIsActive. We also change the behavior of StreamIsActive to return an error when no host is found rather than logging and continuing.
  
### Checklist

- [] Any necessary documentation is added, such as:
  - [`README.md`](/README.md)

Fixes #285

### DIFF
--- a/scaler/handlers_test.go
+++ b/scaler/handlers_test.go
@@ -3,7 +3,6 @@ package main
 import (
 	context "context"
 	"fmt"
-	"log"
 	"net"
 	"testing"
 	"time"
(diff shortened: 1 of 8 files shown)


scaler/handlers_test.go
```diff
@@ -88,9 +93,140 @@ func TestIsActive(t *testing.T) {
...
+				lggr,
+				pinger,
+				table,
+				123,
+				200,
+			)
+
+			bufSize := 1024 * 1024
+			lis := bufconn.Listen(bufSize)
+			grpcServer := grpc.NewServer()
```
arschles, line 110:
```suggestion
			grpcServer := grpc.NewServer()
			defer grpcServer.Stop()
```

if you do this, then likely need to remove the `log.Fatalf` below if the `grpcServer.Serve(lis)` function returns a non-nil error

arschles (changes requested):
So far so good! I think we still need to add some tests for error & edge cases for `StreamIsActive`

arschles (approved):
Looks great, thanks @asw101! We'll do the edge case/error checking tests in a follow-up PR

PR #386, opened 2022-01-27
calling fetchAndSaveCounts in the scaler after any interceptor deployment events

### LINKED ISSUE #284: Have the scaler call requestCounts immediately after the interceptor Deployment changes # of replicas
If the interceptor deployment scales up or down, that means there is a new set of endpoints that the scaler should ping for metrics. Currently, we just wait until the next scaler ping duration (currently hard coded to 500ms) to get the new metrics from the new complete set of interceptors. This change would potentially reduce the time from interceptor scale to capturing the new metrics, thus reducing the total time from change in request volume to scaling event.

### Use-Case

Reducing the time from interceptor scale to capturing the new metrics, thus reducing the total time from change in request volume to scaling event.

### Specification

When the interceptor `Deployment` changes # of replicas, the scaler should _immediately_ call `requestCounts`.

>Note: It would be very useful to use `client-go`'s cache & informer frameworks to do this work.

arschles:
This pull request adds functionality to the scaler that watches the interceptor `Deployment` and, when any events come into the scaler for that `Deployment`, the scaler will immediately call `fetchAndSaveCounts` to get updated information on queue counts from each interceptor in the newly changed `Deployment`

### Checklist

  - [`README.md`](/README.md)

Fixes #284


bketelsen (approved):
LGTM

PR #398, opened 2022-02-14
Fix scaler crash

### LINKED ISSUE #397: Scaler component restart while multiple HttpScaledObjects
### Report

When multiple services (usually >5) are deployed with HttpScaledObjects, scaling stops working properly. This looks to be caused by scaler component restarts.

### Expected Behavior

Scaling should be working with any number of depoyments/HttpScaledObjects.

### Actual Behavior

Scaling takes long time to happen or doesn't happen at all(issue is worse with higher number of deployments)

1. Prepare deployment of service which is using scaling provided by http-add-on
2. Deploy same service multiple times to k8s cluster
3. Try to connect to one of services

Log from scaler component

[keda-add-ons-http-external-scaler.log](https://github.com/kedacore/http-add-on/files/8057911/keda-add-ons-http-external-scaler.log)

### What version of the KEDA HTTP Addon are you running?

0.3.0

### Kubernetes Version

1.21

### Platform

Amazon Web Services

### Anything else?

_No response_

leska-j:
Fixes #397
by adding write lock to counts map, there is second alternative to this [PR](https://github.com/kedacore/http-add-on/pull/399), please chose one which is better

### DIFF
--- a/scaler/queue_pinger.go
+++ b/scaler/queue_pinger.go
@@ -15,6 +15,25 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
+type SafeCount struct {
+	counts    map[string]int
+	modifyMut *sync.RWMutex
+}
+
+func newSafeCount() *SafeCount {
+	return &SafeCount{
+		counts:    map[string]int{},
+		modifyMut: new(sync.RWMutex),
+	}
+}
+
+func (s *SafeCount) update(counts map[string]int) *SafeCount {
+	s.modifyMut.Lock()
+	defer s.modifyMut.Unlock()
+	s.counts = counts
+	return s
+}
+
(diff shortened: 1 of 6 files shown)


scaler/queue_pinger.go
```diff
@@ -40,7 +59,7 @@ type queuePinger struct {
 	adminPort          string
 	pingMut            *sync.RWMutex
```
arschles, line 60: @leska-j do we still need this mutex, now that `SafeCount` has an internal one?

arschles (changes requested):
@leska-j thanks for this, and special thanks for submitting two options for this fix!

I'm leaving comments in this PR and #399 to help determine which is more appropriate.

PR #399, opened 2022-02-14
Fix scaler crash

### LINKED ISSUE #397: Scaler component restart while multiple HttpScaledObjects

leska-j:
Fixes #397
fixes issue by replacing merge of counts and routing table entries with lookup,
there is second alternative to this [PR](https://github.com/kedacore/http-add-on/pull/398), please chose one which is better

Closes #398


scaler/main.go
```diff
@@ -184,7 +186,7 @@ func startAdminServer(
 	})
 	mux.HandleFunc("/queue", func(w http.ResponseWriter, r *http.Request) {
 		lggr = lggr.WithName("route.counts")
-		cts := pinger.counts()
+		cts := mergeCountsWithRoutingTable(pinger.counts(), routingTable)
```
arschles, line 189: this endpoint only needs to return the current state of the queue, not the state of the queue merged with the routing table. I think we can keep it as-is.

arschles (changes requested):
@leska-j thanks for this one too. I think it looks like a better option than #398. Do you have a preference? Also, I left one request for a change here.

scaler/main.go
leska-j: does this apply to the /queue_ping as well?

leska-j:
> @leska-j thanks for this one too. I think it looks like a better option than #398. Do you have a preference? Also, I left one request for a change here.

Hi, i prefere this one as well, I provided second one just in case that merged count are required by some planned functionality

scaler/main.go
arschles: >does this apply to the "/queue_ping" as well?

yes

>for now I removed it from there as well

thank you :)

arschles (approved):
@leska-j this looks great now. There is [one test failing](https://github.com/kedacore/http-add-on/runs/5205538452?check_suite_focus=true#step:4:86) still, but once you fix that we can merge this.

leska-j:
@arschles thanks for review & merge

arschles:
you're welcome @leska-j. feel free to contribute further, if you would like.

PR #404, opened 2022-02-18
Ensuring that the proper host is set on all Location headers

### LINKED ISSUE #331: Proxy redirection to the wrong URL
Interceptor redirect to `http://service:80/path` instead of `http://url.com/path`

## Expected Behavior

Redirect to `http://url.com/path`

## Actual Behavior

Redirect from `http://url.com`to `http://service:80/path`

  1. Create service and a deployment with HTTPScaledObject (in my case it was WordPress)
  2. Create an Ingress
  3. Go to the url (`http://example.com`) and it will redirect to `http://service:80`

## Specifications

- **KEDA Version:** v2.4.0
- **KEDA HTTP Add-on Version:** v0.2.0
- **Platform & Version:** M1 version v1.22.4
- **Kubernetes Version:** v1.21.5-gke.1802

A way to fix it would be to specify the correct host on the forwardRequest function instead of giving the service host. I don't really get why you don't use the host instead of the service url? ([here](https://github.com/kedacore/http-add-on/blob/main/interceptor/request_forwarder.go#L19))

We are basically setting fake host, so that's why WordPress is not able to redirect correctly. As a workaround, we use `HTTP_X_FORWARDED_HOST` header to get the correct host.

arschles:
This PR changes the behavior of the interceptor when the app returns a `Location` header that does not have a host on it (i.e. `Location: "/abc"`).

Prior to this patch, the interceptor simply passed this value back to the client, but now, it ensures that the _incoming_ host value is prepended to the final header. This new behavior ensures that in all circumstances where an app intends to return a `Location` header, a `Location` header will be returned to the client with the proper hostname in it.

### Checklist

  - [`README.md`](/README.md)

Fixes #331


benjaminwood:
I'll try it out and follow up!

arschles:
@benjaminwood thank you kindly!

obito:
Any follow up on this ?

abatilo (approved):

proligde:
@arschles Hi! I just stumbled across this issue and instead of opening a new one, it might make sense to follow up on this one first: I think I have a scenario where the original host header should have been kept while forwarding the request to the target pod:

I have a web service that is hostname-aware and changes behavior based on it. Unfortunately it cannot be configured to use the x-forwarded-for header, which indeed contains the correct, original request hostname. 

For example: My original request was https://foo.bar.com. Without KEDA (Only with nginx ingress) my target pod sees "foo.bar.com" as the hostname (no x-forwarded-for header is set). After enabling KEDA with the http-add-on the target pod only sees "foo.default".

Is there a way to configure that behavior so the proxy preserves the original hostname? Thanks in advance - max

PR #610, opened 2023-02-23
controller: use kedav1alpha1.ScaledObject default values

### LINKED ISSUE #607: HTTP Add-On uses wrong default value for maxReplicaCount
### Report

When creating a HTTPScaledObject, it's optional to set `.spec.replicas.max`. This should result on the creation of a ScaledObject with `.spec.maxReplicaCount` unset, causing the ScaledObject controller use the default value of 100 for the field. But the add-on is creating a ScaledObject with `.spec.maxReplicaCount` set to 0, leading to an unexpected behaviour to users going through Keda's documentation.

### Expected Behavior

The add-on shouldn't set `.spec.maxReplicaCount` to 0 unless explicitly defined by the user.

### Actual Behavior

The add-on is setting `.spec.maxReplicaCount` to 0 by default.

1. Create a HTTPScaledObject without specifying `.spec.replicas.max`.
2. Check that the ScaledObject created by the add-on has `.spec.maxReplicaCount` set to 0 instead of being undefined.

N/A

0.4.0

### Kubernetes Version

1.24

### Platform

Microsoft Azure

### Anything else?

The problem originates on the struct definition of HTTPScaledObject for `.spec.replicas`. Its members are not pointers, so they will have Golang Zero Values (https://go.dev/tour/basics/12) by default. This is not a problem on most fields, but `.spec.maxReplicaCount`, just like some other fields (https://keda.sh/docs/2.0/concepts/scaling-deployments/#scaledobject-spec), has a non-zero default value, making the use of non-pointer attributes a problem.

https://github.com/kedacore/http-add-on/blob/ab7ce0941f1dd569205a2f9e0674a6f0308e78f5/operator/api/v1alpha1/httpscaledobject_types.go#L82-L87

t0rr3sp3dr0:
Make the HTTPScaledObject controller use nil instead of Golang Zero Values for optional fields of ScaledObjects.

### Checklist

  - [`README.md`](/README.md)

Fixes #607


t0rr3sp3dr0:
@JorTurFer @tomkerkhove, could you take a look at this PR and allow the workflows to run?

tomkerkhove (commented):
LGTM, but I think it might be best if @JorTurFer reviews as well

JorTurFer (approved):
LGTM!
Thanks for the contribution ❤️

go.mod
```diff
@@ -6,6 +6,7 @@ require (
 	github.com/codeskyblue/go-sh v0.0.0-20200712050446-30169cf553fe
 	github.com/go-logr/logr v1.2.3
 	github.com/go-logr/zapr v1.2.3
+	github.com/kedacore/keda/v2 v2.9.3
```
JorTurFer, line 9: Does registering versions relations make sense @tomkerkhove ?
I guess that this dep works also with newer code because we haven't introduced any change here, but maybe having a table relating HTTP Add-on version with KEDA versions could be nice. 
At least with KEDA main and KEDA -1.
WDYT @tomkerkhove ?
We should introduce multiple KEDA versions for the e2e tests too (just editing a bit the execution script to allow use arguments for setting KEDA version)

tomkerkhove: I don't have a preference here

JorTurFer: I'd create a compatibility table in docs to list these relations something like:
|HTTP-AddOn|KEDA|Kubernetes|
|---------------|-----|-------------|
| 0.4.0|v2.x-v2.z|1.xx-1.zz|

At least with known version, like we do in [core docs](https://keda.sh/docs/latest/operate/cluster/), basically I'd like to avoid issue about incompatibilities, or at least a documented section to point them.
Could you add something like this to docs (maybe at the end of [install section](https://github.com/kedacore/http-add-on/blob/main/docs/install.md))

t0rr3sp3dr0: Done!

JorTurFer (approved):

PR #613, opened 2023-02-24
rbac: narrow permissions

### LINKED ISSUE #612: Permissions given to the add-on components are too broad
### Report

Part of #611. The current role of the HTTP Add-On is too powerful and gives read-write access to a lot of different resources in all namespaces. This same role is given to all components of the add-on even though this is not necessary.

### Expected Behavior

- K8s permissions given to the add-on should be on as-needed basis.
- The permissions should be scoped based on each component needs.

### Actual Behavior

- Permissions give cluster-wide read-write access to resources like ConfigMaps, Deployments, Endpoints, Pods, and Services.
- All add-on components share the same ServiceAccount, thus the same permission set.

N/A

N/A

0.4.0

### Kubernetes Version

1.24

### Platform

Microsoft Azure

### Anything else?

_No response_

t0rr3sp3dr0:
Create new Roles and ServiceAccounts for the Interceptor and for the Scaler with granular permissions given to them. The Operator Role was changed to narrow its scope and remove write access to resources not owned by it.

### Checklist

  - [`README.md`](/README.md)

Fixes #612


t0rr3sp3dr0:
@JorTurFer @tomkerkhove, could you take a look at this PR and allow the workflows to run?

Changelog.md
```diff
@@ -25,7 +25,7 @@ This changelog keeps track of work items that have been completed and are ready
 
 ### Fixes
 
-- **General**: TODO ([#TODO](https://github.com/kedacore/http-add-on/issues/TODO))
+- **RBAC**: Narrow permissions given to components ([#612](https://github.com/kedacore/http-add-on/issues/612))
```
tomkerkhove, line 28: This might be more customer-friendly so they understand what has changed?
```suggestion
- **RBAC**: Introduce fine-grained permissions per component and reduce required permissions ([#612](https://github.com/kedacore/http-add-on/issues/612))
```

interceptor/main.go
```diff
@@ -90,6 +90,7 @@ func main() {
 		lggr,
 		cl,
 		servingCfg.ConfigMapCacheRsyncPeriod,
+		servingCfg.CurrentNamespace,
```
tomkerkhove, line 93: Can you elaborate on this please?

tomkerkhove (approved):

interceptor/main.go
t0rr3sp3dr0: Sure! When I changed the permissions so the add-on components only had access to ConfigMaps on the `keda` namespace, it started to fail with the following error:

```log
W0224 02:33:59.494889       1 reflector.go:424] pkg/mod/k8s.io/client-go@v0.26.1/tools/cache/reflector.go:169: failed to list *v1.ConfigMap: configmaps is forbidden: User "system:serviceaccount:keda:keda-http-add-on-interceptor" cannot list resource "configmaps" in API group "" at the cluster scope
```

After investigating, I found that the ConfigMap Informer was watching all ConfigMaps of the cluster, not just the ones on the `keda` namespace. Then I changed it to only look at ConfigMaps from the `keda` namespace (https://github.com/kedacore/http-add-on/pull/613/files#diff-a2bfc341efdd78af356e3724892bfd7ef8c5cfb3cc9cf14b209df74db39249b8L129-L132). But a user may want to install KEDA on another namespace and this is parameterised, so I need to tell the informer which namespace the add-on is installed for it to watch at ConfigMaps there. That's why I had to add this extra parameter.

JorTurFer (commented):
Nice improvement!
More fine permission management is always worth :)
How did you generate the RBAC files? [We have this command](https://github.com/kedacore/http-add-on/blob/main/Makefile#L88-L89) in make file for generating current RBAC file based on the RBAC annotations. I like the idea of a better permission segmentation, but we should maintain a way to automatically generate manifests (including RBAC files). 
I guess that we can adapt it for the new scenario

t0rr3sp3dr0:
@JorTurFer, I changed the `// +kubebuilder:` comments so the RBAC manifests are automatically generated by controller-gen for the operator when we run `make manifests`. The YAMLs for the interceptor and the scaler are completely unmanaged in the current implementation, so I created the new manifests manually.

JorTurFer:
> I created the new manifests manually

I think that we should generate them automatically. It's more complicated but it's doable and we can rely on kubebuilder for generating the RBAC manifests. I can help you with that if you want, I did it in KEDA (but finally it wasn't necessary so I removed it).

Let me know if you want to do that part or you prefer that I do it (and in that case, allow maintainers editions for the PR)

t0rr3sp3dr0:
@JorTurFer, done! I've configured manifest generation for interceptor and scaler with kubebuilder.

JorTurFer (approved):
LGTM from code pov, but I' m not totally sure about having 3 different service accounts, I see the improvement of role segmentation, but I'm not sure if having 1 sa per component is overkill.
Apart from that, should we update the helm chart now or maybe as the changes include the RBAC files we can do it for the release?
WDYT about these questions @tomkerkhove ?

t0rr3sp3dr0:
I created three different ServiceAccounts to reduce the blast radius in case some component gets compromised. The interceptor receives arbitrary requests, parses their them, and then forwards to the appropriate services, so I don't like the idea of giving this component any kind of write access.

tomkerkhove:
We can update it now already, but ship once we ship newer version

JorTurFer:
@t0rr3sp3dr0 , do you know how to do it?
You need to open a PR reflecting this RBAC changes in this repo https://github.com/kedacore/charts/tree/main/http-add-on

t0rr3sp3dr0:
@JorTurFer, here is the PR with the changes to the Helm Chart: kedacore/charts#398

PR #669, opened 2023-05-09
HTTPSO-based Routing Table

### LINKED ISSUE #605: Changes on HTTPScaledObjects don't update routing table
### Report

The current implementation of the HTTP Add-on allows users to modify all attributes of the HTTPScaledObject, but the internal routing table implementation also assumes some of these fields are immutable. Modifying these fields causes unexpected behaviour, the routing table is never updated but the K8s objects are. I experienced that after changing the value of `.spec.targetPendingRequests` and having the scaler not behave as expected.

### Expected Behavior

Either `.spec.scaleTargetRef` and `.spec.targetPendingRequests` should be immutable, or we should update the routing table when these fields get updated on the K8s object.

### Actual Behavior

Changes to `.spec.scaleTargetRef` and `.spec.targetPendingRequests` are successfully applied to K8s, but the internal routing table is not updated to reflect these new values.

1. Install Keda and the HTTP Add-on to a brand new K8s cluster and also deploy some example HTTP application to it
2. Check the routing table of the HTTP Add-on is empty (http://keda-add-ons-http-interceptor-admin.svc.cluster.local:9090/routing_table)
3. Create a valid HTTPScaledObject pointing to the application you deployed and with `.spec.targetPendingRequests` set to 2
4. Check the routing table was updated with the details of the HTTPScaledObject created (http://keda-add-ons-http-interceptor-admin.svc.cluster.local:9090/routing_table)
5. Edit the HTTPScaledObject and set `.spec.targetPendingRequests` set to 1
6. Check the routing table was NOT updated and `TargetPendingRequests` is still 2, not 1 (http://keda-add-ons-http-interceptor-admin.svc.cluster.local:9090/routing_table)

```text
2023-02-17T22:14:46Z	INFO	controllers.HTTPScaledObject	Reconciliation start	{"HTTPScaledObject.Namespace": "default", "HTTPScaledObject.Name": "awesome-project"}

2023-02-17T22:14:46Z	INFO	controllers.HTTPScaledObject	Reconciling HTTPScaledObject	{"HTTPScaledObject.Namespace": "default", "HTTPScaledObject.Name": "awesome-project", "Namespace": "default", "DeploymentName": "awesome-project"}

2023-02-17T22:14:46Z	INFO	controllers.HTTPScaledObject	Creating scaled objects{"HTTPScaledObject.Namespace": "default", "HTTPScaledObject.Name": "awesome-project", "reconciler.appObjects": "addObjects", "HTTPScaledObject.name": "awesome-project", "HTTPScaledObject.namespace": "default", "external scaler host name": "keda-add-ons-http-external-scaler.keda:9090"}

2023-02-17T22:14:46Z	INFO	controllers.HTTPScaledObject	Creating App ScaledObject	{"HTTPScaledObject.Namespace": "default", "HTTPScaledObject.Name": "awesome-project", "reconciler.appObjects": "addObjects", "HTTPScaledObject.name": "awesome-project", "HTTPScaledObject.namespace": "default", "ScaledObject": {"Object":{"apiVersion":"keda.sh/v1alpha1","kind":"ScaledObject","metadata":{"labels":{"app":"kedahttp-awesome-project-app","name":"awesome-project-app"},"name":"awesome-project-app","namespace":"default"},"spec":{"advanced":{"restoreToOriginalReplicaCount":true},"maxReplicaCount":100,"minReplicaCount":0,"pollingInterval":1,"scaleTargetRef":{"kind":"Deployment","name":"awesome-project"},"triggers":[{"metadata":{"host":"awesome-project.t0rr3sp3dr0.dev","scalerAddress":"keda-add-ons-http-external-scaler.keda:9090"},"type":"external-push"}]}}}}

2023-02-17T22:14:47Z	ERROR	controllers.HTTPScaledObject.addAndUpdateRoutingTable	could not add host to routing table, progressing anyway	{"HTTPScaledObject.Namespace": "default", "HTTPScaledObject.Name": "awesome-project", "reconciler.appObjects": "addObjects", "HTTPScaledObject.name": "awesome-project", "HTTPScaledObject.namespace": "default", "host": "awesome-project.t0rr3sp3dr0.dev", "error": "host awesome-project.t0rr3sp3dr0.dev is already registered in the routing table"}
github.com/kedacore/http-add-on/operator/controllers.addAndUpdateRoutingTable
	/workspace/operator/controllers/routing_table.go:46
github.com/kedacore/http-add-on/operator/controllers.createOrUpdateApplicationResources
	/workspace/operator/controllers/app.go:132
github.com/kedacore/http-add-on/operator/controllers.(*HTTPScaledObjectReconciler).Reconcile
	/workspace/operator/controllers/httpscaledobject_controller.go:123
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.1/pkg/internal/controller/controller.go:122
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.1/pkg/internal/controller/controller.go:323
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.1/pkg/internal/controller/controller.go:274
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.1/pkg/internal/controller/controller.go:235

2023-02-17T22:14:47Z	INFO	controllers.HTTPScaledObject	Updating status on HTTPScaledObject	{"HTTPScaledObject.Namespace": "default", "HTTPScaledObject.Name": "awesome-project", "resource version": "1941396"}

2023-02-17T22:14:47Z	INFO	controllers.HTTPScaledObject	Updated status on HTTPScaledObject	{"HTTPScaledObject.Namespace": "default", "HTTPScaledObject.Name": "awesome-project", "resource version": "1941399"}

2023-02-17T22:14:47Z	INFO	controllers.HTTPScaledObject	Updating status on HTTPScaledObject	{"HTTPScaledObject.Namespace": "default", "HTTPScaledObject.Name": "awesome-project", "resource version": "1941399"}

2023-02-17T22:14:47Z	INFO	controllers.HTTPScaledObject	Updated status on HTTPScaledObject	{"HTTPScaledObject.Namespace": "default", "HTTPScaledObject.Name": "awesome-project", "resource version": "1941400"}

2023-02-17T22:14:47Z	INFO	controllers.HTTPScaledObject	Reconcile success	{"HTTPScaledObject.Namespace": "default", "HTTPScaledObject.Name": "awesome-project"}
```

0.4.0

### Kubernetes Version

1.24

### Platform

Microsoft Azure

### Anything else?

I think the current behaviour is related to our lack of support for multiple HTTPScaledObject to have the same host. This implementation prevents a new HTTPScaledObject that uses the host of an existing HTTPScaledObject to override the configuration of the latter. This would also be a security issue, as users that have access to some namespace on a cluster could after resources on other namespaces.

I took a look at the implementation details of the routing table and propose we move from having it stored in a ConfigMap to compute its state based on the HTTPScaledObject present on the cluster and have the built routing table cached in-memory, having it refreshed periodically. Another option is to keep the current ConfigMap-based implementation, but it has some drawbacks that I'll list bellow.

#### Proposal 1 - HTTPScaledObject-based Routing Table
Rework the routing table implementation by relying on HTTPScaledObjects. Interceptors would compute the routing table based on these objects and cache the result in-memory. The cache would be updated based on K8s events and a full reconciliation performed periodically.

##### Benefits
- Have the K8s API Server as the single source of truth
- Internal routing table state isn't stored on user-writable storage and is not affected by external factors
- Interceptors no longer have to rely on the controller to have accurate routing table information
- Single place to have input validations performed

##### Drawbacks
- Interceptors become watchers of all HTTPScaledObjects
- Increased number of API calls to the K8s API Server (distributed routing table computation on interceptors)

#### Proposal 2 - ConfigMap-based Routing Table
Fix the current routing table implementation based on ConfigMap to have it updated when HTTPScaledObjects change.

##### Benefits
- Reuse existing code/architecture
- Reduced number of API calls to the K8s API Server (centralised routing table computation on controller)

##### Drawbacks
- Makes harder for users to debug if something goes wrong (users need to know the source of truth is the ConfigMap, not the HTTPScaledObject instances)
- Need to maintain the ConfigMap structure in sync with the HTTPScaledObject definition (developers need to know they need to make changes in both places for everything to work as expected)
- The ConfigMap is basically an aggregate view of all HTTPScaledObjects that can get out-of-sync due to business logic bugs or outage of the controller-manager
- Additional work to make ConfigMap solution more resilient (the HTTP Add-on is currently unable to self-heal in case the ConfigMap gets deleted or corrupted)

#### Proposal 3 - Make fields immutable
Update the CRD to make immutable the fields the routing table rely on.

##### Benefits
- Quick fix

##### Drawbacks
- Degraded user experience (users would have to delete HTTPScaledObject and create a new one HTTPScaledObject to update the settings)
- Unable to update configuration without downtime (deletion of an HTTPScaledObject make the service unreachable)

t0rr3sp3dr0:
Reimplements routing table of the HTTP Add-On to make it using HTTPSO objects as the source of truth instead of relying on a ConfigMap that is updated periodicaly by the operator.

Highlights:
- Remove all ConfigMap-related routines and permissions.
- Use a SharedInformer to interface with the API Server. HTTPSO objects are cached on memory and events are emitted when an object change. It's possible to pull objects from this cached manually instead of relying on events.
- Routing table indexes routes using a [Radix Tree](https://en.wikipedia.org/wiki/Radix_tree), similar to what one of the fastest Golang HTTP Routers do: https://github.com/julienschmidt/httprouter#how-does-it-work. This also allows path-based routing to be added in the near future.
- No user-facing interface was broken, the change is transparent for the end-user.

### Checklist

  - [`README.md`](/README.md)

Fixes #605


t0rr3sp3dr0:
@JorTurFer @tomkerkhove, sorry for the delay! Here we have the new Routing Table implementation.

I'll be adding new tests and fixing existing ones in the following days. I would appreciate reviews in the meantime.

I left the support for path-based routing out of this PR because it alredy changes a lot of stuff, but I architected the new RT with that feature in mind. I'll be working on that next.

config/interceptor/deployment.yaml
```diff
@@ -70,6 +70,7 @@ spec:
           allowPrivilegeEscalation: false
           readOnlyRootFilesystem: true
           runAsNonRoot: true
+          runAsUser: 65532
```
JorTurFer, line 73: I'm not totally sure if we should set this value here. I remember problems related with userIds in OpenShift. If I'm not wrong, OpenShift generate random used id ranges for each workload. Being honest, this is not a problem IMO because users can use helm chart to deploy the add-on, but just in case...
Any thought @tomkerkhove ?

config/operator/deployment.yaml
```diff
@@ -58,6 +58,7 @@ spec:
           allowPrivilegeEscalation: false
           readOnlyRootFilesystem: true
           runAsNonRoot: true
+          runAsUser: 65532
```
JorTurFer, line 61: same as above

config/scaler/deployment.yaml
```diff
@@ -57,6 +57,7 @@ spec:
           allowPrivilegeEscalation: false
           readOnlyRootFilesystem: true
           runAsNonRoot: true
+          runAsUser: 65532
```
JorTurFer, line 60: same as above

pkg/routing/table.go
```diff
@@ -1,135 +1,211 @@
...
+	Route(req *http.Request) *httpv1alpha1.HTTPScaledObject
+	HasSynced() bool
 }
-type Table struct {
-	fmt.Stringer
-	m map[string]Target
-	l *sync.RWMutex
+
+type table struct {
+	// TODO(pedrotorres): remove after upgrading k8s.io/client-go to v0.27.0
```
JorTurFer, line 27: Are you waiting for KEDA using that dep? If yes, [this PR](https://github.com/kedacore/keda/pull/4473) could be interesting for you and will be merged soon

interceptor/main.go
```diff
@@ -77,35 +79,20 @@ func main() {
...
-		routingTable,
-		q,
-	); err != nil {
+	httpCl, err := clientset.NewForConfig(cfg)
+	if err != nil {
+		lggr.Error(err, "creating new HTTP ClientSet")
+		os.Exit(1)
+	}
+	sharedInformerFactory := informers.NewSharedInformerFactory(httpCl, servingCfg.ConfigMapCacheRsyncPeriod)
+	routingTable, err := routing.NewTable(sharedInformerFactory, "")
```
JorTurFer, line 94: Should we use an env var for the namespace? I mean, maybe some user wants to limit the interceptor to a specific namespace (or maybe not, in this stage we can check all the namespaces)

scaler/main.go
```diff
@@ -50,9 +55,15 @@ func main() {
 	targetPendingRequests := cfg.TargetPendingRequests
 	targetPendingRequestsInterceptor := cfg.TargetPendingRequestsInterceptor
 
-	k8sCl, _, err := k8s.NewClientset()
+	k8sCfg, err := rest.InClusterConfig()
```
JorTurFer, line 58: I know that other components don't use it, but I think that it's better if we do
```suggestion
        // import ctrl "sigs.k8s.io/controller-runtime"
	k8sCfg := ctrl.GetConfigOrDie()
```

Thanks to this change, we can debug the code locally without any change. (and maybe we should modify the other places to use this approach)

JorTurFer (commented):
Looking fantastic!
Some comments inline

config/interceptor/deployment.yaml
t0rr3sp3dr0: I set this because the pod was refusing to go up on AKS. I can get the specific error for you latter, but the problem is that on the Dockerfile we are setting the user to this value and on the K8s manifest we are setting `runAsNonRoot: true`, this combination mandates that we also specify `runAsUser` here.

pkg/routing/table.go
t0rr3sp3dr0: Yes! When you PR gets merged and we update the KEDA version here, we can make the modifications I’ve commented on the code.

t0rr3sp3dr0: Yes! When you PR gets merged and we update the KEDA version here, we can make the modifications I’ve commented on the code.

interceptor/main.go
t0rr3sp3dr0: That was something I thought when writing the code but I wasn’t sure if we were supporting namespaces installs of the add-on. I’ll check the state of that and make a patch here if we should already support that. Otherwise, I think we can implement that on another PR.

scaler/main.go
t0rr3sp3dr0: Great suggestion! I’ll change that.

similark:
Hi @t0rr3sp3dr0 
Nice work!
I actually started working on integrating that very same [httpRouter](https://github.com/julienschmidt/httprouter#how-does-it-work) lib in order to add support for path-based routing.
Now, assuming this pr will be merged (hopefully 🙏 ) - I'll be very happy to help in whatever way to achieve it using your implementation.
Please let me know what's the status on that one?

Thanks and well done!

t0rr3sp3dr0:
@similark, it is WIP and I should have a PR by Monday. On this very first version we will only allow one path per HTTPScaledObject, every URL that matches such prefix will be routed to the target specified on the manifest. Is that enough for your use case?

The 2nd version will allow multiple paths to be directed to a single target. But this would be better implemented by having a new API Version of HTTPScaledObject and that needs to be discussed an may take longer to be implemented.

config/interceptor/deployment.yaml
t0rr3sp3dr0: For future reference, the issue regarding runAsUser is kedacore/keda#3012 and it was fixed by kedacore/keda#3015.

In core we have two different security contexts, one for the container and another one for the pod. Replicating the same structure here, the problem I was experiencing is gone and we don't need to set runAsUser anymore.

similark:
@t0rr3sp3dr0 Yes! the 1st version answers our use case perfectly as we have nginx-ingress routing traffic into the cluster and we must support multiple applications serving the same host header but different `path`s. Thanks again, I went over this and it looks much better overall, solving other issues we've encountered.

t0rr3sp3dr0:
@similark, I'll do that next time I push something here. But for now, I'm just writing some tests. I'm satisfied with the implementation and should only change that if someone suggests something on the review. Or in case @JorTurFer's PR gets merged on kedacore/keda in time for this patch.

pkg/routing/table.go
t0rr3sp3dr0: @JorTurFer, given your PR was merged, should we wait for KEDA 2.10.2 or 2.11.0 before merging this patch?

JorTurFer: TBH, I wouldn't wait. This project is in beta and we also have [the compatibility table](https://github.com/kedacore/http-add-on/blob/main/docs/install.md#compatibility-table):
image
If this breaks something with other KEDA versions, we can just limit the compatibility

JorTurFer:
> Or in case @JorTurFer's PR gets merged on kedacore/keda in time for this patch.

Merged 😄

pkg/routing/table.go
t0rr3sp3dr0: Are you suggesting that I update the dependency of this project to the master version of KEDA? Or that I uncomment the code that depends on k8s.io/client-go@v0.27.0 later, on another PR?

JorTurFer: I'd use main branch from KEDA in order to release the feature, and once we release KEDA v2.11 we can update it again to fix the version

t0rr3sp3dr0: I'll do that

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ This changelog keeps track of work items that have been completed and are ready
 - **General**: Automatically tag Docker image with commit SHA ([#567](https://github.com/kedacore/http-add-on/issues/567))
 - **RBAC**: Introduce fine-grained permissions per component and reduce required permissions ([#612](https://github.com/kedacore/http-add-on/issues/612))
 - **Operator**: Migrate project to Kubebuilder v3 ([#625](https://github.com/kedacore/http-add-on/issues/625))
+- **Routing**: New HTTPSO-based Routing Table ([#605](https://github.com/kedacore/http-add-on/issues/605))
```
tomkerkhove, line 27: Let's be a bit more concrete here for our end-users - What does this exactly mean for them?

tomkerkhove (commented):
Did a skim, but will mainly leave it up to @JorTurFer

scaler/main.go
```diff
@@ -6,54 +6,58 @@ package main
...
 	targetPortStr := fmt.Sprintf("%d", cfg.TargetPort)
 	targetPendingRequests := cfg.TargetPendingRequests
 	targetPendingRequestsInterceptor := cfg.TargetPendingRequestsInterceptor
 
-	k8sCl, _, err := k8s.NewClientset()
+	k8sCfg, err := ctrl.GetConfig()
+	if err != nil {
+		lggr.Error(err, "Kubernetes client config not found")
+		os.Exit(1)
+	}
```
JorTurFer, line 57:
```suggestion
	k8sCfg := ctrl.GetConfigOrDie()	
```

scaler/main.go
```diff
@@ -161,85 +140,44 @@ func startGrpcServer(
...
 	addr := fmt.Sprintf("0.0.0.0:%d", port)
 	lggr.Info("starting grpc server", "address", addr)
+
 	lis, err := net.Listen("tcp", addr)
 	if err != nil {
 		return err
 	}
 
 	grpcServer := grpc.NewServer()
+	reflection.Register(grpcServer)
```
JorTurFer, line 156: Does the reflection make sense? (just asking, I agree with enabling it)

t0rr3sp3dr0: I like to have it enabled because it allows the use of the gRPC CLI and other tools that require this service.

andrey-bondar:
Hi! I've just started using this addon and this feature is really crucial as well as https://github.com/kedacore/http-add-on/pull/677.
@t0rr3sp3dr0 can you predict when it could be completed?

t0rr3sp3dr0:
@andrey-bondar, this patch is currently feature-complete and I’m adding more tests while I wait for an in-depth review. The review may take some time, given the size of the PR and that we currently have a small number of maintainers, but it is being worked on. In the meantime, feel free to leave a review as well.

JorTurFer:
I have checked the PR and looks good, but I'd like to review it in depth with calm to understand correctly what is happening

pkg/routing/table.go
```diff
@@ -1,135 +1,211 @@
...
 }
-type Table struct {
-	fmt.Stringer
-	m map[string]Target
-	l *sync.RWMutex
+
+type table struct {
+	httpScaledObjectInformer                 sharedIndexInformer
+	httpScaledObjectEventHandlerRegistration cache.ResourceEventHandlerRegistration
+	httpScaledObjects                        map[types.NamespacedName]*httpv1alpha1.HTTPScaledObject
```
JorTurFer, line 39: This map use a "complex" (types.NamespacedName has 2 strings inside). 
Does it make sense if we generate the string concatenating both from our side and using `map[string]*httpv1alpha1.HTTPScaledObject`?

pkg/routing/key.go
```diff
@@ -0,0 +1,82 @@
...
+
+type Keys []Key
+
+func NewKeysFromHTTPSO(httpso *httpv1alpha1.HTTPScaledObject) Keys {
+	if httpso == nil {
+		return nil
+	}
+	spec := httpso.Spec
+
+	size := len(spec.Hosts)
```
JorTurFer, line 68: Be careful here because `Host` is still supported. In this case, if someone is using the old property, this size will be 0, is this correct?

interceptor/middleware/logging.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+func (lm *Logging) logAsync(w http.ResponseWriter, r *http.Request, sw *util.Stopwatch) func() {
+	signaler := util.NewSignaler()
+
+	go lm.log(w, r, sw, signaler)
+
+	return func() {
+		go signaler.Signal()
+	}
```
JorTurFer, line 53: Could we have perf problems creating 2 routines for each request? Can't we just avoid the signaler doing something like:
```golang
return func() {
		go lm.log(w, r, sw)
	}
```

Or even if maybe making it sync

interceptor/middleware/counting.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+func (cm *Counting) countAsync(ctx context.Context) func() {
+	signaler := util.NewSignaler()
+
+	go cm.count(ctx, signaler)
+
+	return func() {
+		go signaler.Signal()
+	}
```
JorTurFer, line 44: Could we have perf problems creating 2 routines for each request? Can't we just avoid the signaler doing something like:
```golang
return func() {
		go cm.count(ctx, sw)
	}
```

Or even if maybe making it sync

JorTurFer (commented):
You have done an awesome job @t0rr3sp3dr0, I have left some small comments inline

interceptor/middleware/logging.go
t0rr3sp3dr0: I made this async to avoid adding extra latency to the request. Logs are generally slow to write, so I don't like the idea of doing this synchronously.

I'm not sure how long does it take to invoke `go`, so I'll take a better look at this. Regarding the runtime impact of the new Go Routine, it will stay parked for the majority of its life, so that's negligible.

t0rr3sp3dr0: The alternative where we just use a single Go Routine would look more like:
```go
	t0 := time.Now()

	return func() {
		d := time.Since(t0)
		go lm.log(w, r, sw, d)
	}
```

That actually seems a better implementation, as we would have more accurate timing and I don't think the calls to `time.Now` and `time.Since` would impact the latency significantly.

interceptor/middleware/counting.go
t0rr3sp3dr0: For this one, I also wouldn't like to perform the operation synchronously. We are locking/unlocking a mutex and operating on a map, this takes time.

interceptor/middleware/logging.go
JorTurFer: so, will you update the implementation?

t0rr3sp3dr0: Yes, I will be doing that

t0rr3sp3dr0: I reread the implementation and noticed we already have the Stopwatch instance there. So your original code suggestion was perfect, there's no need to call time.Now/Since.

t0rr3sp3dr0: This is now patched

JorTurFer (approved):

JorTurFer:
LGTM!
Thanks a lot for this awesome improvement 😄

PR #677, opened 2023-05-19
PathPrefix-based Routing

### LINKED ISSUE #338: Support routing and scaling based on path
The interceptor should be able to route requests for the same host to different services based on the request path.

### Use-Case

When using an ingress controller, you can create Ingress objects that route different requests for the same host to different backend services, depending on the _path_ portion of the request. This can be done by within a single Ingress object or across multiple of them.

When using the HTTP add-on, however, this won't work; you can, of course, still define multiple Ingress objects with different paths (or one Ingress object with multiple path entries), but they all need to point to the `keda-add-ons-http-interceptor-proxy` as their backend. Unfortunately, once the interceptor has intercepted them, it's not able to route them or scale them appropriately. The _routing_ portion of this problem could be cumbersomely worked around by having a second ingress controller that the proxy could just route all traffic too (actually you could just make the proxy a service of type LoadBalancer and _only_ have an ingress controller behind the proxy), but that still means that the scaling won't be handled as intended.

Consider the following setup (pre-KEDA):

```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: example-media
  annotations:
    kubernetes.io/ingress.class: nginx
spec:
  rules:
    - host: api.example.com
      http:
        paths:
          - path: /media
            pathType: Prefix
            backend:
              service:
                name: media-provider
                port:
                  number: 80
---
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: example-customers
  annotations:
    kubernetes.io/ingress.class: nginx
spec:
  rules:
    - host: api.example.com
      http:
        paths:
          - path: /customer
            pathType: Prefix
            backend:
              service:
                name: customer-provider
                port:
                  number: 80
```
This will route a request for e.g. `https://api.example.com/media/foo` to the `media-provider` service, and a request to `https://api.example.com/customer/bar?baz=1` to the `customer-provider` service.

Once you put the KEDA HTTP Add-on in place, things begin to break down:

```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: example-media
  annotations:
    kubernetes.io/ingress.class: nginx
spec:
  rules:
    - host: api.example.com
      http:
        paths:
          - path: /media
            pathType: Prefix
            backend:
              service:
                name: keda-add-ons-http-interceptor-proxy
                port:
                  number: 80
---
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: example-customers
  annotations:
    kubernetes.io/ingress.class: nginx
spec:
  rules:
    - host: api.example.com
      http:
        paths:
          - path: /customer
            pathType: Prefix
            backend:
              service:
                name: keda-add-ons-http-interceptor-proxy
                port:
                  number: 80
---
kind: HTTPScaledObject
metadata:
  name: media-api
spec:
  host: "api.example.com"
  scaleTargetRef:
    deployment: media-provider
    service: media-provider
    port: 80
    targetPendingRequests: 100
---
kind: HTTPScaledObject
metadata:
  name: customer-api
spec:
  host: "api.example.com"
  scaleTargetRef:
    deployment: customer-provider
    service: customer-provider
    port: 80
    targetPendingRequests: 100
```
Both `HTTPScaledObject` specs have the same `host` field, and there's no `path` field, so there's no way for the scaler to know where to route traffic to. Even if you put the Ingress Controller "behind" the interceptor (once a cluster-global installation is possible) to handle the routing, there's still no way for the scaler to know that when a bunch of requests come in for a `/customer` endpoint, it should scale the `customer-provider` Deployment rather than the `media-provider` Deployment.

To resolve this problem, the `HTTPScaledObject` field should have an optional `spec.path` field, and _possibly_ also an optional `spec.pathType` field. For example:

```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: example
  annotations:
    kubernetes.io/ingress.class: nginx
spec:
  rules:
    - host: api.example.com
      http:
        paths:
          - path: /media
            pathType: Prefix
            backend:
              service:
                name: keda-add-ons-http-interceptor-proxy
                port:
                  number: 80
          - path: /customer
            pathType: Prefix
            backend:
              service:
                name: keda-add-ons-http-interceptor-proxy
                port:
                  number: 80
---
kind: HTTPScaledObject
metadata:
  name: media-api
  path: /media
spec:
  host: "api.example.com"
  scaleTargetRef:
    deployment: media-provider
    service: media-provider
    port: 80
    targetPendingRequests: 100
---
kind: HTTPScaledObject
metadata:
  name: customer-api
spec:
  host: "api.example.com"
  path: /customer
  scaleTargetRef:
    deployment: customer-provider
    service: customer-provider
    port: 80
    targetPendingRequests: 100
```

### Specification

For simplicity, I think you could get away with supporting only either 0 or 1 paths per `HTTPScaledObject` resource. Resolution would work as follows:

1. Any `HTTPScaledObject` which omits `spec.path` is equivalent to one which specifies `spec.path: /`
2. If `spec.pathType` is implemented at all, it should have the same basic semantics as `spec.rules[*].http.paths[*].pathType` in `Ingress`, except that the only supported values would be `Prefix` and `Exact` (no need to include historical baggage like `ImplementationSpecific` since there's only one implementation and there's also no history to derive baggage from)
3. If `spec.pathType` is implemented at all, omitting `spec.pathType` will be equivalent to `spec.pathType: Prefix`
4. When evaluating a request, the interceptor should collect all `HTTPScaledObject`s with a matching `host` and evaluate them as follows:
    1. If `spec.pathType` is implemented, check all matched `HTTPScaledObject` resources with `spec.pathType: Exact` and see if the `spec.path` listed in the `HTTPScaledObject` exactly matches the path part of the request URI. If so, the first match found is selected.
    2. If `spec.pathType` isn't implemented, or if none of the `Exact` type `HTTPScaledObject`s matched the path, take the remaining `HTTPScaledObject`s matching the `host` field and sort them by the length of the `spec.path` field. Check each resource's `spec.path` field to see if it is an "element-wise prefix" of the path portion of the request URI, given a separator of "/"; the first match found is selected.
5. Whichever single `HTTPScaledObject` resource was selected in step 4, route the request to the service specified in that `HTTPScaledObject`, and scale the deployment specified in that `HTTPScaledObject` accordingly.

This procedure defined in step 4 is essentially the same as that defined for matching requests to Ingress objects by the `networking.k8s.io` API, except for the absence of the `ImplementationSpecific` path type.

With Ingress, users have the option to specify multiple `path`s for the same host (or even multiple hosts) in a single Ingress resource, but, I think that would be unnecessary complexity for `HTTPScaledObject`; simply creating a distinct `HTTPScaledObject` per path/prefix should suffice.

t0rr3sp3dr0:
Implements routing based on a path prefix in addition to the host. A new `.spec.pathPrefix` field was created on HTTPScaledObject. When not specified, the routing behaviour is exactly the same as we had previously with hose-based routing.

Depends on: kedacore/http-add-on#669

### Checklist

  - [`README.md`](/README.md)

Fixes #338


t0rr3sp3dr0:
Tests for this feature are coming in the next few days. I'll rebase with #669 when necessary until it gets merged.

t0rr3sp3dr0:
@similark, can you try this version and see if that works as you would expect? Just add set `.spec.pathPrefix` with the value you want:

```yaml
kind: HTTPScaledObject
spec:
  pathPrefix: /foo/bar
...
```

similark:
@t0rr3sp3dr0 I actually did the exact same thing and it works great.. I'll test your version today and update here soon.
Thanks!

JorTurFer:
Let's merge [this before](https://github.com/kedacore/http-add-on/pull/686) in order to add new e2e test for this new feature

JorTurFer:
I have merged https://github.com/kedacore/http-add-on/pull/686
Please rebase your branch 🙏

tests/checks/path_prefix/path_prefix_test.go
```diff
@@ -0,0 +1,222 @@
...
+	KubectlApplyMultipleWithTemplate(t, data, templates)
+	assert.True(t, WaitForDeploymentReplicaReadyCount(t, kc, deploymentName, testNamespace, minReplicaCount, 6, 10),
+		"replica count should be %d after 1 minutes", minReplicaCount)
+
+	data.PathPrefix = pathPrefix
+
+	// not scale
+	KubectlApplyWithTemplate(t, data, "loadJobTemplate", loadJobTemplate)
+	assert.True(t, AssertDeploymentReplicaReadyCountForDuration(t, kc, deploymentName, testNamespace, minReplicaCount, 3*time.Minute),
+		"replica count should be %d for 3 minutes", minReplicaCount)
```
JorTurFer, line 199: I guess that one minute could be enough for checking that it doesn't scale and we don't increase too much the e2e test duration
```suggestion
	KubectlApplyWithTemplate(t, data, "loadJobTemplate", loadJobTemplate)
	assert.True(t, AssertDeploymentReplicaReadyCountForDuration(t, kc, deploymentName, testNamespace, minReplicaCount, 1*time.Minute),
		"replica count should be %d for 1 minutes", minReplicaCount)
```
Have you set 3 for any specific reason?

tests/helper/helper.go
```diff
@@ -635,3 +635,29 @@ func WaitForPodsTerminated(t *testing.T, kc *kubernetes.Clientset, selector, nam
...
+		}
+
+		select {
+		case <-ctx.Done():
+			return true
+		default:
+			time.Sleep(time.Second)
+		}
+	}
+}
```
JorTurFer, line 663: There is already a function `AssertReplicaCountNotChangeDuringTimePeriod` which does almost the same (inherited from core when we ported the e2e test system).
Could using the other one make sense?

tests/run-all.go
```diff
@@ -24,7 +24,7 @@ import (
 
 var (
 	concurrentTests = 2
-	testsTimeout    = "5m"
+	testsTimeout    = "15m"
```
JorTurFer, line 27: If we reduce the e2e test duration with the comment above, I'd reduce this to 10m

JorTurFer (commented):
Really nice improvement! ❤️ 
Small nits inline

tests/checks/path_prefix/path_prefix_test.go
t0rr3sp3dr0: 1 minute is the timeout we define for the scale up operations on the other tests. If for some reason the deadline is too tight and there is a bug that misroutes the request, the deployment may be scaled but we did not observed it long enough to detect the problem. In that scenarios, all tests are successful but we have a problem.

I know that we would have to be very unlucky to meet all conditions necessary for this to happen, but I was trying to make the test more resilient.

What about we meet in the middle and set the value to 2 minutes?

tests/helper/helper.go
t0rr3sp3dr0: Oh, I totally missed that function. It seems I didn't look for it well enough hahaha. I'll change that!

tests/checks/path_prefix/path_prefix_test.go
JorTurFer: Agree :)

JorTurFer (approved):

tests/run-all.go
```diff
@@ -24,7 +24,7 @@ import (
 
 var (
 	concurrentTests = 2
-	testsTimeout    = "5m"
+	testsTimeout    = "10m"
```
JorTurFer, line 27:
```suggestion
	testsTimeout    = "12m"
```

similark:
Don't we need to update the [crd int the chart|https://github.com/kedacore/charts/blob/main/http-add-on/templates/crd.yaml] as well? @JorTurFer @t0rr3sp3dr0 

Magnificent work @t0rr3sp3dr0! We'll start using this very soon

JorTurFer:
Yeah, I forgot that change, let me ship another version with this fixed

PR #678, opened 2023-05-19
Don't override Host when forward-proxying - effectively preserving the transparent proxy approach

### LINKED ISSUE #331: Proxy redirection to the wrong URL

similark:
This small change solves #331 - when building the forward proxy request, there's no need to override the Host field, which trashes the Host header altogether. The request has the proper value in the URL, making it reach the target svc and still preserving the original Host. This makes it possible to treat the request properly down the road (at the application level)

### Checklist

- [X] Commits are signed with Developer Certificate of Origin (DCO)
  - [`README.md`](/README.md)

Fixes #331

### DIFF
--- a/interceptor/request_forwarder.go
+++ b/interceptor/request_forwarder.go
@@ -18,14 +18,18 @@ func forwardRequest(
 ) {
 	proxy := httputil.NewSingleHostReverseProxy(fwdSvcURL)
 	proxy.Transport = roundTripper
+	superDirector := proxy.Director
 	proxy.Director = func(req *http.Request) {
+		superDirector(req)
 		req.URL = fwdSvcURL
-		req.Host = fwdSvcURL.Host
 		req.URL.Path = r.URL.Path
 		req.URL.RawQuery = r.URL.RawQuery
 		// delete the incoming X-Forwarded-For header so the proxy
 		// puts its own in. This is also important to prevent IP spoofing
-		req.Header.Del("X-Forwarded-For ")
+		req.Header.Del("Forwarded")
+		req.Header.Del("X-Forwarded-For")
+		req.Header.Del("X-Forwarded-Host")
+		req.Header.Del("X-Forwarded-Proto")
 	}
 	proxy.ErrorHandler = func(w http.ResponseWriter, r *http.Request, err error) {
 		w.WriteHeader(502)


similark:
Opinions, anyone? @JorTurFer

JorTurFer:
is this correct @t0rr3sp3dr0 ?

similark:
Just making note that this change is covered in the `TestImmediatelySuccessfulProxy` test.. meaning the request reaches its target as expected. Only difference is that now the Host header is kept as it was, and not updated. In a running cluster, the target pod will get the request with the actual Host in place, and not `service_name.svc.NAMESPACE.....`

t0rr3sp3dr0:
I would do something like this:
```golang
	superDirector := proxy.Director
	proxy.Director = func(req *http.Request) {
		host := req.URL.Host
		superDirector(req)
		req.URL.Host = host

		// Strip client-provided forwarding headers to prevent IP spoofing.
		req.Header.Del("Forwarded")
		req.Header.Del("X-Forwarded-For")
		req.Header.Del("X-Forwarded-Host")
		req.Header.Del("X-Forwarded-Proto")
	}
```

similark:
@t0rr3sp3dr0 how about this?

similark:
Opinions, anyone? @tomkerkhove @JorTurFer @t0rr3sp3dr0 
This is kind of a show-stopper for us, as any downstream application can't figure the original request.

t0rr3sp3dr0:
@similark, I’m ok with the changes, but you need to rebase your branch. The file you patched moved to `interceptor/handler/upstream.go`.

t0rr3sp3dr0:
Also, update `CHANGELOG.md` to reflect this change.

Issue #582, opened 2023-01-25
Improve e2e test


JorTurFer:
Currently, we are executing a single bash script for e2e tests which does this:
1. install KEDA
2. deploy the add-on
3. deploy a workload with http-scaledobject
4. check resources
5. wait till workload is scaled to 0
6. add traffic
7. wait till workload is scaled to 1
8. clear resources

To improve the test suite and to add the option to increase covered cases, we should adopt an approach like [KEDA does in the main repo](https://github.com/kedacore/keda/tree/main/tests). 

This will give us the option to easily add new test cases like testing Prometheus metrics

## Action Items

- [x] Align with KEDA Core's e2e testing approach https://github.com/kedacore/http-add-on/pull/686
- [x] Increase test coverage

### New test cases to add

- [x] Prometheus metrics
- [x] More than 1 HttpScaledobject
- [x] The scaling phase (currently, we are testing only the activation phase)
- [x] Test with a single/multiple instances.
- [x] Any bug we can reproduce, should be added as test to reduce the regressions

tomkerkhove:
So what do do you have in mind to improve the e2e experience other than aligning with KEDA core?

JorTurFer:
> So what do you have in mind to improve the e2e experience other than aligning with KEDA core?

That is already a good improvement :)

I think we should add more test cases like:
- Prometheus metrics
- More than 1 HttpScaledobject
- The scaling phase (currently, we are testing only the activation phase)
- Test with a single/multiple instances.
- Any bug we can reproduce, should be added as test to reduce the regressions

tomkerkhove:
Let's list them in the description as well then!

tomkerkhove:
Done

PR #686, opened 2023-05-26
feat: use the same e2e system as in core

JorTurFer:
As the project is growing in features, we need a more sophisticated e2e tool for executing multiple e2e tests of different types. As we have already faced with that scenario in core repo, this PR ports that e2e system into this repo.

**NOTE**: As I have needed to execute the linter in the same version as this repo is, I have fixed the devcontainers too as part of this PR

### Checklist

  - [`README.md`](/README.md)

Related to https://github.com/kedacore/http-add-on/issues/582


Makefile
```diff
@@ -5,7 +5,7 @@ SHELL           = /bin/bash
 
 IMAGE_REGISTRY 	?= ghcr.io
 IMAGE_REPO     	?= kedacore
-VERSION 		?= main
+VERSION 		?= canary
```
zroubalik, line 8: ?

tests/helper/helper.go
```diff
@@ -0,0 +1,637 @@
+//go:build e2e
```
zroubalik, line 1: Just an idea, does it make a sense to use the helper from kedacore/keda to remove rendundancies?

Makefile
JorTurFer: This is because in this repo the `main` tag, but maybe I can set it during externally, you are right

tests/helper/helper.go
JorTurFer: yeah... I thought it too, but I have to add some extra checks and I'm not 100% sure about reusing it because if we change the file on core, we will break the code here without any breaking change documentation (because it isn't a breaking change in core indeed)

zroubalik: yeah, that's right. let's keep it this way for now.

zroubalik (approved):

PR #692, opened 2023-05-31
fix: Verify manifest changes correctly

JorTurFer:
### Checklist


t0rr3sp3dr0:
It seems we had more validation problems than I noticed :laughing:

JorTurFer:
> It seems we had more validation problems than I noticed :laughing: 

Yeah, I did  some mistakes when I ported the thing from core 😅😅
At least we have detected them with only a few commits :)

Issue #665, opened 2023-05-05
XKCD addon is missing arm64


kingdonb:
### Report

The http-add-on has arm64 support, but the example `arschles/xkcd` does not publish a multi-arch image.

### Expected Behavior

The example should run with amd64 or arm64 workers

### Actual Behavior

So, in a multi-arch cluster, if your pod lands on an arm64 node when it scales up you'll sit waiting for a while and then see:

```
error on backend (context marked done while waiting for deployment xkcd to reach > 0 replicas (context deadline exceeded))
```

The pod, having landed on an arm64 node (on the cluster I've prepared for OSSNA/GitOpsCon!) has the predictable error:

```
$ k logs -f xkcd-bb4b89d74-qhrnr
exec /usr/local/bin/docker-entrypoint.sh: exec format error
```

### Steps to Reproduce the Problem

I spun up an eks cluster with `eksctl create cluster --config=file=cluster.config` and the repro is in [kingdon-ci/fleet-infra](https://github.com/kingdon-ci/fleet-infra) with the cluster name `multiarch-ossna23`.

If you omit the amd64 node group then it would fail every time. I included my cluster definition below. (For me it only fails a random percent of the time, because only 50% of nodes are arm64, it depends if xkcd gets scheduled to an arm64 node or not.)

### Logs from KEDA HTTP operator

Not relevant, but here are the chart versions anyway:

```
http-addon           	keda-test                  	1       	2023-05-04 19:51:41.98488364 +0000 UTC 	deployed	keda-add-ons-http-0.4.1      	0.4.0
ingress-public       	ingress-nginx              	2       	2023-05-04 19:47:20.179020377 +0000 UTC	deployed	ingress-nginx-4.6.0          	1.7.0
keda                 	keda-test                  	1       	2023-05-04 19:50:57.338495283 +0000 UTC	deployed	keda-2.10.2                  	2.10.1
```

### What version of the KEDA HTTP Add-on are you running?

chart 0.4.1

### Kubernetes Version

1.25

### Platform

Amazon Web Services

### Anything else?

The content of the cluster is here:

```
$ cat ~/w/eks-cluster/cluster.config
---
apiVersion: eksctl.io/v1alpha5
kind: ClusterConfig

metadata:
  name: multiarch-ossna23
  region: ca-central-1

managedNodeGroups:
  - name: ng-3
    instanceType: m5.large
    desiredCapacity: 1
    iam:
      withAddonPolicies:
        autoScaler: true
[...]

PR #693, opened 2023-05-31
feat: Use `registry.k8s.io/e2e-test-images/agnhost` instead of `arschles/xkcd`

JorTurFer:
Currently, we are using `arschles/xkcd` as docker image. This image isn't under our control and it doesn't support ARM64. As we support ARM64 arch.
As there is an image in Kubernetes registry for e2e tests which matches are requirements and it's multi arch, this PR replaces them

### Checklist

Fixes https://github.com/kedacore/http-add-on/issues/665


t0rr3sp3dr0:
What about using `k8s.gcr.io/e2e-test-images/echoserver` instead? It would allow us to check much more on E2E, like that we are sending the proper URL and Headers to the application.

t0rr3sp3dr0:
There's also the new `registry.k8s.io/e2e-test-images/agnhost:2.39`. It has a lot of different routes that return more structured data: https://github.com/kubernetes/kubernetes/blob/master/test/images/agnhost/netexec/netexec.go#L65.

JorTurFer:
I didn't know about that, I'm going to take a look :heart:

JorTurFer:
PTAL @t0rr3sp3dr0

t0rr3sp3dr0 (approved):
LGTM

PR #702, opened 2023-06-15
Fix flaky E2E tests by changing replica count constants

t0rr3sp3dr0:
Before, the initial replica count was 1, the minimum replica count was 0, and the maximum replica count was 1.

The first step of each E2E test is to wait for the ready replicas of the Deployment to reach the minimum replica count. But any Deployment starts with 0 ready replicas and increases that number up to the initial replica count value as pods are created and become ready.

This was being misinterpreted as the scaler performing these operations, causing some tests to always succeed and other ones to occasionally fail.

### Checklist

  - [`README.md`](/README.md)

### DIFF
--- a/tests/checks/ingress_in_app_namespace/ingress_in_app_namespace_test.go
+++ b/tests/checks/ingress_in_app_namespace/ingress_in_app_namespace_test.go
@@ -23,7 +23,7 @@ var (
 	serviceName          = fmt.Sprintf("%s-service", testName)
 	ingressName          = fmt.Sprintf("%s-ingress", testName)
 	httpScaledObjectName = fmt.Sprintf("%s-http-so", testName)
-	ingressHost          = fmt.Sprintf("http://%s-controller.%s", IngressReleaseName, IngressNamespace)
+	ingressHost          = fmt.Sprintf("http://%s-controller.%s/", IngressReleaseName, IngressNamespace)
 	host                 = testName
 	initReplicaCount     = 0
 	minReplicaCount      = 1
@@ -141,10 +141,13 @@ spec:
   template:
     spec:
       containers:
-      - name: curl-client
-        image: curlimages/curl
+      - name: generate-request
+        image: alpine
         imagePullPolicy: Always
-        command: ["curl", "-H", "Host: {{.Host}}", "{{.IngressHost}}"]
+        command:
+		- sh
+		- -c
+		- 'apk add apache2-utils && ab -H "Host: {{.Host}}" -c 25 -n 1000000 "{{.IngressHost}}"'
       restartPolicy: Never
   activeDeadlineSeconds: 600
   backoffLimit: 5
--- a/tests/checks/ingress_in_app_namespace/ingress_in_app_namespace_test.go
+++ b/tests/checks/ingress_in_app_namespace/ingress_in_app_namespace_test.go
@@ -25,8 +25,9 @@ var (
 	httpScaledObjectName = fmt.Sprintf("%s-http-so", testName)
 	ingressHost          = fmt.Sprintf("http://%s-controller.%s", IngressReleaseName, IngressNamespace)
 	host                 = testName
-	minReplicaCount      = 0
-	maxReplicaCount      = 1
+	initReplicaCount     = 0
+	minReplicaCount      = 1
+	maxReplicaCount      = 2
 )
 
 type templateData struct {
@@ -38,6 +39,7 @@ type templateData struct {
 	IngressHost          string
 	HTTPScaledObjectName string
 	Host                 string
+	Replicas             int
 	MinReplicas          int
 	MaxReplicas          int
 }
@@ -105,7 +107,7 @@ metadata:
   labels:
     app: {{.DeploymentName}}
 spec:
-  replicas: 1
+  replicas: {{.Replicas}}
   selector:
     matchLabels:
       app: {{.DeploymentName}}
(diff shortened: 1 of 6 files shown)


tomkerkhove (approved):

JorTurFer:
The original wait for 0 is just for ensuring that KEDA has established the communication and that's why the operator scales from 1 to zero. 
Your proposal is good but it won't work atm because we don't support the hpa behavior in the HTTPScaledObject, so the HPA controller will respect the default values, not scaling in before 5 minutes. In the other direction, 1 request isn't enough for scaling out.
We test the activation phase (0-1-0) instead of the scaling phase (1-N-1) because it's done by the operator and we can modify its values. Also, because if the operator works and scales 0-1-0 is because the add-on is working, scaling from 1-N-1 too isn't required because that responsibility is in KEDA side (because the activation has already validated that the add-on has counted the request)
If we want to execute the tests in the scaling phase (1-N-1), we have to support the HPA propagation from HTTPScaledObject to ScaledObject and also we need to update e2e tests for doing more than just a single call.

About the flaky test, you are right, the path based e2e test is flaky and IDK why because I can't reproduce it, but I think that this won't fix it xD

JorTurFer:
BTW, I'm checking that test right now 😄

t0rr3sp3dr0:
Yeah, but I think we are not testing the activation phase properly.

Let's take `internal_service_test.go` for example. It performs the following actions:
1. Creates the deployment, among other resources. (Initial replica count set to 1)
2. Waits for the ready replica count of the deployment to be 0. (This is probably true at this point because the initial pods were not ready at that point, not because the add-on scaled the Deployment to 0).
3. Creates a Job to send requests to the interceptor. (This was supposed to scale the Deployment to 1, but we don't know at this point if the add-on scaled it or if the Deployment is just in its original state)
4. Waits for the ready replica count of the deployment to be 1. (Both scenarios mentioned above make this true).
5. Delete the Job.
6. Waits for the ready replica count of the deployment to be 0. (This could indicate that the add-on scaled it down after some time without requests, but this can also indicate the initial just performed its initial scale down)

t0rr3sp3dr0:
This uncertainty about what caused the replica count to go up or down is the reason why the tests on the master branch are failing.

When the test runs slower or the pods became ready faster, the initial wair for 0 ready replicas truly indicates the add-on performed the scale-down operation and the test succeeds.

But when the test runs too fast or the pods take too long to become ready, the test fails because it thinks the add-on performed a scale-up operation. But in reality, it was just the Deployment performing its initial Pod creation.

```
path_prefix_test.go:186: --- testing not scale ---
    helper.go:482: Applying template: deploymentTemplate
    helper.go:482: Applying template: serviceNameTemplate
    helper.go:482: Applying template: httpScaledObjectTemplate
    helper.go:360: Waiting for deployment replicas to hit target. Deployment - path-prefix-test-deployment, Current  - 0, Target - 0
    helper.go:482: Applying template: loadJobTemplate
    helper.go:434: Waiting for some time to ensure deployment replica count doesn't change from 0
    helper.go:441: Deployment - path-prefix-test-deployment, Current  - 1
    helper.go:444: 
        	Error Trace:	/home/runner/work/http-add-on/http-add-on/tests/helper/helper.go:444
        	            				/home/runner/work/http-add-on/http-add-on/tests/checks/path_prefix/path_prefix_test.go:197
        	            				/home/runner/work/http-add-on/http-add-on/tests/checks/path_prefix/path_prefix_test.go:153
        	Error:      	path-prefix-test-deployment replica count has changed from 0 to 1
        	Test:       	TestCheck
```

JorTurFer:
I don't agree, let me answer point by point

> 2\. Waits for the ready replica count of the deployment to be 0. (This is probably true at this point because the initial pods were not ready at that point, not because the add-on scaled the Deployment to 0).

Yes, it's KEDA who scales the workload to 0 indeed because the scaler reports 0:

image

You can verify this just executing the test and stopping it before `testScaleOut`. If you check KEDA operator logs, you will se a record like the picture above saying that KEDA has scaled the workload from 1 to 0

> 3\. Creates a Job to send requests to the interceptor. (This was supposed to scale the Deployment to 1, but we don't know at this point if the add-on scaled it or if the Deployment is just in its original state)

We aren't redeploying all the templates, just the loadJobTemplate:

https://github.com/kedacore/http-add-on/blob/d333a0e7f50376d5645730ac5a07497683ea75bd/tests/checks/internal_service/internal_service_test.go#L150-L157

If the workload scales to 1, it's because the scaler reportes 1 request:
image

To verify this, you can execute the tests and stop it just after `testScaleOut`, you will see a record like the picture above

> 6\. Waits for the ready replica count of the deployment to be 0. (This could indicate that the add-on scaled it down after some time without requests, but this can also indicate the initial just performed its initial scale down)

same as point 2

t0rr3sp3dr0:
Regarding the scale-down operation after scale-up, can't we just change `.spec.cooldownPeriod`?

JorTurFer:
Let's switch to slack, I can explain it faster 😄

t0rr3sp3dr0:
It seems the problem is the lack of owner references when creating the SO. Causing it to be deleted asynchronously when the operator got the HTTPSO deletion event, instead of performing a synchronous cascade deletion. This made Keda reset the replica count in the middle of the following test, when the SO was indeed deleted. That's because we are setting `.spec.advanced.restoreToOriginalReplicaCount`.

PR #732, opened 2023-07-14
Fix Graceful Shutdown of Interceptor and Scaler

### LINKED ISSUE #731: Graceful Shutdown not Implemented Properly
### Report

The Interceptor and the Scaler have a graceful shutdown process triggered when the context is cancelled. But I noticed neither process handles SIGINT or SIGTERM, so they die instantly.

This is not a huge problem for the Scaler, but I can't say the same for the Interceptor. When rolling out the Interceptor, it will abruptly terminate all existing connections, creating problems for clients of services behind the proxy. 

Even if we were handling the signals, the shutdown routine of both services is not implemented correctly, as I explain below. This also means that, in the current implementation, if we have an error or we cancel the context for any reason, it terminates the service in a way very similar to a SIGKILL.

For the Interceptor, we correctly invoke `http.Server#Shutdown` to close the active connections before terminating the process. The problem is that we always invoke that method with a cancelled context, making it fail instantly.

On the Scaler, we are invoking `net.Listener#Close` to shutdown the gRPC server, terminating connections on the TCP level. This is not the proper way of tearing down the server and `grpc.Server#GracefulStop` should be used instead.

### Expected Behavior

Both the Interceptor and the Scaler should shutdown gracefully, just like the Operator does.

### Actual Behavior

The services terminate connections in a bad way, resulting in error on clients connected to them.

1. `kubectl rollout restart -n keda deployment.apps/keda-http-add-on-interceptor deployment.apps/keda-http-add-on-scaler`

Given the processes die upon receiving SIGINT or SIGTERM, there is no log.

### HTTP Add-on Version

0.5.0

### Kubernetes Version

1.26

### Platform

Microsoft Azure

### Anything else?

_No response_

t0rr3sp3dr0:
Interceptor and Scaler will now gracefully shutdown on SIGINT and SIGTERM. It also patches shutdown routines to complete ongoing requests before closing connections.

### Checklist

  - [`README.md`](/README.md)

Fixes #731

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,8 +31,8 @@ This changelog keeps track of work items that have been completed and are ready 
 - **Scaler**: remplement custom interceptor metrics ([#718](https://github.com/kedacore/http-add-on/issues/718))
 - **Operator**: Remove ScaledObject `name` & `app` custom labels ([#717](https://github.com/kedacore/http-add-on/issues/717))
 - **Interceptor**: fatal error: concurrent map iteration and map write ([#726](https://github.com/kedacore/http-add-on/issues/726))
-- **Interceptor**: shutdown http servers gracefully on SIGINT and SIGTERM ([#731](https://github.com/kedacore/http-add-on/issues/731))
-- **Scaler**: shutdown grpc server gracefully on SIGINT and SIGTERM ([#731](https://github.com/kedacore/http-add-on/issues/731))
+- **Interceptor**: Provide graceful shutdown for http servers on SIGINT and SIGTERM ([#731](https://github.com/kedacore/http-add-on/issues/731))
+- **Scaler**: Provide graceful shutdown for grpc server on SIGINT and SIGTERM ([#731](https://github.com/kedacore/http-add-on/issues/731))
 
 ### Deprecations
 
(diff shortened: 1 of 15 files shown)


t0rr3sp3dr0:
@JorTurFer

CHANGELOG.md
```diff
@@ -31,6 +31,8 @@ This changelog keeps track of work items that have been completed and are ready
 - **Scaler**: remplement custom interceptor metrics ([#718](https://github.com/kedacore/http-add-on/issues/718))
 - **Operator**: Remove ScaledObject `name` & `app` custom labels ([#717](https://github.com/kedacore/http-add-on/issues/717))
 - **Interceptor**: fatal error: concurrent map iteration and map write ([#726](https://github.com/kedacore/http-add-on/issues/726))
+- **Interceptor**: shutdown http servers gracefully on SIGINT and SIGTERM ([#731](https://github.com/kedacore/http-add-on/issues/731))
+- **Scaler**: shutdown grpc server gracefully on SIGINT and SIGTERM ([#731](https://github.com/kedacore/http-add-on/issues/731))
```
tomkerkhove, line 35: Massive nit
```suggestion
- **Interceptor**: Provide graceful shutdown for http servers on SIGINT and SIGTERM ([#731](https://github.com/kedacore/http-add-on/issues/731))
- **Scaler**: Provide graceful shutdown for grpc server on SIGINT and SIGTERM ([#731](https://github.com/kedacore/http-add-on/issues/731))
```

tomkerkhove (commented):
LGTM, but I'll leave it up to the expert

JorTurFer:
LGTM! We can merge the PR when @tomkerkhove 's comments are addressed

t0rr3sp3dr0:
Done

PR #745, opened 2023-07-31
Increase scaler stream interval to 200ms

tommy351:
Currently the scaler uses a lot of memory when there're a lot of `HTTPScaledObject`s (We have 300+). Based on the heap snapshots, it seems that the `StreamIsActive` handler consumes most of memory. I adjusted the interval from 5ms to 200ms to fix this issue.

Part of heap graph, you can open heap snapshot files below for details.

截圖 2023-07-31 16 01 35

Heap snapshots:
- [Before](https://github.com/kedacore/http-add-on/files/12212783/keda-http-addon-before.gz)
- [After](https://github.com/kedacore/http-add-on/files/12212790/keda-http-addon-after.gz)

Memory usage:

Before:

截圖 2023-07-31 15 28 09拷貝

After:

截圖 2023-07-31 15 34 16

### Checklist

  - [`README.md`](/README.md)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ This changelog keeps track of work items that have been completed and are ready 
 ### Improvements
 
 - **General**: TODO ([#TODO](https://github.com/kedacore/http-add-on/issues/TODO))
+- **Scaler**: Decrease memory usage by increasing stream interval from 5ms to 200ms.
 
 ### Fixes
 
--- a/scaler/handlers.go
+++ b/scaler/handlers.go
@@ -89,7 +89,7 @@ func (e *impl) StreamIsActive(
 	// this function communicates with KEDA via the 'server' parameter.
 	// we call server.Send (below) every 200ms, which tells it to immediately
 	// ping our IsActive RPC
-	ticker := time.NewTicker(5 * time.Millisecond)
+	ticker := time.NewTicker(200 * time.Millisecond)
 	defer ticker.Stop()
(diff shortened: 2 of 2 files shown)


JorTurFer (commented):
This is a nice improvement! The resources saving is awesome, but I'm afraid about if this can impact in current applications, I don't think so, but IDK if it makes sense adding a parameter for setting this value from users side (with a default value).
WDYT?

t0rr3sp3dr0:
I like the idea of making this configurable and also think that 200ms is a good default value for it.

JorTurFer:
@tommy351 , could you add a parameter for that value (and document it) ? 🙏

JorTurFer:
Hi @tommy351 
We will cut a release probably this month (we are coming back from the summer break and we are still a bit slow). Do you have any update about this?

JorTurFer:
Hello @tommy351 
Any update? We plan a release soon and this'd be a nice improvement

tommy351:
Hi @JorTurFer, is the parameter you mentioned application-wide or just for a single `ScaledObject`?

JorTurFer:
I think that we can start with a global parameter because the improvement will be huge. If someone has a use case where it's required at HTTPScaledObject, we can just itererate

JorTurFer (approved):

JorTurFer:
I add the config value in [another PR](https://github.com/kedacore/http-add-on/pull/806) because we want to cut a release tomorrow

PR #795, opened 2023-09-22
feat: allow skip scaled object creation with config

frankliu20:
When integrate Keda-Http with other KEDA scalers, we want to manage `ScaledObject` by ourselves, and add http-scaler as one of the triggers. In this PR, we allow users to do such `ScaledObject` management. When flag `SKIP_SCALED_OBJECT_CREATION` is set to `true`, keda-http-operator will skip the `ScaledObject` creation.

### Checklist

  - [`README.md`](/README.md)


JorTurFer:
I think that this is a nice feature and simplifies the compatibility between KEDA and the HTTP add-on because avoid the requirement of implementing the whole ScaledObject spec in the add-on. WDYT @tomkerkhove  @t0rr3sp3dr0 ?

@frankliu20 , Could you add an [e2e test for testing this feature](https://github.com/kedacore/http-add-on/tree/main/tests)? IMO this requires [an update in docs](https://github.com/kedacore/http-add-on/tree/main/docs) to explain that this can be done for other users

t0rr3sp3dr0:
I like the idea but would rather have this as a field in the HTTPSO definition instead of an Operator argument. That way, you could have some services using the managed SO while others use a custom one.

JorTurFer:
@frankliu20 , any update?

muecahit94:
Hi, any updates on that feature?

I would like to use zero scaling feature, but for that I need a combination of a keda built-in trigger and the http trigger, at the moment its not possible.

JorTurFer:
It looks that the PR has been abandoned by the OP :( 
Are you willing to continue it @muecahit94 ?

muecahit94:
@JorTurFer tbh no, just wanted to ask if there are any updates

devopsdynamo:
I've opened up new PR to allow annotation to be set on the HTTPSO instead as this allows more flexibility.
https://github.com/kedacore/http-add-on/pull/929

JorTurFer:
This feature is already done by https://github.com/kedacore/http-add-on/pull/929

PR #806, opened 2023-10-04
chore: Add missing changes for the release

JorTurFer:
### Checklist


pkg/util/env_resolver.go
```diff
@@ -0,0 +1,54 @@
+/*
```
JorTurFer, line 1: I know that I'm just using one of the helpers, but I copied/pasted them from KEDA and having them here can help in the future (or not ❓ )

zroubalik (approved):

Issue #438, opened 2022-05-05
Support for scaling custom resources and statefulset


raulsh:
Based on the docs and issues, there is no way to scale another kind of resources besides Deployments.  We need scale rollouts (CRD of Argo Rollouts). Even, we use core KEDA to scale rollouts.

### Use-Case

We use Argo Rollouts, so in Keda we can specify the Kind of the resource (https://keda.sh/docs/2.6/concepts/scaling-deployments/). This addon must scale any CRD/Resource with /scale subresource (Statefulset or Argo Rollout resource).

I use rollouts to deploy some HTTP workloads.

### Specification

Probably same as core KEDA, with parameters to specify the kind and the apiVersion of CRDs

tomkerkhove:
Just out of curiosity - How is HTTP scaling required for Argo Rollouts? (I'm not an argo expert)

raulsh:
Rollouts resources are similar to native Deployment resources. They are managed by Argo Rollout controller. The difference is that rollouts support Blue/Green and Canary deployments (and other functions). This addon must scale any CRD/Resource with /scale subresource (Statefulset or Argo Rollout resource).

tomkerkhove:
And those rollouts are being consumed over HTTP then? I'm just asking because I'm not sure why this is a problem for the HTTP add-on and why vanilla KEDA can't help you on this as that's supported there.

raulsh:
Yes... we use to deploy HTTP workloads (APIs, monoliths, etc). But, why the question? That's not the addon purpose? or am I misunderstanding something? (i'm not a english native speaker, sorry)

In fact, this is the perfect solution to our case. Because we have HTTP workloads that have multiple containers, but doesnt have a centralized queue to manage request and manage pending requests properly (while scaling workload). Because we have a high dynamic request throughput in our internal services. And I think this is the case that this addon want to cover.

Of course we can use vanilla Keda, but the "Interceptor" design is the solution that fit to our needs. And the main issue for us Is because we are using rollouts instead native deployments haha.

tomkerkhove:
> Yes... we use to deploy HTTP workloads (APIs, monoliths, etc). But, why the question? That's not the addon purpose? or am I misunderstanding something? (i'm not a english native speaker, sorry)

I was just curious about your scenario because my lack of experience with Argo Rollouts :)

[...]

PR #852, opened 2023-12-09
feat: Add support for any resource which implementes `/scale` subresource

JorTurFer:
This PR enables support for more workload types than just `Deployment`. 
To achieve this, I've changed the logic for detecting if the workload is active or not, using the active endpoints instead of checking deployment replica count. 
As we are routing HTTP traffic, I think that we can assume that scaled workloads always have at least one available endpoint (a service is a requirement, so maybe the endpoints have empty subsets, but the endpoints should be there).

I've added also some e2e tests for new cases

### Checklist

  - [`README.md`](/README.md)

Fixes https://github.com/kedacore/http-add-on/issues/438


JorTurFer:
PTAL: @tomkerkhove  @t0rr3sp3dr0

zroubalik (commented):
Looking good, though the PR is huge, so I might have missed something :)

JorTurFer:
> Looking good, though the PR is huge, so I might have missed something :)

so... can you approve it? xD

zroubalik (approved):

Issue #864, opened 2023-12-22
Forwarded request paths have escaped characters unescaped


worldspawn:
### Report

The request path `api/Spaces-1/projects/Projects-45/refs%2fheads%2fdevelop/deploymentprocesses` fails with 404 with keda http add-on in between. I believe this is because the escaped characters are unescaped in the forwarded request.

I get the same behaviour when i enable path rewriting in ingress.

### Expected Behavior

Path not altered. Don't get a 404.

### Actual Behavior

Path is altered, resulting in a 404.

### Steps to Reproduce the Problem

1. Make a request with an escaped forward slash `%2f` and observe it forwarded as `/`
2.
3.

### Logs from KEDA HTTP operator

```
{"level":"info","ts":1703256074.7780826,"logger":"LoggingMiddleware","caller":"middleware/logging.go:73","msg":"10.50.10.146:39356 - - [22/Dec/2023:14:41:14 +0000] \"GET /api/Spaces-1/projects/Projects-45/refs/heads/develop/deploymentprocesses HTTP/1.1\" 404 69 \"\" \"Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/120.0.0.0 Safari/537.36\""}
```

### HTTP Add-on Version

0.6.0

### Kubernetes Version

1.28

### Platform

Microsoft Azure

tomkerkhove:
Are you willing to contribute a fix?

worldspawn:
I'm completely out of my element here... like I had to lookup how to create a new instance of something go (iow, starting from zero).

From what I can gather, setting RawPath on the new request to the value of RawRequest of the forwarded request **might** fix this. Based on:

```
package main

import (
	"fmt"
	"log"
	"net/url"
)

func main() {
	u, err := url.Parse("http://example.com/api/Spaces-1/projects/Projects-45/refs%2fheads%2fdevelop/deploymentprocesses")
	if err != nil {
		log.Fatal(err)
	}

	p, err := url.Parse("http://example.com/")

	fmt.Println("Path:", u.Path)
	fmt.Println("RawPath:", u.RawPath)
	fmt.Println("EscapedPath:", u.EscapedPath())

	p.Path = u.Path
	fmt.Println("EscapedPath:", p.EscapedPath())
	p.RawPath = u.RawPath
	fmt.Println("EscapedPath:", p.EscapedPath())

}
```

[...]

PR #873, opened 2024-01-04
Set the proxy RawPath to original requests RawPath

worldspawn:
I have set the RawPath of what I think is the forward request to that of the original request with the expectation escaped values in the request path are forwarded unmodified (still escaped).

### Checklist

  - [`README.md`](/README.md)

Fixes #

#864


worldspawn:
I've tried this out (though my test was branched directly from v0.6.0) in our cluster using octopus deploy. It correctly this path `api/Spaces-1/projects/Projects-45/refs%2fheads%2fdevelop/deploymentprocesses` where the original v0.6.0 version of the interceptor would always return 404.

JorTurFer (approved):
LGTM!
Thanks for the fix! ❤️

PR #904, opened 2024-01-26
CooldownPeriod (IsActive) is now managed by the Scaler using the time since the last request and not in-flight requests

### LINKED ISSUE #882: Understanding scaledownPeriod
Hi everyone! 👋🏻

We've been utilizing KEDA successfully for a while now, scaling our deployments and enabling scaling to zero.

Recently, I've been delving into implementing the `http-add-on`.
I expected that `HTTPScaledObjects` would mirror the behavior of `ScaledObjects`, featuring a target value (trigger result or, in this case, pending requests) and a cooldown period that resets with each trigger result / new request.

However, my initial testing suggests that `HTTPScaledObjects` doesn't quite work that way. Regardless of the number of requests sent through, it consistently adheres to the `scaledownPeriod` and scales the deployment down from the moment it scaled up + `scaledownPeriod`, even if it received a request just a second ago.

**Is this the intended behavior**? If so, is there a way to configure the resources to match the behavior of `ScaledObjects`, where the `scaledownPeriod` resets with every received request?

Additionally, we have customers with burst periods where they will send high volumes of requests per second. I want to ensure that we do not lose requests here via the interceptor.
I assume that if the interceptor is not performing, it will act as a sort of queue, and I'd like to avoid that.

Appreciate any insights or tips you can share! 👍🏻

JorTurFer:
### Checklist

Fixes #882

### DIFF
--- a/scaler/handlers_test.go
+++ b/scaler/handlers_test.go
@@ -29,15 +29,16 @@ import (
 func TestStreamIsActive(t *testing.T) {
 	type testCase struct {
 		name           string
-		expected       bool
+		active         bool
 		expectedErr    bool
 		setup          func(t *testing.T, qp *queuePinger)
+		coolDownPeriod *int32
 		scalerMetadata map[string]string
 	}
 	testCases := []testCase{
 		{
-			name:        "Simple host inactive",
-			expected:    false,
+			name:        "Simple host inactive (deafult values)",
+			active:      false,
 			expectedErr: false,
 			setup: func(t *testing.T, qp *queuePinger) {
 				namespacedName := &types.NamespacedName{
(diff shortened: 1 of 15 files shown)


scaler/handlers_test.go
```diff
@@ -185,140 +234,8 @@ func TestStreamIsActive(t *testing.T) {
...
 			}
 
-			if tc.expected != res.Result {
-				t.Fatalf("Expected IsActive result %v, got: %v", tc.expected, res.Result)
-			}
-		})
-	}
-}
-
-func TestIsActive(t *testing.T) {
```
JorTurFer, line 195: This test was duplicated. Tests cases were almost the same as for IsActiveStream and inside IsActiveStream we call to IsActive

JorTurFer:
I'm thinking about this, and we could add a new parameter in HTTPScaledObject to switch between old and this new behaviours. I don't think that it's useful as the old behaviour seems that not worth for HTTP Scaling, but the options is there. WDYT?

PR #919, opened 2024-02-14
fix: keda interceptor proxy service name

lenglet-k:
The name of interceptor is not correct. The default name configured by the helm chart is: keda-add-ons-http-interceptor-proxy

### Checklist

  - [`README.md`](/README.md)

Fixes #


JorTurFer:
I think that helm case is also covered:
image

I agree that maybe we could swap them and using helm chart urls as default adding a note about the raw manifest but as most probably will have to break the chart before GA to rename multiple stuff, we could just break the chart now being in beta. @tomkerkhove  @zroubalik ?

PR #928, opened 2024-02-20
feat: Adding support to the interceptor proxy for TLS on the wire

### LINKED ISSUE #907: Interceptor Proxy - TLS On The Wire
### Proposal

Currently the interceptor proxy does not support TLS on the wire for requests that it receives or forwards on to target services. In order to facilitate running the interceptor proxy in a secure cluster environment it should be able to serve all traffic over a TLS enabled server and be able to forward traffic on to destination services over a TLS connection.

### Use-Case

Running the interceptor proxy in secured cluster environments where TLS on the wire is mandatory.

Yes

### Anything else?

_No response_

zorocloud:
Adding configurable TLS support to the interceptor proxy. Resolves #907

### Checklist

  - [`README.md`](/README.md)

### DIFF
--- a/pkg/http/server.go
+++ b/pkg/http/server.go
@@ -7,7 +7,7 @@ import (
 	"github.com/kedacore/http-add-on/pkg/util"
 )
 
-func ServeContext(ctx context.Context, addr string, hdl http.Handler) error {
+func ServeContext(ctx context.Context, addr string, hdl http.Handler, tlsEnabled bool, tlsConfig map[string]string) error {
 	srv := &http.Server{
 		Handler: hdl,
 		Addr:    addr,
@@ -21,6 +21,10 @@ func ServeContext(ctx context.Context, addr string, hdl http.Handler) error {
 			logger.Error(err, "failed shutting down server")
 		}
 	}()
+
+	if tlsEnabled {
+		return srv.ListenAndServeTLS(tlsConfig["certificatePath"], tlsConfig["keyPath"])
+	}
 
 	return srv.ListenAndServe()
 }
--- a/.gitignore
+++ b/.gitignore
@@ -355,3 +355,10 @@ admin/Cargo.lock
 
 /target
 .envrc
+
+# locally generated certs for testing TLS
+*.crt
+*.pem
+*.csr
+*.srl
+*.ext
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,7 +29,7 @@ This changelog keeps track of work items that have been completed and are ready 
 
 ### Improvements
 
-- **General**: TODO ([#TODO](https://github.com/kedacore/http-add-on/issues/TODO))
+- **General**: Add configurable TLS on the wire support to the interceptor proxy ([#907](https://github.com/kedacore/http-add-on/issues/907))
 
 ### Fixes
 
--- a/Makefile
+++ b/Makefile
@@ -38,6 +38,8 @@ keyUsage = digitalSignature, nonRepudiation, keyEncipherment, dataEncipherment
 subjectAltName = @alt_names
 [alt_names]
 DNS.1 = localhost
+DNS.2 = *.keda
+DNS.3 = *.interceptor-tls-test-ns
 endef
 export DOMAINS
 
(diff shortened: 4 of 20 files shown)


pkg/http/server.go
```diff
@@ -22,5 +22,9 @@ func ServeContext(ctx context.Context, addr string, hdl http.Handler) error {
 		}
 	}()
 
+	if tlsEnabled {
+		return srv.ListenAndServeTLS(tlsConfig["certificatePath"], tlsConfig["keyPath"])
+	}
+
 	return srv.ListenAndServe()
```
JorTurFer, line 29: Shouldn't we allow both? I mean, will we enforce that all the traffic uses HTTPS or HTTP?  I'm not directly against this because we have in the roadmap the plan of scoped interceptors, but I prefer to expose both and route the request based on the incoming protocol

JorTurFer (commented):
I like this feature, could you add an e2e test for this? I think that covering this will prevent problems in the future xD

pkg/http/server.go
zorocloud: We could but I can't see a use case for supporting both simultaneously - unless maybe during a migration scenario? Do you have any examples where you could see serving both and routing based on the protocol to be beneficial?

JorTurFer: The problem is that currently, the interceptor is shared. If you have a cluster with multiple workloads, you could be interested on both scenarios, but it's true that probably you won't be interested on having both during a long time, just during migrations 🤔 

Maybe we can continue with the PR as it is (just supporting one of them) and work on dedicated interceptors before next release. This can allow both scenarios easily
WDYT @tomkerkhove ?

tomkerkhove: I agree exposing both would be good to offer

JorTurFer: I guess that we can go with this as it is and add it as nice to have before GA.

JorTurFer (commented):
Could you add an e2e test case for this scenario @zorocloud?
We should update docs as well for explaining how to configure this

zorocloud:
> Could you add an e2e test case for this scenario @zorocloud? We should update docs as well for explaining how to configure this

Yeah no problem @JorTurFer. Planning to add them in once I get my other PR tied off. Just wanted to get the initial code up for feedback on the implementation :)

JorTurFer:
> Yeah no problem @JorTurFer. Planning to add them in once I get my other PR tied off. Just wanted to get the initial code up for feedback on the implementation :)

sure! no rush, your other PR will be merged today (sorry, I've been on a trip this week)

JorTurFer:
Hello @zorocloud 
We plan to cut a release (v0.8.0) during this week or the next one. By change will you have time to finish the PR before? Don't feel rush or pressure, we can deploy the release a bit, or follow the plan and cut another release (v0.8.1) with this feature and others new, we don't have a super strict release plans for the add-on.

zorocloud:
Hey @JorTurFer, I will try and get it tied off this week, but I'm happy for it to go into a follow up release if it's not done in time. May as well stick with the current plan :)

JorTurFer:
> Hey @JorTurFer, I will try and get it tied off this week, but I'm happy for it to go into a follow up release if it's not done in time. May as well stick with the current plan :)

Well... we don't have any defined roadmap yet for the add-on, so given that you're working on this and I assume that you need the feature, we can cut a release v0.8.1 once it's the feature merged. Would it help? 
We don't need to be in hurry and we can cut the release once the feature is ready, but we can ship current features. WDYT?

zorocloud:
> > Hey @JorTurFer, I will try and get it tied off this week, but I'm happy for it to go into a follow up release if it's not done in time. May as well stick with the current plan :)
> 
> Well... we don't have any defined roadmap yet for the add-on, so given that you're working on this and I assume that you need the feature, we can cut a release v0.8.1 once it's the feature merged. Would it help? We don't need to be in hurry and we can cut the release once the feature is ready, but we can ship current features. WDYT? @tomkerkhove , do you have any objections?

Works for me :+1:

zorocloud:
Couldn't bind 443 in the tests so just pushed a small update that should fix that. Think the tests should be good now and the PR ready for re-review whenever you have time :)

JorTurFer:
Please fix the DCO: 
image
 You can click on `Details` and get the commands to execute

zorocloud:
> Please fix the DCO: image You can click on `Details` and get the commands to execute

Ah forgot to add that to my last commit. I've amended the commit message so we should be sweet now.

JorTurFer:
Nice, I'm going to check it now :)

JorTurFer (approved):
AWESOME JOB! ❤️

PR #929, opened 2024-02-22
feat: provide support to allow HTTP scaler to work alongside other core KEDA scalers

devopsdynamo:
This change provides support to allow HTTP scaler to work alongside other core KEDA scalers by allowing skipScaledObjectCreation annotation to be set on an HTTPScaledObject.  This will ensure the reconciler to skip the KEDA core ScaledObjects creation which alllows you to create and manage your own ScaledObject with addtional triggers.

### Checklist

  - [`README.md`](/README.md)

Fixes #


tomkerkhove:
Hey, mind sharing some samples on what that would look like? Is the problem that you'd otherwise have 2 scaledobjects or?

devopsdynamo:
@tomkerkhove  Yeah the problem we have at the moment is that we already have a scaledobject with a kafka trigger added for scaling but we also have a requirement to add HTTP scaling to the same deployment so we can scale up on http requests also.  This change will allow you to add skip annotation on the HTTPSO so you can add the external-push trigger to your existing scaledobject as an additional trigger etc.

JorTurFer:
I like the solution but I'm worried about how this will scale in the future. Currently, we support `hosts` and `prefixes` properties within external scaler metadata but maybe we need more info in the future. 
I'm thinking that maybe we should replace them with the HTTPScaledObject name instead, but it enforces the scaler to know about the HTTPScaledObject and permissions for pulling them to selfconfigure it :(
WDYT @tomkerkhove ?

operator/controllers/http/app.go
```diff
@@ -41,6 +41,17 @@ func (r *HTTPScaledObjectReconciler) createOrUpdateApplicationResources(
...
+	// scalers. when SkipScaledObjectCreation is set to true,
+	// reconciler will skip the KEDA core ScaledObjects creation.
+	// you can create your own so, and add http scaler as one of your triggers.
+	if httpso.Annotations["skipScaledObjectCreation"] == "true" {
+		logger.Info(
+			"Skip scaled objects creation with flag SkipScaledObjectCreation=true",
+			"HTTPScaledObject", httpso.Name)
+		return nil
+	}
+
```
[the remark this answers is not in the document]
devopsdynamo: @JorTurFer I've a had a look at this,  I have a slight concern in that if you were to create your own ScaledObject with the same name at the same time as adding the annotation on the httpso as part of a helm upgrade etc,  there might be a possibility that the scaledObject could get updated but keep the ownerReferences on the object resulting in the scaledObject getting removed unintentionally by the operator when it checks for the annotation & ownerReferences.  The logic I've tested is scaledObject will get removed if kind=httpscaledobject and and name matches the same name of httpso in ownerreferences.   I'm happy to make these changes but would like to hear your thoughts?

docs/walkthrough.md
```diff
@@ -89,4 +89,23 @@ kubectl port-forward svc/keda-http-add-on-interceptor-proxy -n ${NAMESPACE} 8080
...
+```
+
+Step 2, add an external-push trigger to your existing ScaledObject.  The hosts should match the list of hosts in your HTTPScaledObject.
+```console
+  triggers:
+  - type: external-push
+    metadata:
+      hosts: example-service
+      scalerAddress: keda-add-ons-http-external-scaler.keda:9090  
+```
```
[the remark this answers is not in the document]
devopsdynamo: Ok I'll look to amend the docs

operator/controllers/http/app.go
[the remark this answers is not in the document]
JorTurFer: you are right in that point, but not removing the old one can generate problems:
- KEDA admission webhooks will reject the new one as the workload is already scaled by the old one (if they use different names)
- If both are at the same time in the cluster, they will scale the same workload, generating unexpected scaling behaviors
- The old one will have the current HTTPScaledObject as owner, so if you remove the HTTP scaler but you maintain the same ScaledObject, removing the HTTPScaledObject will remove the ScaledObject, although they are detached

Honestly, I don't have a good way to do this is in a single step :/

Maybe we cloud introduce an annotation in the ScaledObject that the HTTP add-on operator checks before updating it? 🤔  Maybe we can check both resources (HTTPScaledObject and ScaledObject) and if the HTTPScaledObject sets `skipScaledObjectCreation:true`, we only modify/delete the ScaledObject when the ownership reference is correct.

IDK if this will work as I expect tbh. WDYT?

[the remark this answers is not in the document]
devopsdynamo: Ok, I already have above solution working which checks the scaledObject  for the ownerReferences and deletes object if kind=httpScaledObject and name matches the httpso name when annotation is set on the httpso.  I was wondering if another (less risky) option would be to instead of deleting the scaledObject, the operator could just remove the OwnerReferences on the scaledObject and the existing scaledObject can then be taken over to self-manage which would ensure that scaledObject does not need to be replaced.  Thoughts?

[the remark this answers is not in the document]
JorTurFer: The problem is that we should ensure that resources are removed correctly or we could keep dangling resources :/ 

Maybe we could just remove the reference but in that case, we should document this case well to prevent future issues complaining about that

operator/controllers/http/httpscaledobject_controller_test.go
```diff
@@ -0,0 +1,73 @@
...
+	_, err = getSO(
+		testInfra.ctx,
+		testInfra.cl,
+		testInfra.httpso,
+	)
+	r.NoError(err)
+
+}
+
+func TestHttpScaledObjectControllerWhenSkipAnnotationSet(t *testing.T) {
```
JorTurFer, line 42: Could we add a test covering the scenario where the ScaledObject is removed and a normal scenario where a regular ScaledObject (not deleted). Adding those 2 scenarios we can cover that everything works as expected

docs/walkthrough.md
```diff
@@ -89,4 +89,38 @@ kubectl port-forward svc/keda-http-add-on-interceptor-proxy -n ${NAMESPACE} 8080
 curl -H "Host: myhost.com" localhost:8080/path1
 ```
 
+### Integrating HTTP Add-On Scaler with other KEDA scalers
```
JorTurFer, line 92: @tomkerkhove PTAL

JorTurFer (commented):
Awesome contribution ❤️ 
Just small feedback inline

JorTurFer:
There are also some static checks failures: https://github.com/kedacore/http-add-on/actions/runs/8230990823/job/22505562138?pr=929

operator/controllers/http/httpscaledobject_controller_test.go
devopsdynamo: sure, will take a look at this

operator/controllers/http/httpscaledobject_controller_test.go
```diff
@@ -0,0 +1,73 @@
+package http
+
+import (
+	"github.com/kedacore/http-add-on/operator/controllers/http/config"
+	"github.com/stretchr/testify/require"
+	"testing"
```
JorTurFer, line 6: We split the imports in 3 groups:
- runtime
- deps
- internals
```suggestion
	"testing"
	
	"github.com/stretchr/testify/require"

	"github.com/kedacore/http-add-on/operator/controllers/http/config"
```

(it's one of the static checks' failure)

operator/controllers/http/httpscaledobject_controller_test.go
devopsdynamo: @JorTurFer just looking at this, would you mind on expanding/clarifying on what tests you require for this as I'm not 100% on what your'e asking?   The 2 tests I created are already checking if the scaledObject exists or not depending if annotation is set

JorTurFer: sorry, I didn't see the test above (`TestHttpScaledObjectController`) which is one of them indeed. The other test that I said is:
- Deploy a regular HTTPScaledObject
- Check that the SO is correctly generated
- Update the HTTPScaledObject with the skip flag
- Check that the SO has been removed

JorTurFer:
We need to fix the DCO check too, you can check the steps just clicking the link:
image

operator/controllers/http/httpscaledobject_controller_test.go
devopsdynamo: added a new test for above if you can review, thanks

docs/walkthrough.md
```diff
@@ -89,4 +89,38 @@ kubectl port-forward svc/keda-http-add-on-interceptor-proxy -n ${NAMESPACE} 8080
...
+
+For scenerios where you want to integrate HTTP Add-On scaler with other keda scalers, you can set the SkipScaledObjectCreation annotation to true on your HTTPScaledObject.  The reconciler will then skip the KEDA core ScaledObject creation which will allow you to create your own ScaledObject and add http scaler as one of your triggers.  Please note that you should ensure the ScaledObject is created with a different name to the HTTPScaledObject to ensure your ScaledObject is not removed by the reconciler.
```
tomkerkhove, line 94:
```suggestion
For scenerios where you want to integrate HTTP Add-On scaler with other keda scalers, you can set the `SkipScaledObjectCreation` annotation to true on your `HTTPScaledObject`.  The reconciler will then skip the KEDA core ScaledObject creation which will allow you to create your own `ScaledObject` and add HTTP scaler as one of your triggers.

> 💡 Ensure that your ScaledObject is created with a different name than the `HTTPScaledObject` to ensure your ScaledObject is not removed by the reconciler.
```

docs/walkthrough.md
```diff
@@ -89,4 +89,38 @@ kubectl port-forward svc/keda-http-add-on-interceptor-proxy -n ${NAMESPACE} 8080
...
+
+It is reccomended that you first deploy your HTTPScaledObject with no annotation set in order to obtain the latest trigger spec to use on your own managed ScaledObject.
+
+Step 1, first deploy your HTTPSCaledObject with annotation set to false
```
tomkerkhove, line 98:
```suggestion
1. Deploy your `HTTPScaledObject` with annotation set to false
```

docs/walkthrough.md
```diff
@@ -89,4 +89,38 @@ kubectl port-forward svc/keda-http-add-on-interceptor-proxy -n ${NAMESPACE} 8080
...
+
+```console
+annotations:
+  skipScaledObjectCreation: false
+```
+
+Step 2, take a copy of the current generated external-push trigger spec on the generated ScaledObject.  Please find example below, however this spec is likely to change on future releases.
```
tomkerkhove, line 105:
```suggestion
2. Copy of the current generated external-push trigger spec on the generated ScaledObject.  ```

File: docs/walkthrough.md [hunk 9]
```diff
@@ -89,4 +89,38 @@ kubectl port-forward svc/keda-http-add-on-interceptor-proxy -n ${NAMESPACE} 8080
...
+
+```console
```
tomkerkhove, line 107:
```suggestion
For example:

```console
```

docs/walkthrough.md
```diff
@@ -89,4 +89,38 @@ kubectl port-forward svc/keda-http-add-on-interceptor-proxy -n ${NAMESPACE} 8080
...
+  triggers:
+  - type: external-push
+    metadata:
+      hosts: example-service
+      pathPrefixes: ""
+      scalerAddress: keda-http-add-on-external-scaler.keda:9090
+```
+
+Step 3, update the skipScaledObjectCreation annotation to true and re-deploy.   This will remove the ScaledObject and allow you to then create your own.
```
tomkerkhove, line 116:
```suggestion
3. Apply the `skipScaledObjectCreation` annotation with `true` and apply the change. This will remove the originally created `ScaledObject` allowing you to create your own.
```

docs/walkthrough.md
```diff
@@ -89,4 +89,38 @@ kubectl port-forward svc/keda-http-add-on-interceptor-proxy -n ${NAMESPACE} 8080
...
+
+```console
+annotations:
+  skipScaledObjectCreation: true
+```
+
+Step 4, add the external-push trigger taken from step 2 to your own ScaledObject and apply this.
+
```
tomkerkhove, line 124:
```suggestion
4. Add the `external-push` trigger taken from step 2 to your own ScaledObject and apply this.
```

tomkerkhove (commented):
Good guide, added a few suggestion!

JorTurFer:
There are also some linting errors, could you take a look? https://github.com/kedacore/http-add-on/actions/runs/8261625169/job/22724018316?pr=929

devopsdynamo:
@tomkerkhove updated docs as per your suggestions, thanks.  @JorTurFer lint errors should be fixed now also,  my bad :)

CHANGELOG.md
```diff
@@ -26,7 +26,7 @@ This changelog keeps track of work items that have been completed and are ready
 
 ### Improvements
 
-- **General**: TODO ([#TODO](https://github.com/kedacore/http-add-on/issues/TODO))
+- **Operator**: Provide support to allow HTTP scaler to work alongside other core KEDA scalers ([#489](https://github.com/kedacore/http-add-on/issues/489))
```
tomkerkhove, line 29: @JorTurFer i'd say this is a "general" feature and not an improvement, but up to you

docs/walkthrough.md
```diff
@@ -89,4 +89,43 @@ kubectl port-forward svc/keda-http-add-on-interceptor-proxy -n ${NAMESPACE} 8080
...
+3. Apply the `skipScaledObjectCreation` annotation with `true` and apply the change. This will remove the originally created `ScaledObject` allowing you to create your own.
+
+```console
+annotations:
+  skipScaledObjectCreation: true
+```
+
+4. Add the `external-push` trigger taken from step 2 to your own ScaledObject and apply this.
+
+
```
tomkerkhove, line 130:
```suggestion

```

docs/walkthrough.md
```diff
@@ -89,4 +89,43 @@ kubectl port-forward svc/keda-http-add-on-interceptor-proxy -n ${NAMESPACE} 8080
...
+1. Deploy your `HTTPScaledObject` with annotation set to false
+
+```console
+annotations:
+  skipScaledObjectCreation: false
+```
+
+2. Take copy of the current generated external-push trigger spec on the generated ScaledObject. 
+
+
```
tomkerkhove, line 109:
```suggestion

```

tomkerkhove (commented):
Nits

CHANGELOG.md
JorTurFer: Yeah, let's move it to `New` section and rename it to **General**

JorTurFer (commented):
LGTM! Just the change in the changelog and we are ready to merge this! ❤️

devopsdynamo:
> LGTM! Just the change in the changelog and we are ready to merge this! ❤️

Updated as requested

JorTurFer (approved):
Awesome contribution! Thanks for it ❤️

Issue #33, opened 2021-02-02
Support Kubernetes Gateway APIs for Ingress


beriberikix:
The network SIG has been hard at work developing the new [Gateway API](https://gateway-api.sigs.k8s.io/), aka Ingress V2. A Gateway is responsible for routing a request to a service. There's already an HTTPRoute [implementation](https://kubernetes-sigs.github.io/service-apis/httproute/). The API has several [integration points](https://kubernetes-sigs.github.io/service-apis/api-overview/#extension-points) and I think the current Interceptor could be integrated or evolved to integrate with the existing HTTP implementation. Alternatively, a new [GatewayClass](https://kubernetes-sigs.github.io/service-apis/api-overview/#gatewayclass) can be designed to work specifically with KEDA and still make use of the standardized API.

Might relate to #10 and possibly #6. /cc folks from Services API who know way more than I do :) @hbagdi @jpeach @robscott

_Edit @tomkerkhove: Updated link given it was renamed to Gateway API._

arschles:
@beriberikix sounds like it'd be useful to have the operate create service API resources instead of (in addition to?) plain old `Ingress` resources that #10 calls for.

Another aspect is metrics. We can do routing with this, but is there a standard metrics API that this comes with? SMI is certainly one solution but something designed explicitly for north-south traffic, as you said in today's meeting, could be more beneficial.

Looking forward to hearing your thoughts!

tomkerkhove:
I think the same here @arschles - Folks should be able to choose if we manage it or they bring their own imo

beriberikix:
💯! It's also entirely possible someone has a very custom Ingress solution and this shouldn't prevent them as well. Services API should be *in addition* to `Ingress` (and user choice which to instantiate), especially since Ingress V1 is currently the only GA supported way and Service API is still in Beta. 

@arschles great questions on Metrics. I've asked in #sig-network-service-apis channel for input!

hbagdi:
> Service API is still in Beta.

nit: Service APIs (or Gateway API in future) is in alpha.

> It's also entirely possible someone has a very custom Ingress solution and this shouldn't prevent them as well. Services API should be in addition to Ingress

That's correct. Based on the community interest, it seems like most ingress controllers will add support for Service APIs in future. Ingress isn't going anywhere.

beriberikix:
RE: metrics @robscott:

[...]

PR #1040, opened 2024-05-24
document GatewayAPI example with xkcd application

wozniakjan:
This PR documents additional steps to try http-add-on along with GatewayAPI, similar to how `Igress` setup is documented.

The selected GatewayAPI implementation [Envoy Gateway](https://github.com/envoyproxy/gateway/) is stable and well adopted by the community, which should complement the `Ingress` documentation with `nginx-ingress`. The `HTTPRoute` interceptor network path is very similar to the `Ingress` interceptor path.

### Checklist

see also: https://github.com/kedacore/http-add-on/issues/33

### DIFF
--- a/docs/walkthrough.md
+++ b/docs/walkthrough.md
@@ -14,6 +14,13 @@ You'll need to install a `Deployment` and `Service` first. You'll tell the add-o
 helm install xkcd ./examples/xkcd -n ${NAMESPACE}
 ```
 
+#### xkcd exposed with GatewayAPI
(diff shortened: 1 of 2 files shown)


JorTurFer:
Nice improvement! 
Does it make sense covering the GatewayAPI integration using an e2e tests? I mean, not for the docs but for the behaviour as a way to don't miss the compatibility (or maybe it's totally an overkill 🤔 )

wozniakjan:
> Does it make sense covering the GatewayAPI integration using an e2e tests?

that is a good point, ok if I tackle that in a standalone PR?

JorTurFer (approved):

Issue #1075, opened 2024-06-24
RPS doesn't decrease when requests stop suddenly


wozniakjan:
### Report

When using RPS as a scaler metric, it works well as long as there is some traffic for a particular `HTTPScaledObject`. But when the traffic suddenly stops, it takes the duration of the window for the RPS metric to jump to `0` and until then, it remains steady.

### Expected Behavior

The RPS should monotonously decrement during the window where there is no traffic.

### Actual Behavior

The RPS doesn't decrement if traffic suddenly stops.

### Steps to Reproduce the Problem

1. create `HTTPScaledObject` with `scalingMetric` `requestRate`
2. generate traffic for the application
3. observe RPS grow according to the traffic
4. cut the traffic suddenly
5. RPS remains close to the last value, application doesn't get scaled down
6. after the entire `window` elapses, RPS drops to 0 and application finally scales down 

### Logs from KEDA HTTP operator

attached as files
[keda-add-ons-http-controller-manager.log](https://github.com/user-attachments/files/15954329/keda-add-ons-http-controller-manager.log)
[keda-add-ons-http-interceptor.log](https://github.com/user-attachments/files/15954330/keda-add-ons-http-interceptor.log)

### HTTP Add-on Version

0.8.0

### Kubernetes Version

1.28

### Platform

Other

### Anything else?

after a little bit of digging, I think this might be due to the implementation of counter middleware. The `Record` for RPS seems to be only called when there is traffic for that particular application and values to the buckets get inserted. It should probably also insert zero values automatically to the buckets when there is no traffic.

https://github.com/kedacore/http-add-on/blob/1e27b395bd5db2e37b11bd41e142d841583eb591/pkg/queue/queue.go#L67-L73

wozniakjan:
ah, never mind, looks like this is by design
https://github.com/kedacore/http-add-on/blob/1e27b395bd5db2e37b11bd41e142d841583eb591/pkg/queue/bucketing.go#L105-L118

I'd still like to make an argument that the RPS should decrease if the traffic goes to zero. At the moment, as implemented, if traffic drops from 10 to 0 RPS, it yields a higher RPS metric compared to traffic dropping from 10 to 1 RPS. This intuitively doesn't feel correct to me.

Wdyt @JorTurFer, @zroubalik?

PR #1076, opened 2024-06-24
fix: include trailing 0 window buckets in RPS calculation

wozniakjan:
As mentioned in https://github.com/kedacore/http-add-on/issues/1075, the RPS calculation excludes trailing 0 request buckets from RPS windows. This may result in counterintuitive RPS measurements, for example:
* for an increasing window with buckets 1, 2, 3, 4 ... when traffic stops suddenly with 0, the RPS actually continues to grow until the entire window expires and then RPS drops to 0
* for constant window with buckets of values 10, 10, 10 ... when traffic stops, the RPS remains 10 but when traffic just becomes 1, the RPS decreases

Both examples would treat 0 requests to be more than 1 request in a window bucket which feels incorrect.

### Checklist

Fixes https://github.com/kedacore/http-add-on/issues/1075

### DIFF
--- a/pkg/queue/bucketing_test.go
+++ b/pkg/queue/bucketing_test.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	"github.com/google/go-cmp/cmp"
+	. "github.com/onsi/gomega"
 )
 
 const granularity = time.Second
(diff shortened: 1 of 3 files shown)


JorTurFer (approved):
Nice improvement! Could we ensure somehow that 1 request during the given period keeps the workload active (basically because we don't round the value to 0 before exposing it to KEDA)

zroubalik (approved):

wozniakjan:
> Could we ensure somehow that 1 request during the given period keeps the workload active

unless I'm missing something, it should remain operational, I will include one more test to make the small impact of this a bit more apparent.

pkg/queue/bucketing_test.go
```diff
@@ -252,6 +253,87 @@ func TestDescendingRecord(t *testing.T) {
 	}
 }
 
+func TestRequestsSuddenStop(t *testing.T) {
+	RegisterTestingT(t)
+	now := time.Date(2024, 6, 26, 12, 0, 0, 0, time.UTC)
```
wozniakjan, line 258: using a fixed timestamp to get deterministic buckets, given the circular buffer uses modulo on seconds, this is one way to ensure the bucket values in the window buffer can be asserted against easily

pkg/queue/bucketing_test.go
```diff
@@ -252,6 +253,87 @@ func TestDescendingRecord(t *testing.T) {
...
+	Expect(buckets.buckets).To(Equal([]int{1, 2, 3, 4, 0}))
+
+	// fifth bucket with 5 requests
+	buckets.Record(now.Add(4*time.Second), 5)
+	Expect(buckets.WindowAverage(now.Add(4 * time.Second))).To(Equal(3.0))
+	Expect(buckets.buckets).To(Equal([]int{1, 2, 3, 4, 5}))
+
+	// first bucket (sixth time window), we don't have any requests, so the average should be 0+2+3+4+5/5 = 2.8
+	// but the buckets don't change until new value is recorded or until the window expires
+	Expect(buckets.WindowAverage(now.Add(5 * time.Second))).To(Equal(2.8))
```
wozniakjan, line 292: without this change, this would have been `(2+3+4+5)/(5-1) = 3.5` which means increase from `3.0` after just receiving 0 requests in the sixth time window

pkg/queue/bucketing_test.go
```diff
@@ -252,6 +253,87 @@ func TestDescendingRecord(t *testing.T) {
...
+	buckets.Record(now.Add(11*time.Second), 12)
+	Expect(buckets.WindowAverage(now.Add(11 * time.Second))).To(Equal(10.0))
+	Expect(buckets.buckets).To(Equal([]int{11, 12, 8, 9, 10}))
+
+	// now requests stop entirely and time window average decreases all the way to 0
+	Expect(buckets.WindowAverage(now.Add(12 * time.Second))).To(Equal(8.4))
+	Expect(buckets.WindowAverage(now.Add(13 * time.Second))).To(Equal(6.6))
+	Expect(buckets.WindowAverage(now.Add(14 * time.Second))).To(Equal(4.6))
+	Expect(buckets.WindowAverage(now.Add(15 * time.Second))).To(Equal(2.4))
+	Expect(buckets.WindowAverage(now.Add(16 * time.Second))).To(Equal(0.0))
```
wozniakjan, line 329: and even more visible here, right now RPS goes from `10` when traffic stops to 

```
new behaviour: 10.0 ->  8.4 ->  6.6 ->  4.6 ->  2.4 -> 0.0
old behaviour: 10.0 -> 10.5 -> 11.0 -> 11.5 -> 12.0 -> 0.0
```
the start and end result is the same, just the behaviour in between was imho misleading

wozniakjan (commented):
@JorTurFer I think this low level test makes it a bit more clear what this change means, ptal

Issue #1123, opened 2024-08-30
Critical Vulnerability - CVE-2024-24790


willemvs:
Good day,

I have been on journey to utilize KEDA Http Add On for a PoC and through the adoption process scanned the container images for potential vulnerabilities. The scanning tool used is Prisma, and it has indicated that there are in fact a critical vulnerability for most of the images utilized in kedacore/http-add-on.

It might also be worth considering updating of gcr.io/kubebuilder/kube-rbac-proxy:v0.13.0 to v0.14.1 or higher as there are also [vulnerabilities](https://github.com/ansible/awx-operator/issues/1437) that have been addressed.

The images that was scanned on Prisma was the following:
- ghcr.io/kedacore/http-add-on-operator:0.8.0
- ghcr.io/kedacore/http-add-on-scaler:0.8.0
- ghcr.io/kedacore/http-add-on-interceptor:0.8.0 

+----------------+----------+------+-----------+---------+--------------------------+-----------+------------+------------+----------------------------------------------------+-------------------+
|      CVE       | SEVERITY | CVSS |  PACKAGE  | VERSION |          STATUS          | PUBLISHED | DISCOVERED | GRACE DAYS |                    DESCRIPTION                     | TRIGGERED FAILURE |
+----------------+----------+------+-----------+---------+--------------------------+-----------+------------+------------+----------------------------------------------------+-------------------+
| CVE-2024-24790 | critical | 9.80 | net/netip | 1.22.2  | fixed in 1.21.11, 1.22.4 | 84 days   | < 1 hour   | -25        | The various Is methods (IsPrivate, IsLoopback,     | Yes               |
|                |          |      |           |         | 85 days ago              |           |            |            | etc) did not work as expected for IPv4-mapped IPv6 |                   |
|                |          |      |           |         |                          |           |            |            | addresses, returning false for addresses which     |                   |
|                |          |      |           |         |                          |           |            |            | would...                                           |                   |
+----------------+----------+------+-----------+---------+--------------------------+-----------+------------+------------+----------------------------------------------------+-------------------+

Regards

JorTurFer:
Thanks for reporting! We are going to upgrade golang version asap

khauser:
@JorTurFer : gcr.io/kubebuilder/kube-rbac-proxy:v0.13.0 is still referenced in the `values.yaml`. Why is that?

Maybe:
```
kubeRbacProxy:
        name: quay.io/brancz/kube-rbac-proxy
[...]

PR #1135, opened 2024-09-02
bump deps

zroubalik:


JorTurFer (approved):

JorTurFer:
Could you upgrade golang version too? To fix [this issue](https://github.com/kedacore/http-add-on/issues/1123) (9.8), we need to use at least go 1.22.4. As we have generated the image for 1.22.6, we could just bump golang version too

JorTurFer (approved):

wozniakjan (approved):

PR #1174, opened 2024-10-23
support portName in HTTPScaledObject service scaleTargetRef

wozniakjan:
The `HTTPScaledObject` allowed setting only numerical `port` value for the scaled services. For improved UX, it also allows to set `namedPort` to match common conventions in the kubernetes API regarding `Services` and ports.

### Checklist

### DIFF
--- a/config/interceptor/role.yaml
+++ b/config/interceptor/role.yaml
@@ -13,6 +13,14 @@ rules:
   - list
   - watch
 - apiGroups:
+  - ""
+  resources:
+  - services
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
   - http.keda.sh
   resources:
   - httpscaledobjects
--- a/operator/apis/http/v1alpha1/httpscaledobject_types.go
+++ b/operator/apis/http/v1alpha1/httpscaledobject_types.go
@@ -31,10 +31,10 @@ type ScaleTargetRef struct {
 	// The name of the service to route to
 	Service string `json:"service"`
 	// The port to route to
-	// +optional
+	// +kubebuilder:validation:XValidation:rule=!has(self.PortName) && has(self.Port)
 	Port int32 `json:"port,omitempty"`
 	// The port to route to referenced by name
-	// +optional
+	// +kubebuilder:validation:XValidation:rule=!has(self.Port) && has(self.PortName)
 	PortName string `json:"portName,omitempty"`
 }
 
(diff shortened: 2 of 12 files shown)


operator/apis/http/v1alpha1/httpscaledobject_types.go
```diff
@@ -31,7 +31,11 @@ type ScaleTargetRef struct {
 	// The name of the service to route to
 	Service string `json:"service"`
 	// The port to route to
-	Port int32 `json:"port"`
+	// +optional
+	Port int32 `json:"port,omitempty"`
+	// The port to route to referenced by name
+	// +optional
```
jkremser, line 37:
```suggestion
	// +kubebuilder:validation:XValidation:rule=!has(self.Port) && has(self.PortName)
```

operator/apis/http/v1alpha1/httpscaledobject_types.go
```diff
@@ -31,7 +31,11 @@ type ScaleTargetRef struct {
 	// The name of the service to route to
 	Service string `json:"service"`
 	// The port to route to
-	Port int32 `json:"port"`
+	// +optional
```
jkremser, line 34:
```suggestion
	// +kubebuilder:validation:XValidation:rule=!has(self.PortName) && has(self.Port)
```

jkremser (approved):
👍 maybe they should be mutually exclusive, can be done using [x-kubernetes-validations](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation-rules). If not, we can check that at least one of them is set, also by these X rules. Something like:

```
// +kubebuilder:validation:XValidation:rule=(has(self.PortName) || has(self.Port))
```

zroubalik (commented):
Looking good, e2e test would be nice

wozniakjan:
> maybe they should be mutually exclusive, can be done using x-kubernetes-validations.

This is neat! I will explore it a bit more and definitely find a place for this in my toolset :)

config/interceptor/role.yaml
```diff
@@ -12,6 +12,14 @@ rules:
...
   - list
   - watch
+- apiGroups:
+  - ""
+  resources:
+  - services
+  verbs:
+  - get
+  - list
+  - watch
```
wozniakjan, line 22: highlighting that newly `services` RBAC will be necessary for the `interceptor`

PR #1183, opened 2024-10-28
Add support for http header routing

### LINKED ISSUE #1177: Http header routing pattern
### Proposal

Would like to use a HTTP header in the HTTP request to determine which service to route to using [http header based routing pattern](https://docs.aws.amazon.com/prescriptive-guidance/latest/cloud-design-patterns/api-routing-http.html).

### Use-Case

Imagine hundreds of customers each with 100s of different domains pointed to their own HttpScaledObject. 

Rather than maintaining the 100 domains on `HTTPScaledObject`, you send a custom header `X-Customer-Id: customer-id-1` and register the hosts as `customer-id-1`, `customer-id-2`.

This means your upstream can add/update/delete host names without needing to update HTTPScaledObjects or extra k8s ingress/services.

Yes

### Anything else?

Can see this being implemented like
- An env variable on the interceptor, `KEDA_HTTP_ADDTL_ROUTING_HEADER=X-Customer-Id` when blank or missing from the request, it still uses the HTTP Host header

This gives you an easy way to opt into the feature and have fallback/main site domains.
```yaml
kind: HTTPScaledObject
spec:
  hosts:
  - my-main-site.example.com
  - customer-id-1
  replicas:
    max: 1
    min: 0
```

While your remaining domains might look like
```yaml
- custom-domain-1.com
- ...
- custom-domain-99.com
- *.svc.domain.com
```

gjreasoner:
Adds support for https://github.com/kedacore/http-add-on/issues/1177, [http header based routing pattern](https://docs.aws.amazon.com/prescriptive-guidance/latest/cloud-design-patterns/api-routing-http.html).

### Checklist

  - [`README.md`](/README.md)

Fixes #1177


wozniakjan (commented):
thank you for contributing!

Before we move forward, I'd like to propose an alternative path for configuring the headers routing as part of the `HTTPScaledObject` CRD. Imho it would allow more flexibility if both, the header name as well as header value, could be configured together, similar to how rule matches are defined in `HTTPRoutes` in [Gateway API](https://gateway-api.sigs.k8s.io/api-types/httproute/#matches)

For example:
```yaml
kind: HTTPScaledObject
spec:
  hosts:
  - my.domain.com
  - my2.domain.com
  pathPrefixes:
  - /root
  - /new-feature
  headers:
  - name: x-header-test
    value: abc
  - name: x-header-test2
    value: def
```

xref https://github.com/kedacore/http-add-on/issues/1177#issuecomment-2482415519

PR #1212, opened 2024-12-11
Add support for initialCooldownPeriod On httpScaledObjects

### LINKED ISSUE #1213: Add support for InitialCooldownPeriod
### Proposal

Add support for the scaled object generated by the HTTPScaledObject to have it's initialCooldownPeriod defined.

Add the field `InitialCooldownPeriod` to the httpScaledObject spec.

The generated ScaledObject should have the same value.

Opened a [PR](https://github.com/kedacore/http-add-on/pull/1212)  to add this functionality.

### Use-Case

In our use case we use the HTTP addon to deploy sandbox environments for developers. and we use the http scaler to scaled them to 0 when no requets are happening for some time to reduce costs. however without this option, the users have to first make some requests for the sandbox to be available after the deployment.
with this feature the replicas can be up for some time right after the deployment.

Yes

### Anything else?

_No response_

shayrybak:
Add support for initialCooldownPeriod On httpScaledObjects

### Checklist

  - [`README.md`](/README.md)

Fixes #1213

### DIFF
--- a/operator/apis/http/v1alpha1/httpscaledobject_types.go
+++ b/operator/apis/http/v1alpha1/httpscaledobject_types.go
@@ -104,7 +104,7 @@ type HTTPScaledObjectSpec struct {
 	CooldownPeriod *int32 `json:"scaledownPeriod,omitempty" description:"Cooldown period (seconds) for resources to scale down (Default 300)"`
 	// (optional) Initial period before scaling
 	// +optional
-	InitialCooldownPeriod int32 `json:"initialCooldownPeriod,omitempty" description:"Initial period (seconds) before scaling (Default 0)"`
+	InitialCooldownPeriod *int32 `json:"initialCooldownPeriod,omitempty" description:"Initial period (seconds) before scaling (Default 0)"`
 	// (optional) Configuration for the metric used for scaling
 	// +optional
 	ScalingMetric *ScalingMetricSpec `json:"scalingMetric,omitempty" description:"Configuration for the metric used for scaling. If empty 'concurrency' will be used"`
--- a/operator/apis/http/v1alpha1/httpscaledobject_types.go
+++ b/operator/apis/http/v1alpha1/httpscaledobject_types.go
@@ -102,6 +102,9 @@ type HTTPScaledObjectSpec struct {
 	// (optional) Cooldown period value
 	// +optional
 	CooldownPeriod *int32 `json:"scaledownPeriod,omitempty" description:"Cooldown period (seconds) for resources to scale down (Default 300)"`
+	// (optional) Initial period before scaling
+	// +optional
+	InitialCooldownPeriod int32 `json:"initialCooldownPeriod,omitempty" description:"Initial period (seconds) before scaling (Default 0)"`
 	// (optional) Configuration for the metric used for scaling
 	// +optional
 	ScalingMetric *ScalingMetricSpec `json:"scalingMetric,omitempty" description:"Configuration for the metric used for scaling. If empty 'concurrency' will be used"`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ This changelog keeps track of work items that have been completed and are ready 
 
 - **General**: TODO ([#TODO](https://github.com/kedacore/http-add-on/issues/TODO))
 - **General**: Fix kubectl active printcolumn ([#1211](https://github.com/kedacore/http-add-on/issues/1211))
+- **General**: Support InitialCooldownPeriod for HTTPScaledObject [#1213](https://github.com/kedacore/http-add-on/issues/1213)
 
 ### Improvements
 
--- a/config/crd/bases/http.keda.sh_httpscaledobjects.yaml
+++ b/config/crd/bases/http.keda.sh_httpscaledobjects.yaml
@@ -69,6 +69,10 @@ spec:
                 items:
                   type: string
                 type: array
+              initialCooldownPeriod:
+                description: (optional) Initial period before scaling
+                format: int32
+                type: integer
               pathPrefixes:
                 description: |-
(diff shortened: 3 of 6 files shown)


wozniakjan (commented):
lgtm, thank you! Can you please also re-generate the CRDs?
```
make manifests
```

shayrybak:
@wozniakjan added, I always forget that part.

wozniakjan:
perfect, thank you. One more thing, can you please sign the last commit?

zroubalik:
Should we really do this? We could easily end up in 1-1 relation with ScaledObject's field, I wonder whether we should users encourage to specify this on the SO directly 🤔

wozniakjan:
> Should we really do this?

given this can affect scaling right after `SO` creation, before metrics had a chance to get to KEDA and can lead to an immediate scale to 0 on an active deployment that was just onboarded to autoscaling, I think there is a stronger case for having `initialCooldownPeriod` in `HSO` than `cooldownPeriod` which is already there. The `cooldownPeriod` could be easily changed after `HSO` creation directly in `SO` without any negative effects.

shayrybak:
updated.`
@zroubalik this is what we've done temporarily to work around this, but it seems like a lot of effort for such a small change. I understand the use case to have my own SO if I mix other scaler tiriggers, but for a simple use case it's nice to have this.

Like I said in the issue, the use case is starting up sandbox environments so it's nice to have them up before they automatically scale down based on usage.

zroubalik:
Makes sense, no objections from my side.

operator/apis/http/v1alpha1/httpscaledobject_types.go
```diff
@@ -102,6 +102,9 @@ type HTTPScaledObjectSpec struct {
 	// (optional) Cooldown period value
 	// +optional
 	CooldownPeriod *int32 `json:"scaledownPeriod,omitempty" description:"Cooldown period (seconds) for resources to scale down (Default 300)"`
+	// (optional) Initial period before scaling
+	// +optional
+	InitialCooldownPeriod int32 `json:"initialCooldownPeriod,omitempty" description:"Initial period (seconds) before scaling (Default 0)"`
```
zroubalik, line 107:
```suggestion
	InitialCooldownPeriod *int32 `json:"initialCooldownPeriod,omitempty" description:"Initial period (seconds) before scaling (Default 0)"`
```
I think this is better

shayrybak: I can change this, however the scaledObject parameter is not a pointer, which means we'll have to add a nil check.
let me know if you think it's worth.

zroubalik: I see, that was an oversight on the upstream KEDA. Let's use pointers here, I will see if I can safely change the upstream as well.

wozniakjan: yeah, that seems to be correct. In KEDA the [`CooldownPeriod`](
https://github.com/kedacore/keda/blob/27c99dc20a18f85295ef2b32f33534d7545cdd33/apis/keda/v1alpha1/scaledobject_types.go#L91) is a pointer while [`InitialCooldownPeriod`](https://github.com/kedacore/keda/blob/27c99dc20a18f85295ef2b32f33534d7545cdd33/apis/keda/v1alpha1/scaledobject_types.go#L105) is not a pointer. I think it's because default for `CooldownPeriod` is `300` and user setting `0` must be distinguishable from user not setting anything and expecting KEDA to set the default. For `InitialCooldownPeriod`, the default is `0` so setting `nil` would be semantically equivalent.

I'd actually vote for keeping the API consistent with KEDA, `CooldownPeriod` as a pointer and `InitialCooldownPeriod` not a pointer.

shayrybak: I'm fine either way, just let me know what to do

zroubalik: @wozniakjan I am actually seriously considering changing KEDA api on this. Optional fields should be imho pointers.

@shayrybak will let you know

zroubalik: Tracking here: https://github.com/kedacore/keda/issues/6423

zero101010:
> Should we really do this? We could easily end up in 1-1 relation with ScaledObject's field, I wonder whether we should users encourage to specify this on the SO directly 🤔

Yes. it's really useful. Yesterday I need to do a workaround to use this. I'm really excited to have this on keda.

zroubalik:
> Yesterday I need to do a workaround to use this

Just out of curiosity, what workaround that was?

operator/apis/http/v1alpha1/httpscaledobject_types.go
shayrybak: @zroubalik I see the upstream was merged. should I update this one accordingly? do we want to set default values here or just pass whatever we get to the scaled object?

zroubalik: Yes, please also use pointer here and I think that we should keep the defaults to the core, CC @wozniakjan

shayrybak: ok, we will probably have to wait for kedacore to cut a release version we can reference with the new type. unless you want to point to a master commit.

shayrybak: I've updated the code for now.

JorTurFer (approved):

zero101010:
> > Yesterday I need to do a workaround to use this
> 
> Just out of curiosity, what workaround that was?

I created the httpScaleObject and patch the scaleObject to add initalCooldownPeriod and then apply the other resources.

PR #1245, opened 2025-02-04
Fix KEDA Interceptor/Scaler issues when deleting ScaledObject while scaling

kmd922:
In the current KEDA logic, if a ScaledObject is deleted while a request is in transit, then `Memory.RemoveKey()` is called **before** `Memory.Decrease()`. This causes two issues:

1. Once `Memory.RemoveKey()` is called, calling `Memory.Decrease()` will cause the keys of `Memory.concurrentMap` and `Memory.rpsMap` to diverge, leading to an error [here](https://github.com/kedacore/http-add-on/blob/ddbb17ce0b902cbb9134c3811a97d9a1efa61fee/pkg/queue/queue.go#L126).
2. Calling `Memory.Decrease()` will cause the values in `Memory.concurrentMap` to drop below 0, leading to an error [here](https://github.com/kedacore/http-add-on/blob/ddbb17ce0b902cbb9134c3811a97d9a1efa61fee/scaler/handlers.go#L253).

This PR changes the logic so that when `Memory.Decrease()` is called, we check if the key is already deleted from `Memory.concurrentMap` - if so, don't remove anything from `Memory.concurrentMap`. 

### Checklist

  - [`README.md`](/README.md)

### DIFF
--- a/pkg/queue/queue.go
+++ b/pkg/queue/queue.go
@@ -76,7 +76,25 @@ func (r *Memory) Increase(host string, delta int) error {
 func (r *Memory) Decrease(host string, delta int) error {
 	r.mut.Lock()
 	defer r.mut.Unlock()
-	r.concurrentMap[host] -= delta
+
+	current, exists := r.concurrentMap[host]
+	if !exists {
+		// Key doesn't exist; nothing to do
+		return nil
+	}
+
+	// Decrement and clamp concurrency to zero
+	newVal := current - delta
+	if newVal < 0 {
+		newVal = 0
+	}
+	r.concurrentMap[host] = newVal
+
+	// Update rpsMap if the key exists
+	if rpsItem, ok := r.rpsMap[host]; ok {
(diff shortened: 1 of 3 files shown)


pkg/queue/queue.go
```diff
@@ -76,7 +76,25 @@ func (r *Memory) Increase(host string, delta int) error {
...
+	newVal := current - delta
+	if newVal < 0 {
+		newVal = 0
+	}
+	r.concurrentMap[host] = newVal
+
+	// Update rpsMap if the key exists
+	if rpsItem, ok := r.rpsMap[host]; ok {
+		rpsItem.Record(time.Now(), -delta)
+	}
```
JorTurFer, line 96: As RPS calculation works bucketing the request over the time, decreasing the counter will break the correct calculation (it relies on a circular buffer to count the requests). Please remove this code updating the `rpsMap`

JorTurFer (approved):

PR #1252, opened 2025-02-11
Make PollingInterval configurable in HTTPScaledObject

### LINKED ISSUE #1251: Add PollingInterval configuration to HTTPScaledObject
### Proposal

The ScaledObject has a PollingInterval configuration, which managed how often the scaling metric is polled.
The HTTPScaledObject uses a constant hardcoded value of 15 seconds: https://github.com/kedacore/http-add-on/blob/main/pkg/k8s/scaledobject.go#L12

We wish for that PollingInterval to be configurable in the HTTPScaledObject.

### Use-Case

A polling interval of 15 seconds isn't perfect for every use case. Instead of trying to find a magic number that works for most, exposing it as config, allows everyone to set their own polling interval.

Maybe

### Anything else?

There was the same feature request in the past (closed due to inactivity): https://github.com/kedacore/http-add-on/issues/1184

Spenhouet:
Provide the PollingInterval of the ScaledObject as parameter in the HTTPScaledObject.

### Checklist

  - [`README.md`](/README.md)

Fixes #1251

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ This changelog keeps track of work items that have been completed and are ready 
 
 - **General**: TODO ([#TODO](https://github.com/kedacore/http-add-on/issues/TODO))
 - **General**: Fix kubectl active printcolumn ([#1211](https://github.com/kedacore/http-add-on/issues/1211))
+- **General**: Allow configuration of PollingInterval in HTTPScaledObject ([#1251](https://github.com/kedacore/http-add-on/issues/1251))
 
 ### Improvements
 
--- a/config/crd/bases/http.keda.sh_httpscaledobjects.yaml
+++ b/config/crd/bases/http.keda.sh_httpscaledobjects.yaml
@@ -78,6 +78,11 @@ spec:
                 items:
                   type: string
                 type: array
+              pollingInterval:
+                description: (optional) Configuration for the polling interval of
+                  the scaling metric
+                format: int32
+                type: integer
               replicas:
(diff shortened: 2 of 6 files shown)


wozniakjan (approved):
lgtm, thank you!

ptal @zroubalik, I know you are not a big fan of duplicating fields between `HTTPScaledObject` and `ScaledObject` but imho this one also makes sense because the operator sets this to a hardcoded 15s and patches over any custom changes to the `ScaledObject` made by the user

zroubalik (commented):
Alternate proposal: 


What if we just remove the hardcoded 15s and let it on ScaledObject?

This project is still beta so backwards compatibility is something we can ingore for certain cases and I think this one is a good one.

Spenhouet:
@JorTurFer Maybe I'm misunderstanding something, but our goal is to reduce the time of the polling interval to ensure fast scaling. Right now we solved it by not using the HTTPScaledObject and instead directly using a ScaledObject, which allows to set the polling interval. Adding this config enables us to remove that workaround.

JorTurFer:
My point is that I'm not sure if you're actually getting an advantage in the scaling or it's just a useless parameter. 
For the scaling phase (1-N-1), the querying period is every 15 seconds as default (HPA default value configured at control plane level) and for the activation phase (0-1-0) I'm 99% sure that it doesn't apply because HTTP Add-on uses `external-push` scaler, that actively monitors the `active` status, scaling from 0 as soon as the http-add-on external scaler detects a single request

PR #1266, opened 2025-03-09
Change path in walkthrough call to match httpscaledobject from example

GilbertBckr:
Changed the curl command in the walktrough to match the path of httpscaledobject which was previously created,
following the walktrough guide (see [examples/v0.10.0/httpscaledobject.yaml](https://github.com/kedacore/http-add-on/blob/main/examples/v0.10.0/httpscaledobject.yaml))

Previously the curl command would request the service on ```/path``` now it uses the ```/test``` path.

I previously followed the walkthrough and wondered why my setup did not work until I noticed this problem.


zroubalik (approved):
LGTM, thanks!

PR #1222, opened 2024-12-17
Add support for http header routing

erich23:
_Provide a description of what has been changed_

### Checklist

- [ ] Commits are signed with Developer Certificate of Origin (DCO)
- [ ] Changelog has been updated and is aligned with our [changelog requirements](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#Changelog)
- [ ] Any necessary documentation is added, such as:
  - [`README.md`](/README.md)
  - [The `docs/` directory](./docs)
  - [The docs repo](https://github.com/kedacore/keda-docs)

Fixes #

PR #1289, opened 2025-05-07
[WIP] interceptor: support for HTTP header routing

### LINKED ISSUE #1177: Http header routing pattern

wozniakjan:
wip, credits to https://github.com/kedacore/http-add-on/pull/1222

### Checklist

  - [`README.md`](/README.md)

Fixes #1177


wozniakjan:
> I've been waiting for this PR!

soon @kahirokunn :)

operator/apis/http/v1alpha1/httpscaledobject_types.go
```diff
@@ -132,6 +132,12 @@ type HTTPScaledObjectTimeoutsSpec struct {
 	ResponseHeader metav1.Duration `json:"responseHeader" description:"How long to wait between when the HTTP request is sent to the backing app and when response headers need to arrive"`
 }
 
+// Header contains the definition for matching on header name and/or value
+type Header struct {
+	Name  string `json:"name"`
```
linkvt: We could add `+kubebuilder:validation:MinLength=1` here (hope that it's correct, please check :) )

pkg/routing/tablememory.go
```diff
@@ -28,42 +31,150 @@ func (tm *TableMemory) Remember(httpso *httpv1alpha1.HTTPScaledObject) *TableMem
...
+		}
+		// then by specificity of header values, more specific first
+		specificityI, specificityJ := specificityOfHeaders(hi), specificityOfHeaders(hj)
+		if specificityI != specificityJ {
+			return specificityI > specificityJ
+		}
+
+		// then by creation timestamp, older first
+		if !hi.CreationTimestamp.Time.Equal(hj.CreationTimestamp.Time) {
+			return hj.CreationTimestamp.After(hi.CreationTimestamp.Time)
```
linkvt: Could we add a test for that? I think there is already one for the case without headers, just a bit hard to wrap my head around whether Before or After here is correct.

wozniakjan:
> I usually leave more than less comments, some might be subjective but I'd rather have a you disagree on something that missing something we both would have changed 🙂

yeah, all good input! Changes in https://github.com/kedacore/http-add-on/pull/1289/changes/b817396cc3d595178afddbe8487a4823eb165c17. I also clicked on "resolve" for all where I thought my changes were in line with your proposal. The only remaining is https://github.com/kedacore/http-add-on/pull/1289#discussion_r2708258975. Not that I would disagree but I have been punting on improving this piece without refactoring a larger part of the codebase.

pkg/routing/tablememory.go
```diff
@@ -28,42 +31,150 @@ func (tm *TableMemory) Remember(httpso *httpv1alpha1.HTTPScaledObject) *TableMem
...
 	for _, key := range keys {
-		newStore, oldHTTPSO, _ := store.Insert(key, httpso)
-
-		// oldest HTTPScaledObject has precedence
-		if oldHTTPSO != nil && httpso.GetCreationTimestamp().After(oldHTTPSO.GetCreationTimestamp().Time) {
-			continue
-		}
-
-		store = newStore
+		store = remember(store, key, httpso)
```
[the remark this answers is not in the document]
linkvt: Ah I missed that Remember returns a new tablememory, ok then this makes sense of course. We could change TableMemory to update itself instead of returning a new one, tried that locally real quick but thats probably also subjective, both is fine for me.

Name is also fine for me, also couldn't come up with something better :)

linkvt (approved):
Nice!

PR #1300, opened 2025-05-21
e2e: fix nginx installation via helm

zroubalik:
There's an issue in Helm causing problems in e2e tests: https://github.com/helm/helm/issues/30878

```
* Deployment.apps "ingress-controller" is invalid: spec.progressDeadlineSeconds: Invalid value: 0: must be greater than minReadySeconds
```

### DIFF
--- a/tests/utils/setup_test.go
+++ b/tests/utils/setup_test.go
@@ -176,7 +176,7 @@ func TestSetupIngress(t *testing.T) {
 	_, err = ExecuteCommand("helm repo update ingress-nginx")
 	require.NoErrorf(t, err, "cannot update ingress-nginx helm repo - %s", err)
 
-	_, err = ExecuteCommand(fmt.Sprintf("helm upgrade --install %s ingress-nginx/ingress-nginx --set fullnameOverride=%s --set controller.service.type=ClusterIP --namespace %s --wait",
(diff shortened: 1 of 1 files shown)


tests/utils/setup_test.go
```diff
@@ -176,7 +176,7 @@ func TestSetupIngress(t *testing.T) {
 	_, err = ExecuteCommand("helm repo update ingress-nginx")
 	require.NoErrorf(t, err, "cannot update ingress-nginx helm repo - %s", err)
 
-	_, err = ExecuteCommand(fmt.Sprintf("helm upgrade --install %s ingress-nginx/ingress-nginx --set fullnameOverride=%s --set controller.service.type=ClusterIP --namespace %s --wait",
+	_, err = ExecuteCommand(fmt.Sprintf("helm upgrade --install %s ingress-nginx/ingress-nginx --set fullnameOverride=%s --set controller.service.type=ClusterIP --set controller.progressDeadlineSeconds=0 --namespace %s --wait",
```
wozniakjan, line 179: I think this would result in the exact same rendering, shouldn't this be a different value or `null`?
```suggestion
	_, err = ExecuteCommand(fmt.Sprintf("helm upgrade --install %s ingress-nginx/ingress-nginx --set fullnameOverride=%s --set controller.service.type=ClusterIP --set controller.progressDeadlineSeconds=null --namespace %s --wait",
```

wozniakjan (approved):

PR #1389, opened 2025-12-04
ci: use GitHub-hosted ARM64 runners for e2e tests

### LINKED ISSUE #1388: Migrate ARM64 e2e tests to GitHub-hosted runners
GitHub offers free ARM64 runners (`ubuntu-24.04-arm`) for public repositories since August last year (see reference below).
We can replace the self-hosted ARM64 runner infrastructure and simplify the e2e workflow.

### Current state

The `e2e-tests.yaml` workflow has 3 jobs:
- `e2e_tests` - AMD64 on `ubuntu-latest`
- `arm_image_generation` - builds images on self-hosted `ARM64` runner
- `arm_e2e_tests` - runs tests on self-hosted `http-add-on-e2e` runner

### Benefits

- No self-hosted runner maintenance
- Kind images are multi-arch, so the same workflow steps work on both architectures

### References

- https://github.blog/changelog/2025-08-07-arm64-hosted-runners-for-public-repositories-are-now-generally-available/
- https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/

I'll create a PR to test this.

linkvt:
Fixes #1388

Replace self-hosted `ARM64` and `http-add-on-e2e` runners with GitHub's free `ubuntu-24.04-arm` runner.
GitHub announcement: https://github.blog/changelog/2025-08-07-arm64-hosted-runners-for-public-repositories-are-now-generally-available/

Not sure if we want to include such changes in the CHANGELOG as they aren't really relevant for the end user?
Also: Do we need some kind of cleanup of the existing ARM runners?

## Changes

- Add architecture matrix dimension with their images (`ubuntu-latest`, `ubuntu-24.04-arm`)
- Delete ARM specific e2e-test jobs
- This should also fix ARM64 image build flakes like this: https://github.com/kedacore/http-add-on/actions/runs/19926374893/job/57127087014?pr=1387

## Checklist

  - [`README.md`](../README.md)

### DIFF
--- a/tests/checks/interceptor_websocket/interceptor_websocket_test.go
+++ b/tests/checks/interceptor_websocket/interceptor_websocket_test.go
@@ -184,7 +184,8 @@ spec:
     spec:
       containers:
       - name: websocat-test
-        image: ghcr.io/vi/websocat:v1.14.0
+        # v1.14.0 was amd64-only; :latest includes arm64 (https://github.com/vi/websocat/pull/283)
+        image: ghcr.io/vi/websocat:latest
         command: ["/bin/sh"]
         args:
         - -c
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -14,19 +14,23 @@ permissions:
 
 jobs:
   e2e_tests:
-    runs-on: ubuntu-latest
-    name: Execute e2e test on AMD64 ${{ matrix.kubernetesVersion }}
+    runs-on: ${{ matrix.runner }}
+    name: e2e ${{ matrix.arch }} / ${{ matrix.kindVersion }}
     strategy:
       fail-fast: false
       matrix:
-        kubernetesVersion: [v1.33, v1.32, v1.31]
+        arch:
+          - AMD64
+          - ARM64
+        kindVersion:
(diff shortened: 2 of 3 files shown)


tests/checks/interceptor_websocket/interceptor_websocket_test.go
```diff
@@ -184,7 +184,8 @@ spec:
     spec:
       containers:
       - name: websocat-test
-        image: ghcr.io/vi/websocat:v1.14.0
+        # v1.14.0 was amd64-only; :latest includes arm64 (https://github.com/vi/websocat/pull/283)
+        image: ghcr.io/vi/websocat:latest
```
JorTurFer: wasn't this already solved @wozniakjan ?

JorTurFer (approved):
Nice contribution!
Once this is mergd, we have to update the merge checks to require the new ones

tests/checks/interceptor_websocket/interceptor_websocket_test.go
JorTurFer: do you think that curl could be an option? it'd be nice actually. Less dependencies is always better

PR #1411, opened 2026-01-15
feat: add wildcard host matching for HTTPScaledObject

### LINKED ISSUE #281: Wildcard support for multi-tenant interceptor
Allow for wildcards when specifying Host headers and routing information for interceptor and scaler.

### Use-Case

Some services might have multiple hostnames, say for multiple customers in the form of "customername.service.example.com". So supplying a single Host header for routing/scaling wont suffice. We would need to support simple wildcards (*.service.exmaple.com") at the very least for this case. It might also be desirable to have the host parameter be a list rather than a single value.

### Specification
I'm not intimately familiar with the internals of the project but my first take would be:

1. When defining the `host` in an `HTTPScaledObject` allow for wildcards (and possibly multiple values)
2. The queue used for keeping track of pending requests should match an incoming Host header with all the hosts and wildcards it knows about. 
3. Update the routing table to handle routing for wildcard HTTPScaledObject

linkvt:
Implement wildcard host matching that allows HTTPScaledObjects to use patterns like `*.example.com` to match any subdomain. When an exact host match is not found, the routing table falls back to wildcard matching.

I considered two approaches:
- `*.foo` catches `bar.foo` but not `baz.bar.foo`
	- this is the behavior of a wildcard host in an Ingress rule
- `*.foo` catches `bar.foo`, `baz.bar.foo` and all other nested domains
	- this approach is chosen by Gateway API and others -> this is the way I chose
	- Quote: `The wildcard character will match any number of characters and dots to the left, however, so "*.example.com" will match both "foo.bar.example.com" and "bar.example.com".`
	- Source: https://gateway-api.sigs.k8s.io/reference/spec/

There is also the case of no hosts being specified but this case was currently not really handled as it never matched (host is set to `""`). Kubernetes ingresses etc do interpret an empty/missing host as wildcard which is what I am also doing now.

There will also be a followup PR just with a refactor of the table memory as the `index` was only used in tests.

### Quick Benchmark

I added a benchmark (see changed files) to compare the before and after, this is the summary:
- keep in mind: these are nanoseconds
- reduced allocations and preallocating reduced the exact case from 116ns to 38ns
- the new most simple wildcard case takes now 135ns (almost the exact case duration from before)
- more complex wildcard case `a.b.c.d.e.example.com` for `*.example.com` takes 415ns as every next subdomain is tried in sequence

Linear scaling for long wildcard domains is of course not optimal but:
- this is not really common
- 0.0004ms for a long wildcard is not noticable
- there are likely more major bottlenecks we should focus on

```plain
❯ benchstat before.txt after.txt
goos: darwin
goarch: arm64
pkg: github.com/kedacore/http-add-on/pkg/routing
cpu: Apple M4 Pro
                             │  before.txt  │               after.txt               │
                             │    sec/op    │    sec/op     vs base                 │
ExactMatch/SingleRoute-14      116.70n ± 1%   38.60n ±  1%   -66.92% (p=0.000 n=10)
ExactMatch/Among100Routes-14   139.40n ± 1%   69.41n ±  0%   -50.21% (p=0.000 n=10)
Wildcard/1Subdomain-14          111.8n ± 1%   135.2n ±  1%   +20.97% (p=0.000 n=10)
Wildcard/3Subdomains-14         111.8n ± 1%   279.6n ±  1%  +150.09% (p=0.000 n=10)
Wildcard/5Subdomains-14         113.8n ± 1%   415.2n ±  2%  +264.81% (p=0.000 n=10)
Wildcard/Among100Routes-14      116.1n ± 0%   134.6n ±  1%   +15.93% (p=0.000 n=10)
CatchAll/SingleRoute-14         112.8n ± 1%   231.7n ±  1%  +105.32% (p=0.000 n=10)
CatchAll/Among100Routes-14      116.2n ± 2%   237.5n ±  1%  +104.43% (p=0.000 n=10)
NoMatch/SingleRoute-14          112.3n ± 1%   229.4n ± 22%  +104.37% (p=0.000 n=10)
NoMatch/Among100Routes-14       112.7n ± 1%   256.0n ± 16%  +127.15% (p=0.000 n=10)
geomean                         116.1n        168.4n         +44.99%
```

### Changes 
- Update Route() to try wildcard match when exact match fails
- Simplify key format from "//{host}/{path}/" to "{host}/{path}/"
  - I found no reason for the `//` prefix to exist, all tests also still pass. Seems like this was the most simple key received after removing the protocol
- Update docs to reference wildcard usage

### Checklist

  - [`README.md`](../README.md)

Fixes #281

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ This changelog keeps track of work items that have been completed and are ready 
 
 ### New
 
+- **General**: Add multi-level wildcard host matching for HTTPScaledObject routing ([#281](https://github.com/kedacore/http-add-on/issues/281))
 - **General**: Add environment variables for leader election timing configuration ([#1365](https://github.com/kedacore/http-add-on/pull/1365))
 
 ### Improvements
--- a/docs/ref/vX.X.X/http_scaled_object.md
+++ b/docs/ref/vX.X.X/http_scaled_object.md
@@ -1,6 +1,6 @@
 # The `HTTPScaledObject`
 
->This document reflects the specification of the `HTTPScaledObject` resource for the `vX.X.X` version.
+> This document reflects the specification of the `HTTPScaledObject` resource for the `vX.X.X` version.
 
 Each `HTTPScaledObject` looks approximately like the below:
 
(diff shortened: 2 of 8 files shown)


linkvt:
@JorTurFer I noticed that your approve stayed on this PR despite me pushing changes. What do you think of resetting approvals on pushes to prevent accidental merges of added unreviewed changes?

PR #1418, opened 2026-01-19
Bump golang to 1.25.5 + small fixes

JorTurFer:
As part of this PR:

- (CI) golangci-lint step has been replaced by pre-commit step which does lint and extra checks (and lint version is aligned with makefile)
- bump golangci-lint to 2.7.1
- exclude some linting issues
- fix changelog issue
- migrate pre-commit config

### Checklist


linkvt:
I quickly checked and also had a look at the KEDA repo, it seems there are some dependencies that specify 1.25.0 which forces KEDA to specify the patch version. KEDA does specify 1.25.5 though requiring us to also specify 1.25.5.

1.25.0 seems to come from:
- https://github.com/kubernetes-sigs/controller-runtime/blob/main/tools/setup-envtest/go.mod#L3
- https://github.com/kubernetes-sigs/controller-tools/blob/main/go.mod#L3
- https://github.com/kubernetes/kms/blob/master/go.mod#L5

PR #1434, opened 2026-02-02
fix(interceptor): preserve upstream X-Forwarded headers

### LINKED ISSUE #1432: x-forwarded-proto is not preserved in 0.12.0
### Report

Following https://github.com/kedacore/http-add-on/pull/1404 and the release of 0.12.0. I have some workload that get stuck in a redirect loop. They are behind ingress-nginx which terminates SSL and the workload relies `x-forwarded-proto` to know whether to redirect to https.

### Expected Behavior

The `x-forwarded-proto` should be preserved.

### Actual Behavior

The `x-forwarded-proto` gets overriden here https://github.com/kedacore/http-add-on/blob/main/interceptor/handler/upstream.go#L77 and ends up saying `http` and not `https` anymore

1. Install 0.12.0
2. Have an ingress with TLS enabled that terminates TLS at the entrypoint
3. `x-forwarded-proto` shows `http`

None

### HTTP Add-on Version

0.12.0

### Kubernetes Version

1.33

### Platform

Any

### Anything else?

_No response_

linkvt:
The interceptor should preserve X-Forwarded-... headers of upstream requests as these are used to indicate how the client originally connected to the infrastructure.
The goal is not to force infos about requests to the interceptor into these headers.

We always trust X-Forwarded headers as the HTTP Addon expects an ingress controller or such to protect against X-Forwarded-... headers injected from outside of the cluster.
The implementation is now more in line with others, e.g. how [traefik](https://github.com/traefik/traefik/blob/a4a91344edcdd6276c1b766ca19ee3f0e346480f/pkg/proxy/httputil/proxy.go#L122-L139) handles it.
The `X-Forwarded` (without a suffix like `-For`) header is still ignored here as before as there doesn't seem to be a real adoption also indicated by Go not handling it.

Changes:
- Preserve X-Forwarded-Proto from incoming request if present
- Preserve X-Forwarded-Host from incoming request if present

### Checklist

  - [`README.md`](../README.md)

Fixes #1432

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,7 +36,7 @@ This changelog keeps track of work items that have been completed and are ready 
 
 ### Fixes
 
-- **General**: TODO ([#TODO](https://github.com/kedacore/http-add-on/issues/TODO))
+- **General**: Preserve `X-Forwarded-Proto` and `X-Forwarded-Host` headers from upstream proxies ([#1432](https://github.com/kedacore/http-add-on/issues/1432))
 
 ### Deprecations
 
--- a/interceptor/handler/upstream.go
+++ b/interceptor/handler/upstream.go
@@ -72,9 +72,16 @@ func (uh *Upstream) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 			pr.SetURL(stream)
 			// Preserve original Host header (SetURL rewrites it by default).
 			pr.Out.Host = pr.In.Host
-			// Preserve X-Forwarded-For chain before appending client IP.
+
+			// Preserve and extend X-Forwarded-... headers from upstream proxies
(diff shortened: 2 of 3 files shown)


jocelynthode:
@dbu Probably not needed as https://pkg.go.dev/net/http/httputil#ProxyRequest.SetXForwarded only touches the 3 headers already fixed in this PR.

dbu (approved):
> As these headers are all somewhat non-standard but rather conventional I added a testcase to make sure we don't strip X-Forwarded-Port in the future.

thats the danger when working with not-really standardized headers. sounds great, that should cover all that is needed (i noticed in our infrastructure that the port header was not touched by keda, so thats good already)

thanks a lot for the work!

PR #1435, opened 2026-02-02
refactor(interceptor): extract proxy handler for testability

### LINKED ISSUE #1415: Interceptor proxy server tests use unresolvable DNS and don't verify successful forwarding
## Report

The interceptor proxy server tests (`TestRunProxyServerCountMiddleware`, `TestRunProxyServerWithTLSCountMiddleware`, `TestRunProxyServerWithMultipleCertsTLSCountMiddleware`) have structural issues that prevent them from testing actual end-to-end proxying.

The tests create HTTPScaledObjects with service names like `testservice` in namespace `test-namespace`. The routing middleware constructs stream URLs in Kubernetes DNS format (`http://testservice.test-namespace:<port>`), which doesn't resolve outside a cluster.

The forward handler fails after exhausting connection retries, but tests still pass because they expect `g.Wait()` to return an error - which the failed forward satisfies. The tests verify counting middleware behavior (+1/-1/final count) but never verify the request actually reaches the origin server.

Secondary issue: the non-TLS test uses hardcoded port 8080, causing failures when that port is in use.

## Expected Behavior

- Tests verify successful end-to-end proxying (request reaches origin, response is 200 OK)
- DNS resolution works in the test environment
- Tests use dynamic ports to avoid conflicts

## Actual Behavior

- Forward always fails (DNS doesn't resolve)
- Tests pass anyway because they expect an error
- TLS tests take ~4.2s (connection retry time) before "passing"
- Non-TLS test fails if port 8080 is in use locally

1. Run the affected tests:
   ```
   go test -v ./interceptor/... -run "TestRunProxyServerCountMiddleware|TestRunProxyServerWithTLSCountMiddleware|TestRunProxyServerWithMultipleCertsTLSCountMiddleware"
   ```
2. Observe TLS tests take ~4.2s (time spent on connection retries before failing)
3. Non-TLS test may fail if port 8080 is in use locally

---

I will work on this issue.

linkvt:
The current interceptor/main_test.go is not actually testing anything, see #1415. This PR is a refactoring to fix this by extracting the handlers so that we can use the go httptest package for testing.
I'm not a fan of "plumbing" which is why I tried to not change anything that wasn't needed or trivial hoping to improve this objectively.
Still sorry for this big PR, let me know if we should do a 1-1 where I can give an overview about the changes.

### Changes

Extract building the handler chain for better testability and refactor the interceptor/main.go file to make it cleaner.
I tried to not change any unrelated logic like the whole TLS cert construction but extracted it into a separate file and generated tests for it so that we could refactor it later.

- Extract BuildProxyHandler for shared use by production and tests
- Move TLS configuration logic to dedicated tls_config.go and add tests
- Replace main_test.go with proxy_test.go using the handler builder
- Simplify config parsing to return values instead of pointers
	- passing pointers isn't really best practice for such small structs
- Remove deprecated DeploymentCachePollIntervalMS config option

### Test Changes

Generated comparison between the previous tests and the ones added in this PR.

#### Removed: `interceptor/main_test.go` (475 lines)
Three large integration tests that tested `runProxyServer()` end-to-end with HTTP, TLS via cert path, and TLS via cert store.

#### Added: `interceptor/proxy_test.go` (394 lines)
Focused unit tests for the handler chain:
- `TestProxyHandler_SuccessfulRequest` - basic request flow
- `TestProxyHandler_ColdStartHeader` - tests enabled/disabled and true/false states
- `TestProxyHandler_Tracing` - tests enabled/disabled paths
- `TestProxyHandler_TracingPreservesTraceID` - verifies trace ID preservation
- `TestProxyHandler_BackendReceivesCorrectRequest` - method/path forwarding
- `TestProxyHandler_UnknownHostReturnsError` - 404 for unknown hosts
- `TestProxyHandler_QueueCounting` - increment/decrement/final state
- `TestProxyHandler_TLSBackend` - TLS connection to backend

#### Added: `interceptor/tls_config_test.go` (247 lines)
Comprehensive unit tests for `BuildTLSConfig()`:
- Single cert via path, cert store directories, multiple directories
- SNI-based cert selection and default fallback
- Error handling: missing key, invalid content, non-existent paths
- `.pem` format and IP address SAN support

#### Coverage Summary
| Old Behavior | New Coverage |
|--------------|--------------|
| Queue counting | `TestProxyHandler_QueueCounting` |
| Cold start header | `TestProxyHandler_ColdStartHeader` (now tests all states) |
| Tracing | `TestProxyHandler_Tracing` (now tests disabled path too) |
| TLS config | `tls_config_test.go` (13 edge cases) |
| TLS termination | Unit tests + E2E (no need to test Go's `tls.Listen`) |

### Checklist

  - [`README.md`](../README.md)

Fixes #1415


interceptor/tls_config.go
```diff
@@ -0,0 +1,159 @@
+package main
```
linkvt: I moved the TLS config building functions into a separate file and am using a struct instead of a map now to pass the arguments.

JorTurFer (approved):

PR #1445, opened 2026-02-04
Add stress tests

### LINKED ISSUE #238: End to end and stress tests
We currently have mostly unit tests inside this repository, along with a few integration tests (like one for the interceptor coming in #206). I propose that we add more testing strategies to this project, including:

- **end to end**: a test that installs the addon to a Kubernetes cluster along with KEDA and any other dependencies, installs an app and an `HTTPScaledObject`, and runs a quick suite of requests to trigger scale-from-zero, scale up, and scale back down to zero. this test should check that the app and the interceptor both scale properly, and that the external scaler, operator, and interceptor's admin endpoints all report the expected things (NOTE: KEDA does this, and there is a prototype of something similar in #206)
- **stress**: a test that runs a large number of requests concurrently against an interceptor fleet, optionally for a large period of time. also possibly inspect internal admin endpoints along the way to ensure that metrics report as expected. the main purpose of this test is to ensure that, from the outside, requests get serviced as expected

khauser:
Stress tests seems to be one essential requirement for getting the addon production ready or GA.

### Checklist

  - [`README.md`](../README.md)

Fixes #238

### DIFF
--- a/tests/helper/helper.go
+++ b/tests/helper/helper.go
@@ -145,7 +145,7 @@ func ExecCommandOnSpecificPod(t *testing.T, podName string, namespace string, co
 func WaitForSuccessfulExecCommandOnSpecificPod(t *testing.T, podName string, namespace string, command string, iterations, intervalSeconds int) (bool, string, string, error) {
 	var out, errOut string
 	var err error
-	for i := 0; i < iterations; i++ {
+	for range iterations {
 		out, errOut, err = ExecCommandOnSpecificPod(t, podName, namespace, command)
 		t.Logf("Waiting for successful execution of command on Pod; Output: %s, Error: %s", out, errOut)
 		if err == nil {
@@ -234,7 +234,7 @@ func DeleteNamespace(t *testing.T, nsName string) {
 }
 
 func WaitForJobSuccess(t *testing.T, kc *kubernetes.Clientset, jobName, namespace string, iterations, interval int) bool {
-	for i := 0; i < iterations; i++ {
+	for range iterations {
 		job, err := kc.BatchV1().Jobs(namespace).Get(context.Background(), jobName, metav1.GetOptions{})
 		if err != nil {
 			t.Logf("cannot run job - %s", err)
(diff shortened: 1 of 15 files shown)


tests/helper/helper.go
```diff
@@ -392,6 +392,26 @@ func WaitForDeploymentReplicaReadyCount(t *testing.T, kc *kubernetes.Clientset,
 	return false
 }
 
+// WaitForDeploymentReplicaReadyMinCount waits until deployment ready replicas reach at least the minimum target or number of iterations are done.
+func WaitForDeploymentReplicaReadyMinCount(t *testing.T, kc *kubernetes.Clientset, name, namespace string,
+	minTarget, iterations, intervalSeconds int) bool {
+	for i := 0; i < iterations; i++ {
```
linkvt:
```suggestion
	for range iterations {
```

nit: could be modernized

JorTurFer (commented):
I like this PR!!! ❤️ 

My only question is if we should use the on-prem runner for stress workflow instead of a GH runner to have more resources available (IIRC, machine is like 16CPU and 20GB). The problem is that it's only for ARM64 but I think that it's fine.

If you agree with using it, I can spin up another runner (or 2) just to ensure that only a fixed amount of tests are executed at the same time.

linkvt (changes requested):
Thanks for the PR! I think it is definitely a good idea to add reproducible stress tests and agree with Jorge.
I left a few comments, interested in what you think about them.

I was also wondering if we check e.g. whether the requests were successful in case the HTTP Addon just returns 404 or so?

Also noticed that you chose Apache AB, we could IMO also consider more modern tools like https://github.com/hatoo/oha (also provides a container image) as the ab docs are not even that positive about ab, see https://httpd.apache.org/docs/current/programs/ab.html#bugs (no complete http/1.x support, no http2 support, `"i.e., you would measure the ab performance rather than the server's"`)  😀

khauser:
@JorTurFer: In my opinion stress or performance tests should run once before a release and on demand because of their longer runtime. There should also be a gate or threshold which must not exceed.

tests/helper/helper.go
khauser: Sure

tests/stress-run-all.go
```diff
@@ -0,0 +1,277 @@
+//go:build stress
```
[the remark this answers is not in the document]
linkvt: @JorTurFer do you have an opinion on this? I noticed keda does something similar but think that my comment still somewhat applies considering everything is in one package.

linkvt:
I used this PR as a base for writing #1675 that adds stress tests using the new e2e test framework making this PR obsolete.

Thanks for the contribution! ❤️

Issue #1449, opened 2026-02-05
KEDA_CONDITION_WAIT_TIMEOUT not respected after 0.12.0


rd-zahari-aleksiev:
### Report

before updating to 0.12.0 the connection was cut of at the 30s, because KEDA_CONDITION_WAIT_TIMEOUT is set to 30s
After updating though it is cutting of at 20s, so it seems to be something has changed and there is a regression.

Is it only me, or this is reproducible? 

### Expected Behavior

Bad Gateway to be returned at ~30s mark

### Actual Behavior

the connection cut off at 20s mark

<img width="321" height="18" alt="Image" src="https://github.com/user-attachments/assets/ce5aa7e9-d5d2-45eb-835a-5d8627594451" />

### Steps to Reproduce the Problem

1. set
- name: KEDA_RESPONSE_HEADER_TIMEOUT
  value: 30s
- name: KEDA_CONDITION_WAIT_TIMEOUT
  value: 30s

Scale something slow to start from zero on http request

### Logs from KEDA HTTP Add-on

No logs from the interceptor... but I can see that the timeouts are parsed when starting.
```
2026-02-04T15:28:07Z	INFO	setup	starting interceptor	{"timeoutConfig": {"Connect":1000000000,"KeepAlive":1000000000,"ResponseHeader":30000000000,"WorkloadReplicas":30000000000,"ForceHTTP2":false,"MaxIdleConns":100,"MaxIdleConnsPerHost":20,"IdleConnTimeout":90000000000,"TLSHandshakeTimeout":10000000000,"ExpectContinueTimeout":2000000000}, "servingConfig": {"CurrentNamespace":"keda","WatchNamespace":"","ProxyPort":8080,"AdminPort":9090,"ConfigMapCacheRsyncPeriod":3600000000000,"DeploymentCachePollIntervalMS":250,"EndpointsCachePollIntervalMS":250,"ProxyTLSEnabled":false,"TLSCertPath":"/certs/tls.crt","TLSKeyPath":"/certs/tls.key","TLSCertStorePaths":"","TLSSkipVerify":false,"TLSPort":8443,"ProfilingAddr":"","EnableColdStartHeader":true,"LogRequests":false}, "metricsConfig": {"OtelPrometheusExporterEnabled":true,"OtelPrometheusExporterPort":2223,"OtelHTTPExporterEnabled":false}}
```

### HTTP Add-on Version

0.12.0

### Kubernetes Version

1.33

### Platform

Google Cloud

[...]

PR #1461, opened 2026-02-11
fix: rework dial retry to tolerate slow Service IP propagation

linkvt:
### Problem Description

When a Pod becomes Ready the Endpoint Slice is updated by K8s to contain the new ready addresses.
Every node now needs to be updated so that the ClusterIP of the Service points to the ready addresses from the EndpointSlice which could take up to 10 seconds in GKE clusters using Dataplane v1 (Dataplane v2 seems to use cilium which watches k8s for events and updates the eBPF maps quite immediately after an EndpointSlice was updated).

Our dial retry implementation likely hits this issue as we use the Service ClusterIP as soon as there are EndpointSlices with ready addresses and tries to connect to the ClusterIP - the actual routes from ClusterIP to Pod IP might not be configured yet though.

The exponential backoff in the dial retry logic results in an effective retry window of only ~2.5s, causing premature 502s in environments where Service IP propagation is slow (e.g. in GKE with `iptables-min-sync-period=10s` when using Dataplane V1).

The behavior around this changed in PR #1387 but was also not ideal before that, as the HTTP Addon waited for `KEDA_HTTP_CONNECT_TIMEOUT` before retrying, which was especially bad with high values on cold starts - more infos in the description of that PR.

knative does not hit this issue as they connect to the Pod IPs directly which is a great idea we could adopt in the future as well, will create a feature request for that.

### Solution 

This PR replaces the retry backoff with a simple fixed-interval retry loop bounded by a configurable timeout (KEDA_HTTP_DIAL_RETRY_TIMEOUT, default 15s). A short fixed interval (50ms) keeps latency low when the target becomes reachable.
A retry backoff makes only sense on timeouts if the target is slow to respond and not on refused connections which are immediate.
Right now this is fixed at `KEDA_HTTP_CONNECT_TIMEOUT=500ms`, knative does a backoff there but also has a constant sleep for retry delay, see https://github.com/knative/serving/blob/b929580ceffda5c1f92b2dd7f29bbedf5a56a042/vendor/knative.dev/pkg/network/transports.go#L109 .

This results in the following timeouts we can configure with the last one being new:
- **KEDA_HTTP_CONNECT_TIMEOUT** (500ms): Per-attempt TCP dial timeout — passed to `net.Dialer.Timeout`
- **KEDA_HTTP_KEEP_ALIVE** (1s): TCP keepalive probe interval — passed to `net.Dialer.KeepAlive`
- **KEDA_RESPONSE_HEADER_TIMEOUT** (500ms): Max wait for response headers from backend — set on `http.Transport.ResponseHeaderTimeout` via transport pool
- **KEDA_CONDITION_WAIT_TIMEOUT** (1500ms): Max wait for workload to scale to >=1 replicas — bounds the `waitFunc` context in the proxy handler
- **KEDA_HTTP_IDLE_CONN_TIMEOUT** (90s): Time before idle connections are evicted — `http.Transport.IdleConnTimeout`
- **KEDA_HTTP_TLS_HANDSHAKE_TIMEOUT** (10s): Max wait for TLS handshake — `http.Transport.TLSHandshakeTimeout`
- **KEDA_HTTP_EXPECT_CONTINUE_TIMEOUT** (1s): Max wait for 100-continue response from backend — `http.Transport.ExpectContinueTimeout`
- **KEDA_HTTP_DIAL_RETRY_TIMEOUT** (15s): Total time budget for dial retries — bounds the retry loop via `context.WithTimeout`

### Changes
- Replace wait.Backoff with fixed 50ms retry interval + configurable timeout
- Add KEDA_HTTP_DIAL_RETRY_TIMEOUT env var to Timeouts config
- Remove DefaultBackoff(), MinTotalBackoffDuration and backoff.go
- Update all tests to use simplified DialContextWithRetry signature

### Checklist

  - [`README.md`](../README.md)

See #1449


pkg/net/dial_context.go
```diff
@@ -1,61 +1,54 @@
...
-	numDialTries := backoff.Steps
+// retryInterval is kept short to minimize latency when the target becomes
+// reachable. Not exposed as configuration — retryTimeout is the user-facing knob.
+const retryInterval = 50 * time.Millisecond
+
+// DialContextWithRetry retries failed dials at a fixed interval until
+// retryTimeout expires or the parent context is cancelled.
+func DialContextWithRetry(coreDialer *net.Dialer, retryTimeout time.Duration) DialContextFunc {
+	// Default covers slow Service IP rule propagation to kube-proxy, Cilium, etc
+	retryTimeout = cmp.Or(retryTimeout, 15*time.Second)
```
Fedosin: Maybe it is better to declare 15*time.Second as a constant?

Btw why is it 15 seconds? Current DefaultBackoff is only 3.5 seconds. Do we really need such a big timeout?

PR #1465, opened 2026-02-12
refactor: migrate build system from Docker to ko

### LINKED ISSUE #1464: Improve the build scripts, procedure and documentation
## Problem

The current container build is really slow and doesn't cache the layers well which makes redeployments of changes into a local cluster extremely slow.

The documentation is outdated quite a lot referencing endpoints that don't exist anymore.

The Makefile is quite verbose and forces the user to use its installed tools instead of the ones they have on the system.

## Proposal

Build container images with https://ko.build/ as knative does, which is an extremely fast and simple way to achieve what we do right now.
Update the documentation and explain how this can be used.

I actually am 95% done with the implementation of this proposal, PR will follow very soon.

linkvt:
ko is a simple tool to build containerized go apps by leveraging the go build cache. It supports all of our requirements (cross platform, custom base image which I changed though, retain the existing image names) and has a few major improvements over our current approach:
- Extremely fast rebuild in seconds instead of minutes including pushing changes to the local kind registry
  - It is used by a few major open source projects, see https://ko.build/
  - Additional bonus: space used by docker won't grow ~30 GBs for a build
- Auto replacing the `image: ko://` in generated manifests with the recently build image before applying to them to the cluster
- No need to maintain a Dockerfile, ko basically just builds the binary and put it in the specified base image
- Simple multi platform builds

I also updated the documentation which quite outdated (especially the debugging section) and reorganized the config files a bit while ensuring that the upgrade from the resources from <=0.12.1 does work (upgrade from helm chart was and is still not supported).

#### refactor: migrate build system from Docker to ko

Replace the Docker-based build pipeline with ko for building and
publishing container images, removing the dependency on Dockerfiles
and the keda-tools container.

- Add .ko.yaml with wolfi-dev/static base image, env, and ldflags
- Replace Docker Makefile targets with ko equivalents
- Add KO_RELEASE_PLATFORMS for explicit multi-arch publish/release
- Update CI workflows to use setup-go + setup-ko
- Simplify kustomize overlays and rename config files to kebab-case
- Fix deprecated set-output in release workflow
- Update docs for ko-based workflow

#### chore: update pre-commit config and add lint exclusions

Remove redundant go-fmt hook (already covered by golangci-lint),
update hook repos, and add lint exclusions for package naming
conflicts in interceptor/metrics and pkg/build.

Changes:
- Remove deprecated dnephin/pre-commit-golang hook
- Update pre-commit-hooks to v6.0.0 and doctoc to v2.2.0
- Bump minimum_pre_commit_version to 3.0.0
- Add revive var-naming exclusions for interceptor/metrics and pkg/build
- Document pre-commit setup in docs/developing.md

### Checklist

  - [`README.md`](../README.md)

Fixes #1464


.github/workflows/tests.yaml
```diff
@@ -16,62 +16,41 @@ jobs:
...
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - uses: actions/setup-go@7a3fe6cf4cb3a834922a1244abfce67bcef6a0c5 # v6.2.0
         with:
           go-version-file: "go.mod"
+      - uses: golangci/golangci-lint-action@1e7e51e771db61008b38414a730f564565cf7c20 # v9.2.0
       - uses: pre-commit/action@2c7b3805fd2a0fd8c1884dcaebf91fc102a13ecd # v3.0.1
+        env:
+          SKIP: golangci-lint
```
JorTurFer: Just curiosity, why do we split golangci and pre-commit?

config/interceptor/scaledobject.yaml
```diff
@@ -3,7 +3,7 @@ kind: ScaledObject
 metadata:
   name: interceptor
 spec:
-  minReplicaCount: 3
+  minReplicaCount: 1
```
JorTurFer: I'd keep this as it was because 1 isn't HA and we generate release manifests from kustomize

JorTurFer (commented):
Awesome job! just some comments inline

JorTurFer (approved):

Fedosin (approved):

PR #1488, opened 2026-02-23
fix: handle HTTPScaledObject not found gracefully in scaler

### LINKED ISSUE #1487: IsNotFound errors are not handled gracefully in the scaler
### Report

When HTTPScaledObjects are deployed rapidly (e.g. during batch deployments or performance testing), the scaler's controller-runtime informer cache may not have synced the new HTTPScaledObject by the time KEDA starts polling for metrics. This causes a cascade of errors: `GetMetrics` return hard errors, which kill the `StreamIsActive` gRPC stream, forcing KEDA to reconnect and retry producing many of ERROR-level log entries per deployment wave.

### Expected Behavior

Scaler should gracefully handle IsNotFound errors.

### Actual Behavior

Multiple errors are seen in the logs.

The timeline of the race condition:

1. User deploys an HTTPScaledObject (e.g. `my-app-httpso`)
2. KHA operator reconciles it and creates a KEDA ScaledObject
3. KEDA immediately calls the external scaler's `StreamIsActive` -> `IsActive` -> `GetMetrics`
4. The scaler does `e.reader.Get()` which reads from the controller-runtime informer cache
5. The cache hasn't received the watch event yet -> `"HTTPScaledObject not found"`
6. Hard error propagates up and terminates the gRPC stream
7. KEDA reconnects, hits the same issue, producing error spam

Full [log](https://github.com/user-attachments/files/25498154/kha-memory-test.log)

### HTTP Add-on Version

0.12.2

### Kubernetes Version

1.33

### Platform

Google Cloud

Yes

### Anything else?

_No response_

Fedosin:
This PR fixes the race condition by handling the `IsNotFound` case in `GetMetrics` gracefully: instead of returning an error, it returns metric value `0` (not active) and logs at debug level. Once the cache syncs (typically within a second), subsequent calls return real values. The gRPC stream stays alive, avoiding the reconnection storm.

### Checklist

  - [`README.md`](../README.md)

Fixes #1487

### DIFF
--- a/scaler/handlers.go
+++ b/scaler/handlers.go
@@ -15,6 +15,7 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/kedacore/keda/v2/pkg/scalers/externalscaler"
 	"google.golang.org/protobuf/types/known/emptypb"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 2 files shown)


scaler/handlers.go
```diff
@@ -200,16 +201,27 @@ func (e *impl) GetMetrics(ctx context.Context, metricRequest *externalscaler.Get
 		return nil, err
 	}
 
+	// generated the metric name for HTTPScaledObject
+	namespacedName := k8s.NamespacedNameFromNameAndNamespace(httpScaledObjectName, sor.Namespace)
+	metricName := MetricName(namespacedName)
+
 	httpso := &httpv1alpha1.HTTPScaledObject{}
 	if err := e.reader.Get(ctx, types.NamespacedName{Namespace: sor.Namespace, Name: httpScaledObjectName}, httpso); err != nil {
```
linkvt: Minor nitpicks:
- we have `types.NamespacedName{Namespace: sor.Namespace, Name: httpScaledObjectName}` here and above the `k8s.NamespacedNameFromNameAndNamespace(...)` which also returns a `NamespacedName`
- `not yet available in cache` is maybe a bit specific?

scaler/handlers.go
linkvt: I'm currently not sure if the fix below was enough for the errors to disappear as GetMetricSpec is IIRC called before GetMetrics and would probably have errors as well?

linkvt (commented):
Gave it another look after Jans review request here, only found the other location where the IsNotFound check might possibly missing.

@wozniakjan from what I understand the scaler will not respond until the caches are synced:
https://github.com/kedacore/http-add-on/blob/4dd989f14577052a8a0e8dc1ec7362a4008e9890/scaler/main.go#L125-L137

I guess the fallback scenario would only make sense then if the scaler can't access the HTTPSO/IR for some reason while the interceptor is able and continues to serve traffic or am I missing something?

PR #1531, opened 2026-03-10
fix: drop semconv version-pinned import in favor of schemaless resources

linkvt:
The semconv package uses versioned import paths (e.g. semconv/v1.40.0) that must be manually updated whenever OTel SDK dependencies are bumped. This causes friction with automated dependency updates (Renovate) since each OTel SDK version requires a matching semconv version.

Use resource.NewSchemaless with plain attribute strings instead, which is a widely-used workaround for this known issue
(open-telemetry/opentelemetry-go#4886) as `service.name` and `service.version` basically never change (at least very likely not in otel v1).

### Changes
- Replace resource.NewWithAttributes(semconv.SchemaURL, ...) with resource.NewSchemaless(attribute.String(...), ...)
- Remove semconv import from otelmetrics, prommetrics, and tracing

### Checklist

### DIFF
--- a/interceptor/metrics/otelmetrics.go
+++ b/interceptor/metrics/otelmetrics.go
@@ -9,7 +9,6 @@ import (
 	api "go.opentelemetry.io/otel/metric"
 	"go.opentelemetry.io/otel/sdk/metric"
 	"go.opentelemetry.io/otel/sdk/resource"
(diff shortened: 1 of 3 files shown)


Fedosin (approved):
/approve

Issue #1501, opened 2026-02-27
Implement InterceptorRoute as successor to HTTPScaledObject


linkvt:
**Disclaimer:** This is a proposal and things might change.

## Summary

Replace the `HTTPScaledObject` (HTTPSO) CRD with a new routing CRD (working name: `InterceptorRoute`) for routing, metric collection, and interceptor-specific configuration, paired with a standard KEDA `ScaledObject` for scaling behavior. The HTTPSO will be deprecated and removed.

As discussed with the other KEDA developers in [Slack](https://kubernetes.slack.com/archives/C01JGDP8MB8/p1770991198112379), we want to implement a successor for the HTTPScaledObject - the referenced document might not be public but relevant infos can be found below. This issue should explain the why, how and track the progress.

I would suggest starting [a new discussion](https://github.com/kedacore/http-add-on/discussions/new/choose) in case you have feedback, concerns, suggestions, ... as they have a better support for discussion threads and to keep the issue clean. Feedback is very welcome!

## Motivation

- **Mixed concerns**: HTTPSO mixes routing config (hosts, paths, headers, timeouts, cold-start failover, metrics collection) used by the interceptor with scaling config (replicas, cooldown) translated into a ScaledObject. A new routing CRD + `ScaledObject` cleanly separates these.
- **Hidden ScaledObject**: The operator creates a ScaledObject behind the scenes with hardcoded values, preventing users from leveraging advanced SO features (fallback, multiple triggers, pause annotations). A `skip-scaledobject-creation` annotation exists as a workaround.
- **Inconsistent UX**: HTTP Add-on is the only KEDA scaler requiring a separate CRD instead of using `ScaledObject` directly.
- **Independent release cycles**: The routing CRD lives within the HTTP Add-on and can evolve without coordinated KEDA core releases.
- **Better naming**: `HTTPScaledObject` is not directly a ScaledObject. A new routing CRD describes the resource accurately and accommodates non-HTTP protocols (WebSocket, gRPC).

## Proposed Design

- **Users create two resources**: `InterceptorRoute` for routing/interceptor config, `ScaledObject` for scaling config.
- **The SO trigger references the InterceptorRoute** by name in its metadata, so the external scaler knows which route's metrics to report.
- **The operator** no longer creates ScaledObjects. Validation is handled by CEL rules on the CRD. The operator may gain additional responsibilities in the future (e.g. deploy scoped Interceptor instances, act as an envoy XDS server, ...) or get removed.
[...]

Issue #1474, opened 2026-02-19
Rethink which timeouts should be configurable


linkvt:
### Proposal

We currently allow users to configure quite a few timeouts in a few different ways (some global, some overridable per HTTPSO).
These timeouts are often quite technical and might even leak internal implementation details like `KEDA_HTTP_CONNECT_TIMEOUT` or `KEDA_HTTP_DIAL_RETRY_TIMEOUT`.
The naming could also be improved for some others like `KEDA_CONDITION_WAIT_TIMEOUT` which doesn't state which condition we wait for.

We should consider the users perspective and requirements and derive the timeouts we want to expose based on that.
Something that isn't easy right now is configuring a simple timeout for the time the response from the backend service takes as you can only configure very fine granular timeouts right now.
We should also keep the coldstart failover timeout in mind.

Here is a list of the current timeouts and their function (might be outdated when you read this so make sure to check the code):
- **KEDA_HTTP_CONNECT_TIMEOUT** (500ms): Per-attempt TCP dial timeout — passed to `net.Dialer.Timeout`
- **KEDA_HTTP_KEEP_ALIVE** (1s): TCP keepalive probe interval — passed to `net.Dialer.KeepAlive`
- **KEDA_RESPONSE_HEADER_TIMEOUT** (500ms): Max wait for response headers from backend — set on `http.Transport.ResponseHeaderTimeout` via transport pool
- **KEDA_CONDITION_WAIT_TIMEOUT** (20s): Max wait for workload to scale to >=1 replicas — bounds the `waitFunc` context in the proxy handler
- **KEDA_HTTP_IDLE_CONN_TIMEOUT** (90s): Time before idle connections are evicted — `http.Transport.IdleConnTimeout`
- **KEDA_HTTP_TLS_HANDSHAKE_TIMEOUT** (10s): Max wait for TLS handshake — `http.Transport.TLSHandshakeTimeout`
- **KEDA_HTTP_EXPECT_CONTINUE_TIMEOUT** (1s): Max wait for 100-continue response from backend — `http.Transport.ExpectContinueTimeout`
- **KEDA_HTTP_DIAL_RETRY_TIMEOUT** (15s): Total time budget for dial retries — bounds the retry loop via `context.WithTimeout`

We should also consider choosing a consistent parsing approach, e.g. to not name fields `xyzSeconds` but use go parseable durations instead.

As this is likely a breaking change and quite a big topic, please provide a design document for discussion before implementing.

[...]

PR #1532, opened 2026-03-10
feat: implement InterceptorRoute CRD

linkvt:
### TODO

- [x] check E2E tests
- [x] discuss highlighted issues with reviewers (see ⚠️ )
- [x] check all non-v1 TODOs in the PR
- [x] add follow-up TODOs to tracking issue (docs, migration guide)

### Summary

Implement InterceptorRoute while still supporting the HTTPScaledObject as configuration resource. Internal routing implementation converts HTTPSOs to IRs.
HTTPSO related code should be removed before the release of v1.

### Changes

- CRD Changes compared to HTTPSO
    - timeouts (also coldstart.timeout) are not configurable (yet) via the CR
    	- this is IMO a big topic, see below
	- scalingMetric
		- allow specifying rate AND concurrency
		- ⚠️ discuss `requestRate` - should it be named differently like `rate`, or `rps`? we dont use `requestConcurrency` either. Should keep future metrics in mind.
	- coldstart.failover was renamed to coldstart.fallback, see below
		- it is still optional via a pointer as we will likely add placeholder there soon as another optional field
	- rules
		- as discussed in the concept, allows configuring multiple distinct matchers
		- hosts is a string list as opposed to the path/header struct, as we can probably do everything with it (exact, wildcard, possibly regex)
		- paths is a struct list to allow future exact path matching
		- headers is a struct list to check for any or a specific header (and possibly more types in the future
- Operator
	- doesn't do much right now, it's there with a TODO before v1 release
	- possible tasks in the future
		- updating the status for keda Addon triggers (future) to contain the scalerAddress
		- provide XDS server configuration for e.g. envoy or such
		- handle other CRDs like the deploying interceptors
	- CRD status is still there and set to ready by default
- Interceptor
	- rewritten to only use InterceptorRoute in basically all places
	- converts HTTPSO to InterceptorRoute when setting up the routing table for backward compatibility
- Scaler
	- added InterceptorRoute handling in front of the existing HTTPSO handling
	- if KEDA asks about metrics for a HTTPSO, we fallback to the existing logic
- E2E test changes
	- HTTPSO exclusive E2E tests were kept as is
	- some central original HTTPSO tests were duplicated to still test it until we remove the HTTPSO
	- the rest of the HTTPSO tests were updated with minimal changes to use InterceptorRoute

Documentation will be added in a follow-up PR as this PR is already quite large.

#### Failover vs Fallback
I let claude review all field names with the result of renaming failover to fallback. Failover seems to be generally referring to sending the request to another equal system, like another replica. Fallback (not failback) seems to be referring to reacting to a degradation by still responding something that might not be the same as the default target would have responded, but better than nothing or failing. 

#### Timeouts

We should reconsider the whole timeout implementation from the CRD and the env vars.
The current HTTPSO timeouts are very technical and fine granular, we should rather think about timeouts from a user perspective starting from the request start time or such.

This includes the coldstart.timeout.
This does not mean we should hardcode timeouts.

Related issue #1474 

### Checklist

See #1501


Fedosin (changes requested):

JorTurFer (commented):
Awesome job! I hope that I have not missed any important point , just 2 small details inline

Fedosin (commented):
The patch looks great now! I've pointed out a few nits we could improve before merging, but they can definitely be addressed in a follow-up.

Fedosin (approved):
Let's merge this and address some nits later.

PR #1538, opened 2026-03-11
chore: provide minimal devcontainer config

linkvt:
We are not actually sure whether the devcontainer is used at all, we only know the content was copied here from the keda config and is in fact quite outdated.
I propose a minimal devcontainer config, golangci-lint and make are already inside and vscode sets up the go extension. Missing are ko, docker, operator-sdk (unused anyways), lots of old go tools mostly replaced by gopls or pre-commit which is also run in CI. It also doesn't work well when using podman instead of docker but at that point I the user is IMO advanced enough to use his local dev env.

In case there is an actual request we can add features again via https://containers.dev/features . Until then I prefer this over some complex outdated config we don't support and maintain.

This also reduces noisy renovate PRs.

### Checklist


Fedosin (approved):
/approve

PR #1539, opened 2026-03-11
chore: group regex-managed action versions with GitHub Actions updates

linkvt:
PR #1537 tracked golangci-lint via renovate, it would be a separate PR though as shown in the dashboard issue.
To reduce PR noise we should group all GHA related changes.

image

### Checklist


Fedosin (approved):
/approve

PR #1555, opened 2026-03-23
refactor: simplify responseWriter using Unwrap pattern

linkvt:
Use Unwrap() so http.ResponseController discovers optional interfaces on the underlying writer automatically, replacing manual Hijack and EnableFullDuplex delegation and removing the forced per-write Flush. The httputil.ReverseProxy is also flushing the writer on its own.

### Checklist


Fedosin (approved):
/approve

PR #1558, opened 2026-03-24
refactor: extract endpoint resolution into dedicated middleware

linkvt:
The forwarding handler mixed readiness waiting, endpoint selection, and reverse proxying in a single function. This made it difficult to insert a request-level timeout and to change how endpoints are resolved (e.g. for possible direct pod routing instead of using ClusterIP), since the URL decision was buried in proxy logic.
This PR extracts the whole endpoint resolution into a dedicated middleware.

### Changes
- Extract EndpointResolver middleware from proxy_handlers.go for readiness waiting, cold-start headers, and fallback selection
- Standardize middleware constructors: next http.Handler as first parameter, internal field renamed to "next"
- Remove unused RequestWithFallbackURL helper
- Simplify OTel tracing in Upstream (otelhttp handles propagation)
	- the current code was just overly complicated and manipulating spans it didn't own
	- the tests were updated with the minimal changes, IMO these tests should be reworked in the future
	- `traceparent` headers are basically never added to responses which we did in the past for no reason
 
### Checklist


Fedosin (approved):
LGTM 👍

PR #1563, opened 2026-04-01
feat: redesign interceptor metrics with bounded labels

linkvt:
When looking at the problem causing #1559 I decided to mostly rewrite the metrics we collect and expose due to the following reasons:
- the used labels had a high cardinality causing the underlying issue in the first place
	- aligned the new labels with what other ingress controllers expose
	- limited the max cardinality of user controlled labels like the HTTP method
- metric names were renamed
	- [otel semconv](https://opentelemetry.io/docs/specs/semconv/general/naming/) recommends to pluralize when using "discrete instances of a countable quantity" 
	- _count is usually only used in histograms
- tracing didn't start at the outermost layer
	- reordered handlers to tracing is the first one

### Summary
Replace unbounded path/host metric labels with bounded route identity labels derived from the InterceptorRoute, fixing OOM under high-cardinality traffic.
Consolidate dual MeterProvider into a single provider with multiple readers.

### Changes
- Add new Instruments struct holding counter, histogram, pending gauge
- Add routeInfo context sharing between Metrics and Routing middleware
  - the route name and namespace are only identified in a "lower" middleware, sharing the info back up seems to be a valid pattern
- Rename metrics: interceptor_request_count → interceptor_requests_total, interceptor_pending_request_count → interceptor_pending_requests
- Add interceptor_request_duration_seconds histogram
	- Buckets are following the OTel recommendations https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#metric-httpserverrequestduration
- Normalize non-standard HTTP methods to "_OTHER"
- Reorder handler chain: Tracing → Logging → Metrics → Routing → Counting → EndpointResolver → Upstream
	- spans should include everything and not start somewhere in the middle of the handler chain
- Update e2e observability tests for new metric names
- Update docs with new metric names and labels
- Update operational docs with profiling instructions (was useful for this PR)

### Checklist

  - [`README.md`](../README.md)

Fixes #1559


wozniakjan (approved):
I'm in favour of this!

Even though users might miss some granularity, reconfigure their alert managers, http-add-on's main goal has always been autoscaling, with observability as a bonus.

Given addon is pre v1.0.0 release, breaking changes are expected and this is an example of necessary breaking change. We just need to make sure to include this information visibly in release notes and perhaps also release notes for the http-add-on chart?

Fedosin (approved):
/lgtm 👍

Issue #1473, opened 2026-02-19
Evaluate sending requests to Endpoint IPs directly


linkvt:
### Proposal

We should evaluate if and in which cases it makes sense to forward requests directly to the Endpoint IPs (=Pod IPs) we read from the EndpointSlice instead of the Service IP. knative implements this optimization already if I remember it correctly.

This could be especially useful for a cold-start request where it can take a few seconds depending on the Kubernetes networking setup until the Service IP is reachable on the node where the interceptor runs on.
The current cold-start flow is:
1. Pod is getting Ready after scaling from 0 to 1
2. EndpointSlice is getting updated to include the Pod IP as a Ready endpoint
    - The Service IP can not be used yet, connections are refused
3. kube-proxy (or another networking solution) sees this change and configures the Service IP on the node
    - This is not instant and can take a few seconds
4. Traffic can now be sent from the nodes that have the Service IP configured
       
Some environments like GKE clusters using Dataplane V1 (based on kube-proxy) instead of V2 (based on cilium taking kube-proxys role IIRC) seem to only configure the Service IP every 10 seconds as indicated in #1449 .

We should keep in mind how this optimization:
- works when using TLS
- works when using a ServiceMesh
- affects the distribution of network traffic (i.e. not to send all requests to the first ready endpoint)

Before implementing please investigate and propose a design as this is IMO quite a complex topic.

### Use-Case

The goal is to send the traffic to ready Pods as early as possible in a cold-start scenario.

### Is this a feature you are interested in implementing yourself?

Yes

Fedosin:
Hey @AtharvaPakade, just checking in on this! Are you still actively working on it? If your plate is full at the moment, I'd be glad to pick it up from here. Let me know how you'd like to proceed.

AtharvaPakade:
Hi @Fedosin I’ve started working on this. I was blocked on something last week, but I’ve made progress now and expect to push the final PR in the next 1–2 days. Appreciate the check-in — I’ll keep you posted.

AtharvaPakade:
@linkvt @Fedosin 

After diving deep into this issue I figured. We need to break this issue into mutliple focused PRs.
The Main goal of this issue is to reduce the cold start time.
Following are the Two distinct optimisations that are possible:
1. Direct-to-pod routing on cold start — #1585
[...]

PR #1571, opened 2026-04-08
interceptor: add HTTP keep-alive toggle

ikalnytskyi:
Persistent HTTP connections are not always helpful for all workloads. With synchronous/blocking backends, one interceptor can keep a connection open to a backend worker after a request finishes. Another interceptor may reuse that worker and wait until the first connection is closed, causing cross-replica blocking and higher latency.

Add `KEDA_HTTP_DISABLE_KEEP_ALIVES` so operators can disable connection reuse when this behavior appears.

### Checklist

- [x] Any necessary documentation is added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,7 +35,7 @@ This changelog keeps track of work items that have been completed and are ready 
 
 ### Improvements
 
-- **General**: TODO ([#TODO](https://github.com/kedacore/http-add-on/issues/TODO))
+- **Interceptor**: Add `KEDA_HTTP_DISABLE_KEEP_ALIVES` to disable keep-alive connections from interceptor to backend services ([#1571](https://github.com/kedacore/http-add-on/pull/1571))
 
 ### Fixes
 
--- a/interceptor/config/timeouts.go
+++ b/interceptor/config/timeouts.go
@@ -20,6 +20,8 @@ type Timeouts struct {
 	WorkloadReplicas time.Duration `env:"KEDA_CONDITION_WAIT_TIMEOUT" envDefault:"20s"`
 	// ForceHTTP2 toggles whether to try to force HTTP2 for all requests
 	ForceHTTP2 bool `env:"KEDA_HTTP_FORCE_HTTP2" envDefault:"false"`
+	// DisableKeepAlives disables HTTP keep-alives for requests from interceptor to backend services.
+	DisableKeepAlives bool `env:"KEDA_HTTP_DISABLE_KEEP_ALIVES" envDefault:"false"`
 	// MaxIdleConns is the max number of idle connections to keep in the
 	// interceptor's internal connection pool across all backend services.
 	// Increase this if you proxy to many unique backend services.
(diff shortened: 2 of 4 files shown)


linkvt:
Hi @ikalnytskyi , thanks for your PR!

I understand the solution you proposed but want to understand the problem as well as I'm not really sure about it right now.

Before that: There is currently one issue where existing backends continue to receive more traffic due to TCP connection to them being reused than new backend pods e.g. after scaling up.
When a new connection is made from the interceptor the k8s load balancing resolves an Pod IP, when existing connections are used this doesnt happen.
This should be worked around in most cases in #1473 as connections are then targeting the Pod IPs and not the Services ClusterIP.

Completely disabling connection reuse as done in this PR is quite a heavy solution and affects all InterceptorRoutes/HTTPScaledObjects with massive performance implications...
Is maybe decreasing the idle connection timeout a simpler solution?

Is this the problem that you have?
If not could you describe your exact use-case and problem better so that we're sure we're fixing the right thing? Which languages/frameworks are you using that cause the issue?

Thanks!

ikalnytskyi:
Hey @linkvt,

Sorry for the delay in my response.

> want to understand the problem as well as I'm not really sure about it right now.

Let's assume we have two replicas (A and B) of keda-http-interceptor and a single replica of the upstream service.

* When a user sends a request, keda-http-interceptor (A) opens a connection to my upstream server, serves the request, and puts the connection in the connection pool.
* When a user sends a second request, it may hit either replica A or B.
* If it hits A, everything works fine because it reuses the pooled connection.
* If it hits B, it opens a new connection successfully (socket backlog allows it) and forwards the request, but the upstream server can't serve it because its synchronous worker is still occupied by A's pooled connection.

This happens when the upstream server respects the `Connection` header and doesn't close connections unless asked, and uses synchronous workers that process one request at a time. Common with Python WSGI servers that avoid threads due to GIL.

> Completely disabling connection reuse as done in this PR is quite a heavy solution and affects all InterceptorRoutes/HTTPScaledObjects with massive performance implications...

I agree it should be configurable at the `HTTPScaledObjects` level since it depends on the upstream workload. Though by that logic, `KEDA_HTTP_REQUEST_TIMEOUT` and `KEDA_HTTP_RESPONSE_HEADER_TIMEOUT` should also be per-upstream since each upstream quite often has different SLAs.

I think saying the performance implications are massive is an exaggeration. It's really not that bad, especially without TLS. Nginx has used HTTP/1.0 for upstream proxying for years (i.e., no connection pooling by default), and many keep using it this way to date.

> Is maybe decreasing the idle connection timeout a simpler solution?

That's the workaround that I use: decrease it to 1ms here, and then configure the connection pool to a couple of seconds in my upstream server to prevent races when one party hasn't yet sent FIN/RST. I just feel like this is a workaround, and I'd rather make sure the requests are forwarded to upstream with `Connection: close` to begin with.

linkvt:
Sorry for the late response, I just had another look at this and also thanks for the great explanation!

I'm wondering why the backend/upstream is not responding with the header `Connection: Close`, if I understand it right the interceptor would close the connection so that the upstream isn't "blocked" anymore?

HTTP1.1 exists since quite a long time and defaults to connection reuse as well as HTTP2. Supporting a HTTP1.0 only backend today seems like quite a legacy use case?  
By default , ingress controllers like nginx, traefik, ... seem to keep connections open today as well, not all allow changing that behavior.

Btw, we also support timeouts per upstream (now), see https://keda.sh/http-add-on/0.14/user-guide/configure-timeouts/

PR #1577, opened 2026-04-14
feat: support OTEL_TRACES_SAMPLER and OTEL_TRACES_SAMPLER_ARG settings

### LINKED ISSUE #1534: Support additional OTEL tracing settings
### Proposal

Currently the interceptor supports some options for configuring tracing via OTEL but doesn't support a sampling option. This would be nice to support to avoid issues with overwhelming OTEL endpoints.

Specifically the following options:
* `OTEL_TRACES_SAMPLER`
* `OTEL_TRACES_SAMPLER_ARG`
* 
https://opentelemetry.io/docs/languages/sdk-configuration/general/#otel_traces_sampler

### Use-Case

This would help limit the number of traces sent to OTEL trace endpoints and avoid issues with overwhelming them. It would also help reduce data being retained and avoid needing to configure sampling within the OTEL collector pipeline itself.

Yes

### Anything else?

_No response_

Fedosin:
Support configurable trace sampling in interceptor tracing via `OTEL_TRACES_SAMPLER` and `OTEL_TRACES_SAMPLER_ARG`.

This replaces the hard-coded `AlwaysSample()` default with configurable sampler selection, updates docs and e2e config examples, and adds coverage for sampler behavior. Invalid or unrecognized sampler values/args are now warned and ignored with fallback defaults, aligned with OpenTelemetry SDK environment variable guidance.

### Checklist

  - [`README.md`](../README.md)

Fixes #1534

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@ This changelog keeps track of work items that have been completed and are ready 
 ### Improvements
 
 - **General**: TODO ([#TODO](https://github.com/kedacore/http-add-on/issues/TODO))
+- **Interceptor**: Support `OTEL_TRACES_SAMPLER` and `OTEL_TRACES_SAMPLER_ARG` for trace sampling configuration ([#1534](https://github.com/kedacore/http-add-on/issues/1534))
 
 ### Fixes
 
--- a/interceptor/tracing/tracing.go
+++ b/interceptor/tracing/tracing.go
@@ -17,7 +17,7 @@ import (
 	"github.com/kedacore/http-add-on/interceptor/config"
 )
 
-var serviceName = "keda-http-interceptor"
+const serviceName = "keda-http-interceptor"
 
 func SetupOTelSDK(ctx context.Context, tCfg config.Tracing) (shutdown func(context.Context) error, err error) {
 	var shutdownFuncs []func(context.Context) error
@@ -79,7 +79,6 @@ func newTraceProvider(ctx context.Context, res *resource.Resource, tCfg config.T
 	}
 
 	traceProvider := trace.NewTracerProvider(
-		trace.WithSampler(trace.AlwaysSample()),
 		trace.WithBatcher(traceExporter),
 		trace.WithResource(res),
 	)
--- a/interceptor/config/tracing.go
+++ b/interceptor/config/tracing.go
@@ -8,7 +8,7 @@ import (
 type Tracing struct {
 	// States whether tracing should be enabled, False by default
 	Enabled bool `env:"OTEL_EXPORTER_OTLP_TRACES_ENABLED" envDefault:"false"`
-	// Sets what tracing export to use, must be one of: console,http/protobuf, grpc
+	// Sets what tracing export to use, must be one of: console, http/protobuf, grpc
 	Exporter string `env:"OTEL_EXPORTER_OTLP_TRACES_PROTOCOL" envDefault:"console"`
 }
 
--- a/interceptor/tracing/tracing_test.go
+++ b/interceptor/tracing/tracing_test.go
@@ -9,10 +9,11 @@ import (
 )
 
 func TestTracingConfig(t *testing.T) {
+	t.Setenv("OTEL_EXPORTER_OTLP_TRACES_PROTOCOL", "console")
+
 	tracingCfg := config.MustParseTracing()
 	tracingCfg.Enabled = true
 
-	// check defaults are set correctly
 	assert.Equal(t, "console", tracingCfg.Exporter)
 	assert.True(t, tracingCfg.Enabled)
 }


interceptor/tracing/tracing.go
```diff
@@ -78,14 +85,76 @@ func newTraceProvider(ctx context.Context, res *resource.Resource, tCfg config.T
 		return nil, err
 	}
 
+	sampler := newSampler(ctx, tCfg.Sampler, tCfg.SamplerArg)
+
 	traceProvider := trace.NewTracerProvider(
-		trace.WithSampler(trace.AlwaysSample()),
+		trace.WithSampler(sampler),
```
JorTurFer: Do we need to manually create the sampler? IIRC, the envs vars are automatically read by the exporter and applied automatically without any code.

JorTurFer: at least, that's what I understand from here https://pkg.go.dev/go.opentelemetry.io/otel/sdk/trace#WithSampler 

Image

Fedosin: Yeah, here we overwrite the built-in OTEL logic a bit. The main reasons for that are:

- Use interceptor’s structured logger for invalid config warnings (instead of SDK/global handler behavior).
- Enforce stricter validation/fallback behavior (notably non-finite values like NaN).
- Keep behavior explicitly unit-tested in our code, independent of SDK internal changes.

If you prefer, I can simplify this and rely directly on the SDK env parsing instead.

@linkvt what do you think?

Fedosin: @JorTurFer Yep, without custom logic it looks better :) We can always return additional validation if necessary. But now we just say that we rely on OTEL validation.

CHANGELOG.md
```diff
@@ -39,6 +39,7 @@ This changelog keeps track of work items that have been completed and are ready
 ### Improvements
 
 - **General**: TODO ([#TODO](https://github.com/kedacore/http-add-on/issues/TODO))
+- **Interceptor**: Support `OTEL_TRACES_SAMPLER` and `OTEL_TRACES_SAMPLER_ARG` for trace sampling configuration, and ignore invalid sampler values with warnings per OTel SDK environment variable guidance ([#1534](https://github.com/kedacore/http-add-on/issues/1534))
```
linkvt: I think the changelog is outdated since the most recent change simplifying things.

config/e2e/otel/interceptor-deployment.yaml
```diff
@@ -28,3 +28,5 @@ spec:
               value: "http://opentelemetry-collector.open-telemetry-system:4318/v1/traces"
             - name: OTEL_EXPORTER_OTLP_TRACES_INSECURE
               value: "true"
+            - name: OTEL_TRACES_SAMPLER
+              value: "parentbased_always_on"
```
linkvt: Do we want to set this env var if it is the default value?

docs/operate.md
```diff
@@ -72,6 +72,20 @@ Optional variables
...
+
+`OTEL_TRACES_SAMPLER` - Sets the sampler used for tracing. Supported values (`parentbased_always_on` by default):
+* `always_on` - Always sample every trace
+* `always_off` - Never sample any trace
+* `traceidratio` - Sample a fraction of traces based on the trace ID
+* `parentbased_always_on` - Respect the parent span's sampling decision; sample root spans
+* `parentbased_always_off` - Respect the parent span's sampling decision; do not sample root spans
+* `parentbased_traceidratio` - Respect the parent span's sampling decision; sample root spans by ratio
+
+`OTEL_TRACES_SAMPLER_ARG` - Argument for ratio-based samplers. Specifies the desired sampling ratio as a floating point number between 0 and 1 (e.g. `0.5` for 50% sampling, `0.1` for 10%). Only used when `OTEL_TRACES_SAMPLER` is set to `traceidratio` or `parentbased_traceidratio`. Defaults to `1.0` (100%) if not set.
```
linkvt: Do we want to duplicate the OTEL docs into our docs? I would assume that the OTEL docs are more uptodate and that a simple reference to them might be good enough?

interceptor/tracing/tracing.go
```diff
@@ -79,7 +79,6 @@ func newTraceProvider(ctx context.Context, res *resource.Resource, tCfg config.T
 	}
 
 	traceProvider := trace.NewTracerProvider(
-		trace.WithSampler(trace.AlwaysSample()),
```
linkvt, line 82: In theory this would be a breaking change but I don't think we need to mention it, wouldn't assume there arent that many users that use the addon AND want only tracing enabled for it but not hte whole request span.

docs/operate.md
linkvt: I see the whole section was removed, also fine for me - quick reference to the OTEL docs would also been fine but you decide 🙂

linkvt (approved):

PR #1580, opened 2026-04-17
refactor: set Content-Type in interceptor queue endpoint handler

linkvt:
The queue counts endpoint did not set Content-Type: application/json, causing clients to receive text/plain, I noticed this while looking at the queue endpoint in the browser.
The error path for JSON encoding failures also attempted to set a 500 status after the response had already started writing, which was a no-op (headers cant be set once body is written).

This is mostly cosmetic and doesn't change behavior but makes debugging/inspecting this endpoint in the browser easier.

### Changes
- Set Content-Type: application/json before writing JSON response
- Use http.Error() for the Current() error path
- Remove unreachable error handling after partial JSON write

### Checklist

### DIFF
--- a/pkg/queue/queue_rpc.go
+++ b/pkg/queue/queue_rpc.go
@@ -28,28 +28,13 @@ func newSizeHandler(
 		cur, err := q.Current()
 		if err != nil {
 			lggr.Error(err, "getting queue size")
-			w.WriteHeader(500)
-			if _, err := w.Write([]byte(
-				"error getting queue size",
-			)); err != nil {
-				lggr.Error(
-					err,
-					"could not send error message to client",
-				)
-			}
+			http.Error(w, "error getting queue size", http.StatusInternalServerError)
 			return
 		}
+		w.Header().Set("Content-Type", "application/json")
 		if err := json.NewEncoder(w).Encode(cur); err != nil {
 			lggr.Error(err, "encoding QueueCounts")
-			w.WriteHeader(500)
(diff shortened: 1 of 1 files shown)


pkg/queue/queue_rpc.go
```diff
@@ -28,28 +28,12 @@ func newSizeHandler(
...
-					err,
-					"could not send error message to client",
-				)
-			}
+			http.Error(w, "error getting queue size", http.StatusInternalServerError)
 			return
 		}
+		w.Header().Set("Content-Type", "application/json")
 		if err := json.NewEncoder(w).Encode(cur); err != nil {
 			lggr.Error(err, "encoding QueueCounts")
```
wozniakjan, line 36: what happens if the marshalling fails entirely, wouldn't it default to `OK 200` then?

Maybe it should also
```
http.Error(w, "error encoding queue counts", http.StatusInternalServerError)
```

wozniakjan (approved):
lgtm

although, I find the "superfluous response.WriteHeader write" an acceptable cost here. When partial response write fails, interceptor will log extra warning but when encode fails, interceptor will set the correct error response to clients. Or is there any better way?

pkg/queue/queue_rpc.go
linkvt: Marshalling can't really fail as it is a map of simple structs with numbers but I agree that being defensive doesn't hurt here 👍

PR #1581, opened 2026-04-17
refactor: simplify queue count types and fix RPS naming

linkvt:
Simplify Counts from a wrapper struct with embedded interfaces to a plain type alias (map[string]Count), use the map directly.
Rename Count.RPS to RequestRate and move it out of the shared Count type into a scaler-internal aggregatedCount, since the interceptor never uses it and since it was a bit confusing.
Add JSON tags to Count fields to be explicit.

### Changes
- Replace Counts struct with type alias map[string]Count
- Remove unused RPS from exposed queue endpoint
- Introduce scaler-local aggregatedCount for computed rate data
- Remove unused code and its tests
- Add json struct tags to Count fields

### Checklist


pkg/queue/queue_rpc.go
```diff
@@ -46,15 +46,15 @@ func newSizeHandler(
...
 	if err != nil {
 		return nil, fmt.Errorf("requesting the queue counts from %s: %w", interceptorURL.String(), err)
 	}
 	defer func() { _ = resp.Body.Close() }()
-	counts := NewCounts()
-	if err := json.NewDecoder(resp.Body).Decode(counts); err != nil {
+	var counts Counts
+	if err := json.NewDecoder(resp.Body).Decode(&counts); err != nil {
 		return nil, fmt.Errorf("decoding response from the interceptor at %s: %w", interceptorURL.String(), err)
 	}
```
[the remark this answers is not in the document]
linkvt: preexisting

pkg/queue/queue_fakes.go
```diff
@@ -92,32 +92,25 @@ func (f *FakeCounter) RemoveKey(host string) bool {
...
 
-func (f *FakeCounter) Current() (*Counts, error) {
-	ret := NewCounts()
+func (f *FakeCounter) Current() (Counts, error) {
 	f.mapMut.RLock()
 	defer f.mapMut.RUnlock()
-	retMap := f.RetMap
-	ret.Counts = retMap
-	return ret, nil
+	return f.RetMap, nil
```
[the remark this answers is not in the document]
linkvt: only used in tests

pkg/queue/queue_counts.go
```diff
@@ -1,63 +1,13 @@
...
-// Concurrency and RequestCount; the scaler computes RPS.
+// Count is a snapshot of the HTTP pending request concurrency
+// and the raw monotonic request counter, as reported by an
+// interceptor pod.
 type Count struct {
-	Concurrency  int
-	RequestCount int64
-	// TODO(v1): the naming "RPS" is incorrect, it is rather "Rate" or "RequestRate"
-	RPS float64
+	Concurrency  int   `json:"Concurrency"`
```
Fedosin, line 7: Do we really need json tags here? I'd suggest to keep it as 
```go
type Count struct {
	Concurrency  int
	RequestCount int64
}
```

pkg/queue/queue_counts.go
```diff
@@ -1,63 +1,13 @@
...
-// UnmarshalJSON implements json.Unmarshaler
-func (q *Counts) UnmarshalJSON(data []byte) error {
-	return json.Unmarshal(data, &q.Counts)
-}
-
-// String implements fmt.Stringer
-func (q *Counts) String() string {
-	return fmt.Sprintf("%v", q.Counts)
-}
+type Counts = map[string]Count
```
Fedosin: Can we create a new type instead of making an alias?
```go
type Counts map[string]Count
```

Fedosin (commented):
Looks good. Just a couple of small nits.

pkg/queue/queue_counts.go
linkvt: We dont need them but the current state shows why being explicit is better than implicit. We use camelCase everywhere else but someone forgot to specify `json:"concurrency"` in the past causing the default to be used here. It might not be obvious to the reader that this struct is serialized and I didn't want to add a breaking change there yet considering we might switch to pushing these via GRPC instead.

pkg/queue/queue_counts.go
linkvt: Good suggestion, fixed and also updated all references using `map[string]Count` 👍

Fedosin (approved):

PR #1583, opened 2026-04-18
refactor/feat(e2e): improve test reliability and add new InterceptorRoute e2e tests

Fedosin:
Refactor the modern e2e test suite for improved reliability, reduced duplication, and broader InterceptorRoute (v1beta1) coverage.

### Refactoring

- **Replace fixed 5s sleeps** in `CreateInterceptorRoute`/`UpdateInterceptorRoute` with polling for the `Ready` status condition — adapts to cluster speed instead of being too slow or too fast
- **Add shared assertion helpers** (`AssertRouteReachesApp`, `AssertRouteRejected`, `AssertStatus`) and apply across all routing/timeout tests, removing ~280 lines of duplicated boilerplate
- **Soften exact replica assertions** in scaling test: use `WaitForReplicaRange` instead of exact count, reducing flakiness from scaler rounding
- **Add WebSocket dialer timeouts** (30s handshake, 10s dial) to prevent indefinite hangs
- **Guard context accessors** with descriptive panic messages instead of bare type assertions
- **Extract shared `InterceptorMetrics`** slice for prometheus/otel tests
- **Add `WaitForAddonReady`** helper for waiting after add-on pod restarts

### New e2e tests

Seven new tests covering previously untested InterceptorRoute areas:

| Test | Area | What it verifies |
|------|------|-----------------|
| `concurrency_scaling_test` | scaling | Scale based on concurrency target (all existing tests use requestRate) |
| `status_condition_test` | status | IR `Ready` condition after creation and after update with `ObservedGeneration` checks |
| `route_deletion_test` | routing | Delete one IR, verify it is rejected while sibling route still works |
| `interceptor_restart_test` | resilience | Kill interceptor pods, verify routing recovers |
| `operator_recovery_test` | resilience | Delete and recreate a ScaledObject, verify routing still functions |
| `multi_route_test` | routing | Two IRs (host-based and path-based) targeting the same backend |
| `validation_test` | validation | CRD validation rejects IRs with no scaling metric or both port+portName |

### Checklist

  - [`README.md`](../README.md)

### DIFF
--- a/test/e2e/default/concurrency_scaling_test.go
+++ b/test/e2e/default/concurrency_scaling_test.go
@@ -1,0 +1,83 @@
+//go:build e2e
+
+package default_test
+
+import (
+	"context"
+	"net/url"
+	"testing"
+	"time"
+
+	kedav1alpha1 "github.com/kedacore/keda/v2/apis/keda/v1alpha1"
+	vegeta "github.com/tsenart/vegeta/v12/lib"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+	"sigs.k8s.io/e2e-framework/pkg/features"
+
+	h "github.com/kedacore/http-add-on/test/helpers"
+)
+
+func TestConcurrencyScaling(t *testing.T) {
+	t.Parallel()
+
+	var app *h.TestApp
+
+	feat := features.New("concurrency-scaling").
+		WithLabel("area", "scaling").
+		Setup(func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
+			f := h.NewFramework(ctx, t)
+
+			app = f.CreateTestApp("concurrency-app")
+			ir := f.CreateInterceptorRoute("concurrency-ir", app,
+				h.IRWithHosts(f.Hostname()),
+				// Each replica should handle 5 concurrent requests. With
+				// sustained load that keeps ~10 connections open, we expect
+				// around 2 replicas.
+				h.IRWithConcurrency(5),
+			)
+			f.CreateScaledObject("concurrency-so", app, ir, func(so *kedav1alpha1.ScaledObject) {
+				so.Spec.MaxReplicaCount = ptr.To(int32(10))
+			})
+
+			return ctx
+		}).
+		Assess("starts at zero replicas", func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
+			f := h.NewFramework(ctx, t)
+			f.WaitForReplicas(app, 0)
+			return ctx
+		}).
+		Assess("scales up under sustained concurrent load", func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
+			f := h.NewFramework(ctx, t)
+
+			// Use a slow backend (wait=500ms per request) with 10 req/s.
+			// At 10 req/s with 500ms latency, ~5 requests are in-flight
+			// concurrently. With a concurrency target of 5 per replica,
+			// this keeps the system just at the boundary so we expect at
+			// least 1 replica (cold-start proven) and up to 2.
+			stopLoad := f.GenerateLoad(
+				url.URL{Host: f.Hostname(), RawQuery: "wait=500ms"},
+				vegeta.Rate{Freq: 10, Per: time.Second},
+				30*time.Second,
+			)
+
+			f.WaitForReplicas(app, 1)
+
+			metrics := stopLoad()
+			if metrics.Requests == 0 {
+				t.Fatal("no requests were sent during load generation")
+			}
+
+			return ctx
+		}).
+		Assess("scales back to zero after load stops", func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
+			f := h.NewFramework(ctx, t)
+
+			f.WaitForReplicas(app, 0)
+			f.AssertReplicasStable(app, 0, 30*time.Second)
+
+			return ctx
+		}).
+		Feature()
+
+	testenv.Test(t, feat)
+}
--- a/test/e2e/default/path_routing_test.go
+++ b/test/e2e/default/path_routing_test.go
@@ -44,47 +44,17 @@ func TestPathRouting(t *testing.T) {
 		}).
 		Assess("request to /svc-a routes to app-a", func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
 			f := h.NewFramework(ctx, t)
-
-			resp := f.ProxyRequest(h.Request{Host: f.Hostname(), Path: "/svc-a"})
-
-			if resp.StatusCode != 200 {
-				t.Fatalf("expected status 200, got %d; body: %s", resp.StatusCode, resp.Body)
-			}
-			if resp.Hostname != "app-a" {
-				t.Errorf("expected request to be served by app-a, got %q", resp.Hostname)
-			}
-			if resp.RequestPath != "/svc-a" {
-				t.Errorf("expected path /svc-a, got %q", resp.RequestPath)
-			}
-
+			f.AssertRouteReachesApp(h.Request{Host: f.Hostname(), Path: "/svc-a"}, "app-a")
 			return ctx
 		}).
 		Assess("request to non-matching path is rejected", func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
 			f := h.NewFramework(ctx, t)
-
-			resp := f.ProxyRequest(h.Request{Host: f.Hostname(), Path: "/unknown"})
-
-			if resp.StatusCode == 200 {
-				t.Fatalf("expected non-200 status for unmatched path, got 200; body: %s", resp.Body)
-			}
-
+			f.AssertRouteRejected(h.Request{Host: f.Hostname(), Path: "/unknown"})
 			return ctx
 		}).
 		Assess("request to /svc-b routes to app-b", func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
 			f := h.NewFramework(ctx, t)
-
-			resp := f.ProxyRequest(h.Request{Host: f.Hostname(), Path: "/svc-b"})
-
-			if resp.StatusCode != 200 {
-				t.Fatalf("expected status 200, got %d; body: %s", resp.StatusCode, resp.Body)
-			}
-			if resp.Hostname != "app-b" {
-				t.Errorf("expected request to be served by app-b, got %q", resp.Hostname)
-			}
-			if resp.RequestPath != "/svc-b" {
-				t.Errorf("expected path /svc-b, got %q", resp.RequestPath)
-			}
-
+			f.AssertRouteReachesApp(h.Request{Host: f.Hostname(), Path: "/svc-b"}, "app-b")
 			return ctx
 		}).
 		Feature()
--- a/test/e2e/default/route_deletion_test.go
+++ b/test/e2e/default/route_deletion_test.go
@@ -1,0 +1,73 @@
+//go:build e2e
+
+package default_test
+
+import (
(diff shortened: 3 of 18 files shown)


test/helpers/context.go
```diff
@@ -24,23 +24,35 @@ func contextWithClient(ctx context.Context, client klient.Client) context.Contex
 }
 
 func clientFromContext(ctx context.Context) klient.Client {
-	return ctx.Value(clientKey).(klient.Client)
+	v, ok := ctx.Value(clientKey).(klient.Client)
+	if !ok {
+		panic("e2e: klient.Client not found in context — was BeforeEachTest registered?")
+	}
+	return v
```
linkvt: Wouldn't this code panic anyways in its previous version in case there was something wrong?

test/e2e/default/path_routing_test.go
```diff
@@ -44,47 +44,17 @@ func TestPathRouting(t *testing.T) {
...
-				t.Fatalf("expected status 200, got %d; body: %s", resp.StatusCode, resp.Body)
-			}
-			if resp.Hostname != "app-a" {
-				t.Errorf("expected request to be served by app-a, got %q", resp.Hostname)
-			}
-			if resp.RequestPath != "/svc-a" {
-				t.Errorf("expected path /svc-a, got %q", resp.RequestPath)
-			}
-
+			f.AssertRouteReachesApp(h.Request{Host: f.Hostname(), Path: "/svc-a"}, "app-a")
```
linkvt, line 47: We should probably keep the RequestPath checks, right now they would be lost if I see it right.
Same below in the other test.

test/helpers/traffic.go
```diff
@@ -95,13 +96,58 @@ func (f *Framework) ProxyRequestRaw(r Request) HTTPResponse {
 	}
 }
 
+// AssertRouteReachesApp sends a request through the proxy and verifies it is
+// routed to the expected backend app (by whoami hostname). Fails the test on
+// non-200 status or wrong backend.
```
linkvt: nit: Comment says it fails on wrong backend, but we just Error here. I'm also not sure if we need the whoami reference in the comment, the types are kept generic so that we could replace the image in the future.

test/e2e/default/status_condition_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+	feat := features.New("status-conditions").
+		WithLabel("area", "status").
+		Setup(func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
+			f := h.NewFramework(ctx, t)
+
+			app := f.CreateTestApp("status-app")
+			ir = f.CreateInterceptorRoute("status-ir", app,
+				h.IRWithHosts(f.Hostname()),
+			)
```
linkvt: This function and `UpdateInterceptorRoute` already call `waitForInterceptorRouteReady` so both `Assess` steps are basically just check a condition we already checked.

As these two functions are used throughout the test suite I think we can remove this test?

test/e2e/default/scaling_test.go
```diff
@@ -49,14 +49,15 @@ func TestScaling(t *testing.T) {
...
+			// this should stabilize around 2 replicas (+/-1 from scaler rounding
+			// and timing).
 			stopLoad := f.GenerateLoad(
 				url.URL{Host: f.Hostname()},
 				vegeta.Rate{Freq: 10, Per: time.Second},
 				30*time.Second,
 			)
 
-			f.WaitForReplicas(app, 2)
+			f.WaitForReplicaRange(app, 1, 3)
```
linkvt: Was this actually a problem? E2E tests ran always successful over the last few weeks here in CI and locally, there was never a rounding error once. Timing should be covered by the wait function.
I'm not sure if we worsen things here by accepting 1 replica when 2 were expected.

test/helpers/traffic.go
```diff
@@ -95,13 +96,58 @@ func (f *Framework) ProxyRequestRaw(r Request) HTTPResponse {
...
+
+// AssertRouteRejected sends a request through the proxy and verifies it is
+// NOT served successfully (non-200 status code). Fails the test if the
+// request unexpectedly succeeds.
+func (f *Framework) AssertRouteRejected(r Request) {
+	f.t.Helper()
+
+	resp := f.ProxyRequestRaw(r)
+	if resp.StatusCode == http.StatusOK {
+		f.t.Fatalf("expected non-200 status, got 200; body: %s", string(resp.Body))
```
linkvt, line 125: nit: preexisting but we should maybe fail if we get anything except 404 Not Found in case there are other issues triggering a 500 or a redirect or such.

test/helpers/interceptorroute.go
```diff
@@ -143,7 +144,44 @@ func (f *Framework) UpdateInterceptorRoute(ir *httpv1beta1.InterceptorRoute, opt
...
-	// The interceptor picks up route changes asynchronously via its informer.
-	// Sleep to give it time to reload the route before tests send requests.
-	time.Sleep(5 * time.Second)
+	f.waitForInterceptorRouteReady(ir)
+}
+
+// waitForInterceptorRouteReady polls the InterceptorRoute until the operator
+// has reconciled it (Ready condition = True), then waits briefly for the
+// interceptor's informer to pick up the route. This replaces a fixed sleep
+// and adapts to cluster speed.
```
linkvt: nit: we should IMO keep comments evergreen and not reference previous solutions the future reader doesn't see

test/e2e/default/operator_recovery_test.go
```diff
@@ -0,0 +1,103 @@
+//go:build e2e
```
linkvt: I don't think that this tests anything HTTP Addon related as the SO is only managed by KEDA/the user in the future (don't think we need to add more tests for the deprecated HTTPSO).

test/e2e/default/route_deletion_test.go
```diff
@@ -0,0 +1,80 @@
...
+			// Wait for the IR to be fully removed from the API server,
+			// then allow the interceptor's informer to propagate the deletion.
+			deleted := &httpv1beta1.InterceptorRoute{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      irA.Name,
+					Namespace: irA.Namespace,
+				},
+			}
+			if err := wait.For(
+				conditions.New(cfg.Client().Resources()).ResourceDeleted(deleted),
```
linkvt: Can't we just pass irA, no need for a new deleted route I think?

test/e2e/default/interceptor_restart_test.go
```diff
@@ -0,0 +1,67 @@
+//go:build e2e
```
linkvt: I'm not sure about the value of this test, what do we actually test here?
That the routing table is rebuilt? Already tested in every test creating an IR.
That pods are able to start? same

test/helpers/traffic.go
```diff
@@ -95,13 +96,58 @@ func (f *Framework) ProxyRequestRaw(r Request) HTTPResponse {
...
+	}
+}
+
 // WebSocketDial opens a WebSocket connection through the interceptor proxy.
 // The connection is closed automatically via t.Cleanup.
 func (f *Framework) WebSocketDial(host, path string) *websocket.Conn {
 	f.t.Helper()
 
+	dialer := websocket.Dialer{
+		HandshakeTimeout: 30 * time.Second,
```
linkvt: Default is 45s, is this intended?

test/helpers/traffic.go
```diff
@@ -95,13 +96,58 @@ func (f *Framework) ProxyRequestRaw(r Request) HTTPResponse {
...
+		NetDial: (&net.Dialer{
+			Timeout: 10 * time.Second,
+		}).Dial,
```
linkvt: nit: We could probably use NetDialContext here so that aborted stuck tests don't wait for the timeout. Not that big of a change with the timeout at 10s but if we're already touching it :D

test/e2e/default/concurrency_scaling_test.go
```diff
@@ -0,0 +1,81 @@
...
+			// Use a slow backend (wait=500ms per request) with 10 req/s to keep
+			// ~5 concurrent connections open per second, pushing concurrency
+			// above the target of 5 per replica.
+			stopLoad := f.GenerateLoad(
+				url.URL{Host: f.Hostname(), RawQuery: "wait=500ms"},
+				vegeta.Rate{Freq: 10, Per: time.Second},
+				30*time.Second,
+			)
+
+			f.WaitForMinReplicas(app, 1)
```
linkvt: This only tests the cold start, shouldnt we verify that there are actually roughly the replicas we would expect for the calculation?
IIRC I didn't add the test when rewriting the test suite as it is more inconsistent (as concurrency is really only the currently running reqs) than the requestRate based test.

linkvt (commented):
Good idea extracting the assertions, they were really quite repetitive. I left quite a bunch of comments, hope that they make sense :)

test/helpers/context.go
Fedosin: You're right, it panics either way. Reverted this change.

test/e2e/default/path_routing_test.go
Fedosin: Good catch. `AssertRouteReachesApp` now checks `resp.RequestPath` when the request has a non-empty `Path`, so path assertions are preserved automatically.

test/helpers/traffic.go
Fedosin: Fixed both: removed the whoami reference from the comment, and changed `t.Errorf` to `t.Fatalf` so the test actually fails immediately as advertised.

test/e2e/default/status_condition_test.go
Fedosin: Agreed, removed the test. The Ready condition is already verified inside `CreateInterceptorRoute`/`UpdateInterceptorRoute` via `waitForInterceptorRouteReady`.

test/e2e/default/scaling_test.go
Fedosin: Fair point, reverted to `WaitForReplicas(app, 2)`. Removed the `WaitForReplicaRange` helper entirely.

test/helpers/traffic.go
Fedosin: Good idea, will address in a follow-up.

test/helpers/interceptorroute.go
Fedosin: Good point. Updated the comment to describe what it does without referencing the old approach.

test/e2e/default/operator_recovery_test.go
Fedosin: Makes sense, removed the test. ScaledObject lifecycle is managed by KEDA/the user, not the HTTP add-on operator.

test/e2e/default/route_deletion_test.go
Fedosin: Done, passing `irA` directly now.

test/e2e/default/interceptor_restart_test.go
Fedosin: Agreed, removed the test. Routing table rebuild and pod startup are implicitly tested by every test that creates an IR.

test/helpers/traffic.go
Fedosin: Not intended — changed to 45s to match the gorilla default.

test/helpers/traffic.go
Fedosin: Done, switched to `NetDialContext`.

test/e2e/default/concurrency_scaling_test.go
Fedosin: Updated to `WaitForReplicas(app, 1)` with a more accurate comment. With 10 req/s and 500ms latency, concurrency hovers around 5 which is exactly the target, so 1 replica is the expected steady state.

linkvt (approved):

PR #1585, opened 2026-04-18
feat(interceptor): add direct-to-pod routing for upstreams

### LINKED ISSUE #1473: Evaluate sending requests to Endpoint IPs directly

AtharvaPakade:
### Key changes:
  New `KEDA_HTTP_DIRECT_POD_ROUTING` environment variable
  - Adds DirectPodRoutingMode type with two values: `disabled` (default) and `cold-start-only`.
  - On cold start, the interceptor waits for a pod to become ready then rewrites the upstream URL directly to that pod's ip:port, bypassing the service ClusterIP and avoiding latency from kube-proxy rules that are slow to propagate after scale-from-zero.
  
  `ReadyEndpointsCache` now tracks (ip, port) pairs
  - `WaitForReady` now returns a podHost (ip:port string) in addition to the isColdStart bool. Stores ready endpoint state as a map of **`serviceKey → {ip, ports-by-name}`** backed by EndpointSlice updates, replacing the previous bool-only ready flag.
  - Port lookup is keyed on the named port from the InterceptorRoute's PortName field, so multi-port pods route to thecorrect container port.
  - Returns podHost="" when PortName has no match, leaving the upstream URL unchanged.

  Cold-start URL rewrite in EndpointResolver
  - When DirectPodOnColdStart=true, isColdStart=true, and podHost!="", the upstream URL's Host is replaced with the pod ip:port.
  - The rewrite is skipped when the request falls back to an alternate upstream — the fallback URL is not overwritten.
  - For HTTPS fallback upstreams, the TLS server name context is updated to the fallback hostname.

  TLS SNI preservation
  - Routing middleware now stores the intended TLS server name in context before any downstream middleware can rewrite the upstream URL.
  - `TransportPool.Get` now takes a **serverName** argument and keys transports on `(responseHeaderTimeout, serverName)`, applying `tls.Config.ServerName` per transport.
  - Upstream handler passes `util.UpstreamServerNameFromContext` to `transportPool.Get.`
  
### Checklist

  - [`README.md`](../README.md)

Fixes #1473

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,10 +34,13 @@ This changelog keeps track of work items that have been completed and are ready 
 ### New
 
 - **General**: TODO ([#TODO](https://github.com/kedacore/http-add-on/issues/TODO))
+- **Interceptor**: Add `KEDA_HTTP_DIRECT_POD_ROUTING` environment variable (`true` | `false`, default `false`). When enabled, the interceptor routes requests directly to a ready pod IP instead of through the Service ClusterIP, bypassing kube-proxy and other Service-layer features (Service-level NetworkPolicy, session affinity, topology-aware routing). ([#1473](https://github.com/kedacore/http-add-on/issues/1473))
 
 ### Improvements
 
-- **General**: TODO ([#TODO](https://github.com/kedacore/http-add-on/issues/TODO))
+- **Interceptor**: The forwarding transport sets TLS `ServerName` per-dial via `DialTLSContext`, using the original service hostname captured in context, so SNI stays correct when the upstream URL is rewritten to a pod IP. ([#1473](https://github.com/kedacore/http-add-on/issues/1473))
+- **Interceptor**: TLS server name is captured in context by the routing middleware before any URL rewrites, so downstream transports always use the original service hostname for SNI. The routing middleware also resolves the upstream named port into context to drive direct-pod target selection. ([#1473](https://github.com/kedacore/http-add-on/issues/1473))
+- **Interceptor**: `ReadyEndpointsCache` now tracks full `(ip, port)` pairs per named port from EndpointSlices, enabling direct-pod routing (replaces the previous bool-only ready state). ([#1473](https://github.com/kedacore/http-add-on/issues/1473))
 
 ### Fixes
 
--- a/interceptor/config/serving.go
+++ b/interceptor/config/serving.go
@@ -53,6 +53,10 @@ type Serving struct {
 	EnableColdStartHeader bool `env:"KEDA_HTTP_ENABLE_COLD_START_HEADER" envDefault:"true"`
 	// LogRequests enables/disables logging of incoming requests
 	LogRequests bool `env:"KEDA_HTTP_LOG_REQUESTS" envDefault:"false"`
+	// DirectPodOnColdStart routes requests directly to a ready pod IP (instead of
+	// the ClusterIP service) when a cold start is detected. Use this when the
+	// service mesh sidecar may not be ready at pod startup time.
+	DirectPodOnColdStart bool `env:"KEDA_HTTP_DIRECT_POD_ON_COLD_START" envDefault:"false"`
 }
 
 // MustParseServing parses standard configs and returns the
--- a/interceptor/config/serving.go
+++ b/interceptor/config/serving.go
@@ -65,6 +65,10 @@ type Serving struct {
 	// server continues serving in-flight and new requests, giving Kubernetes
 	// time to propagate endpoint removal.
 	ShutdownDelay time.Duration `env:"KEDA_HTTP_SHUTDOWN_DELAY" envDefault:"5s"`
+	// DirectPodRouting routes requests to a ready pod IP instead of the Service
+	// ClusterIP, bypassing kube-proxy and other Service-layer features
+	// (NetworkPolicy, session affinity, topology-aware routing). Single-stack only.
+	DirectPodRouting bool `env:"KEDA_HTTP_DIRECT_POD_ROUTING" envDefault:"false"`
 }
 
 // MustParseServing parses standard configs and returns the
--- a/interceptor/middleware/endpoint_resolver_test.go
+++ b/interceptor/middleware/endpoint_resolver_test.go
@@ -424,3 +424,245 @@ func addReadyEndpoint(cache *k8s.ReadyEndpointsCache) {
 		{Endpoints: []discov1.Endpoint{{Addresses: []string{"1.2.3.4"}}}},
 	})
 }
+
+func addReadyEndpointWithPort(cache *k8s.ReadyEndpointsCache, port int32) {
+	cache.Update(testNamespace+"/"+testService, []*discov1.EndpointSlice{
+		{
+			Ports:     []discov1.EndpointPort{{Port: &port}},
+			Endpoints: []discov1.Endpoint{{Addresses: []string{"1.2.3.4"}}},
+		},
+	})
+}
+
+func addReadyEndpointWithNamedPorts(cache *k8s.ReadyEndpointsCache, ports []discov1.EndpointPort) {
+	cache.Update(testNamespace+"/"+testService, []*discov1.EndpointSlice{
+		{
+			Ports:     ports,
+			Endpoints: []discov1.Endpoint{{Addresses: []string{"1.2.3.4"}}},
+		},
+	})
+}
+
+// TestEndpointResolver_DirectPodOnColdStart_UsedOnColdStart verifies that when
+// DirectPodOnColdStart=true and the backend undergoes a cold start, the upstream
+// URL is rewritten to the pod IP:containerPort.
+func TestEndpointResolver_DirectPodOnColdStart_UsedOnColdStart(t *testing.T) {
+	cache := k8s.NewReadyEndpointsCache(logr.Discard())
+	// Start with no ready endpoints to force a cold start.
+
+	var capturedUpstream *url.URL
+	next := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		capturedUpstream = util.UpstreamURLFromContext(r.Context())
+		w.WriteHeader(http.StatusOK)
+	})
+
+	mw := NewEndpointResolver(next, cache, EndpointResolverConfig{
+		ReadinessTimeout:     2 * time.Second,
+		DirectPodOnColdStart: true,
+	})
+
+	rec := httptest.NewRecorder()
+	req := newRequest(t, defaultIR())
+
+	// Add endpoint with a known container port after the middleware starts waiting.
+	go func() {
+		time.Sleep(50 * time.Millisecond)
+		addReadyEndpointWithPort(cache, 8080)
+	}()
+
+	mw.ServeHTTP(rec, req)
+
+	if rec.Code != http.StatusOK {
+		t.Fatalf("status = %d, want %d", rec.Code, http.StatusOK)
+	}
+	if capturedUpstream == nil {
+		t.Fatal("upstream URL should not be nil")
+	}
+	if capturedUpstream.Host != "1.2.3.4:8080" {
+		t.Fatalf("upstream host = %q, want %q", capturedUpstream.Host, "1.2.3.4:8080")
+	}
+}
+
+// TestEndpointResolver_DirectPodOnColdStart_MultiPort verifies that when the pod
+// exposes multiple ports (e.g. metrics first, server second), direct-pod routing
+// picks the container port that matches the InterceptorRoute's PortName — not
+// whatever happens to be first in the EndpointSlice.
+func TestEndpointResolver_DirectPodOnColdStart_MultiPort(t *testing.T) {
+	cache := k8s.NewReadyEndpointsCache(logr.Discard())
+
+	metricsName := "metrics"
+	metricsPort := int32(9090)
+	httpName := "http"
+	httpPort := int32(8080)
+
+	var capturedUpstream *url.URL
+	next := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		capturedUpstream = util.UpstreamURLFromContext(r.Context())
+		w.WriteHeader(http.StatusOK)
+	})
+
+	ir := defaultIR()
+	ir.Spec.Target.PortName = "http"
+
+	mw := NewEndpointResolver(next, cache, EndpointResolverConfig{
+		ReadinessTimeout:     2 * time.Second,
+		DirectPodOnColdStart: true,
+	})
+
+	rec := httptest.NewRecorder()
+	req := newRequest(t, ir)
+
+	// metrics port is listed first; server port (http) is second.
+	go func() {
+		time.Sleep(50 * time.Millisecond)
+		addReadyEndpointWithNamedPorts(cache, []discov1.EndpointPort{
+			{Name: &metricsName, Port: &metricsPort},
+			{Name: &httpName, Port: &httpPort},
+		})
+	}()
+
+	mw.ServeHTTP(rec, req)
+
+	if rec.Code != http.StatusOK {
+		t.Fatalf("status = %d, want %d", rec.Code, http.StatusOK)
+	}
+	if capturedUpstream == nil {
+		t.Fatal("upstream URL should not be nil")
+	}
+	if capturedUpstream.Host != "1.2.3.4:8080" {
+		t.Fatalf("upstream host = %q, want %q — should route to http port, not metrics port",
+			capturedUpstream.Host, "1.2.3.4:8080")
+	}
+}
+
+// TestEndpointResolver_DirectPodOnColdStart_EmptyPortName verifies that when the
+// InterceptorRoute uses a numeric Port (PortName="") and the pod has only named
+// ports, direct-pod routing is skipped and the ClusterIP URL is left unchanged.
+func TestEndpointResolver_DirectPodOnColdStart_EmptyPortName(t *testing.T) {
+	cache := k8s.NewReadyEndpointsCache(logr.Discard())
+
+	httpName := "http"
+	httpPort := int32(8080)
+
+	var capturedUpstream *url.URL
+	next := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		capturedUpstream = util.UpstreamURLFromContext(r.Context())
+		w.WriteHeader(http.StatusOK)
+	})
+
+	// IR uses numeric Port — PortName is empty.
+	ir := defaultIR()
+	ir.Spec.Target.PortName = ""
+
+	mw := NewEndpointResolver(next, cache, EndpointResolverConfig{
+		ReadinessTimeout:     2 * time.Second,
+		DirectPodOnColdStart: true,
+	})
+
+	rec := httptest.NewRecorder()
+	req := newRequest(t, ir)
+
+	go func() {
+		time.Sleep(50 * time.Millisecond)
+		addReadyEndpointWithNamedPorts(cache, []discov1.EndpointPort{
+			{Name: &httpName, Port: &httpPort},
+		})
+	}()
+
+	mw.ServeHTTP(rec, req)
+
+	if rec.Code != http.StatusOK {
+		t.Fatalf("status = %d, want %d", rec.Code, http.StatusOK)
+	}
+	if capturedUpstream == nil {
+		t.Fatal("upstream URL should not be nil")
+	}
+	// PortName="" with only named ports → PickReadyEndpoint returns ok=false →
+	// ClusterIP URL must be left unchanged.
+	if capturedUpstream.Host != "upstream" {
+		t.Fatalf("upstream host = %q, want %q — should not rewrite when PortName is empty",
+			capturedUpstream.Host, "upstream")
+	}
+}
+
+// TestEndpointResolver_DirectPodOnColdStart_UnnamedPort verifies that when the
+// InterceptorRoute uses a numeric Port (PortName="") and the EndpointSlice port
+// is also unnamed (Name==nil), direct-pod routing succeeds because both sides use
+// the "" key.
+func TestEndpointResolver_DirectPodOnColdStart_UnnamedPort(t *testing.T) {
+	cache := k8s.NewReadyEndpointsCache(logr.Discard())
+
+	var capturedUpstream *url.URL
+	next := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		capturedUpstream = util.UpstreamURLFromContext(r.Context())
+		w.WriteHeader(http.StatusOK)
+	})
+
+	// IR uses numeric Port — PortName is empty.
+	ir := defaultIR()
+	ir.Spec.Target.PortName = ""
+
+	mw := NewEndpointResolver(next, cache, EndpointResolverConfig{
+		ReadinessTimeout:     2 * time.Second,
+		DirectPodOnColdStart: true,
+	})
+
+	rec := httptest.NewRecorder()
+	req := newRequest(t, ir)
+
+	go func() {
+		time.Sleep(50 * time.Millisecond)
+		// Port with Name==nil → stored under "" in the ports map.
+		addReadyEndpointWithPort(cache, 8080)
+	}()
+
+	mw.ServeHTTP(rec, req)
+
+	if rec.Code != http.StatusOK {
+		t.Fatalf("status = %d, want %d", rec.Code, http.StatusOK)
+	}
+	if capturedUpstream == nil {
+		t.Fatal("upstream URL should not be nil")
+	}
+	// PortName="" with unnamed EndpointSlice port → lookup succeeds → URL rewritten.
+	if capturedUpstream.Host != "1.2.3.4:8080" {
+		t.Fatalf("upstream host = %q, want %q — unnamed port should match empty PortName",
+			capturedUpstream.Host, "1.2.3.4:8080")
+	}
+}
+
+// TestEndpointResolver_DirectPodOnColdStart_NotUsedOnWarmPath verifies that when
+// DirectPodOnColdStart=true but the backend is already warm (isColdStart=false),
+// the upstream URL is NOT rewritten to the pod IP.
+func TestEndpointResolver_DirectPodOnColdStart_NotUsedOnWarmPath(t *testing.T) {
+	cache := k8s.NewReadyEndpointsCache(logr.Discard())
+	// Pre-populate so WaitForReady returns immediately with isColdStart=false.
+	addReadyEndpointWithPort(cache, 8080)
+
+	var capturedUpstream *url.URL
+	next := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		capturedUpstream = util.UpstreamURLFromContext(r.Context())
+		w.WriteHeader(http.StatusOK)
+	})
+
+	mw := NewEndpointResolver(next, cache, EndpointResolverConfig{
+		ReadinessTimeout:     2 * time.Second,
+		DirectPodOnColdStart: true,
+	})
+
+	rec := httptest.NewRecorder()
+	req := newRequest(t, defaultIR())
+	// upstream URL is set to "upstream" in newRequest
+	mw.ServeHTTP(rec, req)
+
+	if rec.Code != http.StatusOK {
+		t.Fatalf("status = %d, want %d", rec.Code, http.StatusOK)
+	}
+	if capturedUpstream == nil {
+		t.Fatal("upstream URL should not be nil")
+	}
+	// Must still be the original service URL, not the pod IP
+	if capturedUpstream.Host != "upstream" {
+		t.Fatalf("upstream host = %q, want %q (should not be rewritten on warm path)", capturedUpstream.Host, "upstream")
+	}
+}
--- a/interceptor/middleware/endpoint_resolver_test.go
+++ b/interceptor/middleware/endpoint_resolver_test.go
@@ -133,6 +133,105 @@ func TestEndpointResolver_Fallback(t *testing.T) {
 	}
 }
 
+// TestEndpointResolver_DirectPodRouting_DoesNotOverwriteFallback verifies that
+// when the backend never becomes ready and we fall back to the alternate upstream,
+// the direct-to-pod rewrite does not overwrite the fallback URL even when
+// DirectPodRouting is enabled.
+func TestEndpointResolver_DirectPodRouting_DoesNotOverwriteFallback(t *testing.T) {
+	cache := k8s.NewReadyEndpointsCache(logr.Discard())
+	// Do not mark ready — backend has no replicas, forcing the fallback path.
+
+	fallbackURL := &url.URL{Scheme: "http", Host: "fallback-svc"}
+
+	var capturedUpstream *url.URL
+	next := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		capturedUpstream = util.UpstreamURLFromContext(r.Context())
+		w.WriteHeader(http.StatusOK)
+	})
+
+	ir := defaultIR()
+	ir.Spec.ColdStart = &httpv1beta1.ColdStartSpec{
+		Fallback: &httpv1beta1.ColdStartFallback{
+			Service: &httpv1beta1.ServiceRef{Name: "fallback-svc"},
+		},
+	}
+
+	mw := NewEndpointResolver(next, cache, EndpointResolverConfig{
+		ReadinessTimeout: 25 * time.Millisecond,
+		DirectPodRouting: true,
+	})
+
+	rec := httptest.NewRecorder()
+	req := newRequest(t, ir)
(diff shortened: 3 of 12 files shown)


interceptor/config/serving.go
```diff
@@ -47,12 +47,20 @@ type Serving struct {
 	// TLSCurvePreferences is a comma-separated list of elliptic curve names
 	// (e.g. "X25519,CurveP256"). If empty, the default Go curve preferences are used.
 	TLSCurvePreferences string `env:"KEDA_HTTP_PROXY_TLS_CURVE_PREFERENCES" envDefault:""`
+	// TLSUpstreamDisabled disables TLS for upstream connections even when the proxy server
+	// itself has TLS enabled. When true, requests are forwarded to upstream pods over plain
+	// HTTP regardless of the proxy TLS setting.
+	TLSUpstreamDisabled bool `env:"KEDA_HTTP_PROXY_TLS_UPSTREAM_DISABLED" envDefault:"false"`
```
linkvt: Continuing here from our chat in Slack:

As far as I understand whether TLS is used is not relevant, we just change the target the traffic is sent to. When TLS is used we can if I understand it correctly set the `tlsConfig.ServerName` to the service name while still sending our request to the Endpoint IP, see https://pkg.go.dev/crypto/tls#Config
```
	// ServerName is used to verify the hostname on the returned
	// certificates unless InsecureSkipVerify is given. It is also included
	// in the client's handshake to support virtual hosting unless it is
	// an IP address.
	ServerName string
```

AtharvaPakade: @linkvt 
Correct, we can use `ServerName` (it will be the SNI) which can make the direct to pod IP TLS handshake successful. 

I have a few implementations in mind. Can you help me figure out which one I should proceed with?

A. The base transport is built at startup, and the upstream handler uses it to seed a pool keyed by responseHeaderTimeout. Should we extend that same pool to key by (timeout, serverName) combinations? This increases pool cardinality, and since serverName-aware transports are only needed during cold start — is that trade-off worth it?

B. Rather than extending the existing pool, have a dedicated cold start transport pool keyed by (timeout, serverName). This keeps the hot path unchanged and scopes the cardinality growth to cold start scenarios only.

C. Use a temporary, non-pooled transport for cold start requests — created on demand, not reused. This leaves the existing pool entirely untouched, but introduces transport selection logic in the upstream handler to switch between pooled (normal) and ephemeral (cold start) transports.

linkvt: Thanks for looking into this, I didn't consider that it would be more complicated 😅 
So I think that means we need to reconsider the current TransportPool implementation, A sounds IMO good, we also don't need it just for the cold-start but for all connections I think?
The direct pod routing is useful for cold-starts where syncing the Services ClusterIP might take some time, but can also be used later as well for non cold-start requests.

Pool size is something we could consider now or later in a follow-up PR, the pool will definitely be larger but it's just a few KBs anyway IIRC.

AtharvaPakade: > we also don't need it just for the cold-start but for all connections I think?

For this PR, IMO we should only optimize direct-to-pod for cold start request. 
But once we have seperate PR which detect readiness based by probing, like I mentioned in this [issue comment](https://github.com/kedacore/http-add-on/issues/1473#issuecomment-4274603525) then we can consider this optimization for all requests. 

Additionally, i will also drop this `KEDA_HTTP_PROXY_TLS_UPSTREAM_DISABLED` flag. I think we also need a PR that give ability to configure the Upstream TLS config . Because currently we use the same TLS config for upstream. Having this can also give user ability for mTLs with upstream pods.

linkvt: Couldn't we just use/modify the current
```
	isColdStart, err := er.readyCache.WaitForReady(waitCtx, serviceKey)
```
to also return an endpoint IP?
We would not do any extra work at all then and could use the endpoint IP in both cases?

AtharvaPakade: Using `WaitForReady` makes sense only when we want to also do direct-to-pod for non cold start request as well.
@linkvt Are you proposing that we do direct-to-pod for all request in this PR itself ?

linkvt: Yes I thought that would be easier, no special handling for cold start, always take the endpoint IP from our readyCache. Add a flag to opt-out (nobody does opt-in usually) to fallback to the non endpoint upstream.
WaitForReady is IIRC called in both cold and non coldstart scenarios and would be an ideal place to change?

also wdyt @Fedosin ?

Fedosin (commented):
## Review

### Design-Level Feedback

**1. `KEDA_HTTP_PROXY_TLS_UPSTREAM_DISABLED` should be dropped (agree with @linkvt)**

This flag globally disables TLS to upstream for *all* requests, not just cold-start direct-pod ones. That's a blunt instrument with security implications. The proper solution is to set `tlsConfig.ServerName` to the original service DNS name when dialing a pod IP directly. That way TLS works correctly without needing a kill-switch. I'd suggest either:
- Handling TLS via `ServerName` in this PR, or
- Deferring TLS+direct-pod to a follow-up and only enabling direct-pod for non-TLS setups here.

**2. Scope: cold-start-only vs. all requests**

@linkvt asked for my opinion. I think starting with cold-start-only is the right incremental approach: it has a clear problem statement (kube-proxy lag) and limits blast radius. However, the API surface should be designed with the eventual "all requests" path in mind. Consider whether the flag name `KEDA_HTTP_DIRECT_POD_ON_COLD_START` locks us in, or if a more general name (e.g. `KEDA_HTTP_DIRECT_POD_ROUTING` with mode values like `cold-start-only` / `always` / `disabled`) would be better.

That said, keeping the current flag and adding a separate one later is acceptable too.

**3. `WaitForReady` could return endpoint info (agree with @linkvt)**

Since `WaitForReady` already blocks until the service has ready endpoints, it's the natural point where we *know* a pod is ready. Having it also return a candidate endpoint would:
- Eliminate the TOCTOU gap between "became ready" and `PickReadyEndpoint`
- Simplify the endpoint resolver middleware (no separate `PickReadyEndpoint` call)
- Enable future "always direct-pod" mode trivially

This doesn't need to happen in this PR, but would be a cleaner design for the next iteration.

---

### Code-Level Issues

**4. `crypto/rand` is overkill for load-balancing selection**

```go
idx, err := rand.Int(rand.Reader, big.NewInt(int64(len(candidates))))
if err != nil {
    return "", 0, false
}
```

Using `crypto/rand` + `big.Int` for a simple random index in a hot path adds unnecessary overhead (syscall + heap allocation per call). `math/rand/v2` (already used in this repo) with its auto-seeded global generator is sufficient for load-balancing randomness. The crypto/rand error path also silently falls back to ClusterIP, which could mask a real system issue (exhausted entropy).

**5. The `Update` hot path has a subtle ordering concern**

```go
if v, ok := c.states.Load(serviceKey); ok {
    ptr := v.(*atomic.Pointer[serviceState])
    oldState := ptr.Load()
    ptr.Store(newState)
    ...
}
```

If two concurrent `Update` calls race on the same key, the non-atomic Load-then-Store means the second writer's state could silently overwrite the first's. In practice informer event handlers are serialized per-resource, so this is likely safe today, but it's fragile. A `CompareAndSwap` loop or brief per-key lock would be more robust and future-proof.

**6. `PickReadyEndpoint` returns `ok=false` when `portName=""` and all ports are named**

This means users who configure their `InterceptorRoute` with a numeric `Port` (resulting in `PortName=""`) will *never* get direct-pod routing even when the feature is enabled. This limitation should be documented clearly so users know they must use named ports for direct-pod routing to activate.

**7. `schemeHTTPS` / `schemeHTTP` constants placement**

These constants are defined in `routing.go` but referenced from `endpoint_resolver.go` (same package, so it compiles). Consider placing shared constants in a dedicated file or at the package level to make the dependency explicit and avoid confusion when reading either file in isolation.

---

### Minor Nits

- The startup validation in `main.go` becomes unnecessary once `TLSUpstreamDisabled` is dropped. If you keep it, note that `runtime.Goexit()` after logging is consistent with the rest of `main.go` but a brief comment explaining why (deferred cleanup) would help future readers.
- Missing changelog entry and documentation (per the unchecked checklist items).

---

### What's Good

- The `serviceState` immutable snapshot approach is well-designed: atomic pointer swap gives readers lock-free access without contention.
- The `(ip, port)` pairing per-slice in `collectServiceState` correctly handles rolling deploys where different slices expose different port numbers for the same port name.
- Deduplication logic prevents the same `(ip, port)` pair from appearing multiple times across slices.
- Test coverage is thorough: multi-port, unnamed port, empty port name, heterogeneous ports, and the rolling-deploy scenario are all exercised.
- Graceful fallback to ClusterIP when `PickReadyEndpoint` returns `ok=false` is safe and correct.

---

### Summary

The core idea and cache refactoring are solid. Main blockers before merge:
1. Resolve the TLS approach (drop `TLSUpstreamDisabled`, use `ServerName` or scope to non-TLS only).
2. Switch `crypto/rand` to `math/rand/v2`.
3. Add changelog + document the named-port requirement.

AtharvaPakade:
Hi @linkvt, @Fedosin, thanks for having a look at the PR! I've pushed the changes based on your suggestions. Please let me know if there's anything else to address.

Fedosin (commented):
## Review

The code looks good overall and I don't see any reason not to merge it except the IPv6 incompatibility in `pickHost`. All previous review feedback has been addressed.

### Issues

**1. IPv6 compatibility in `pickHost` (medium)**

`pickHost` formats the host as:
```go
return fmt.Sprintf("%s:%d", ep.ip, ep.port)
```
For IPv6 pod addresses (e.g. `fd00::1`), this would produce `fd00::1:8080` instead of the correct `[fd00::1]:8080`. Please use `net.JoinHostPort(ep.ip, strconv.Itoa(int(ep.port)))` instead, which correctly handles both IPv4 and IPv6. Kubernetes dual-stack clusters assign IPv6 pod IPs, so this is a real (if uncommon) scenario.

**2. Truncated comment in `context.go` (nit)**

```go
// ContextWithUpstreamServerName stores the intended TLS server name (SNI) for the upstream.
// This must be set before any middleware rewrites the upstream URL (e.g. direct-to-pod),
```
The second line ends with a comma, suggesting a truncated sentence. Should be completed or end with a period.

**3. Broadcast channel comment accuracy (nit)**

The comment on the `notifyCh` field says "closed on any change", but `swapState` now only broadcasts on readiness **transitions** (`oldState.hasReady() != newState.hasReady()`). The comment should reflect this narrower contract.

interceptor/config/serving.go
```diff
@@ -53,10 +66,21 @@ type Serving struct {
...
 func MustParseServing() Serving {
-	return env.Must(env.ParseAs[Serving]())
+	s := env.Must(env.ParseAs[Serving]())
+	switch s.DirectPodRouting {
+	case DirectPodRoutingDisabled, DirectPodRoutingColdStartOnly:
+		// valid
+	default:
+		panic(fmt.Sprintf("invalid KEDA_HTTP_DIRECT_POD_ROUTING value %q: must be %q or %q",
+			s.DirectPodRouting, DirectPodRoutingDisabled, DirectPodRoutingColdStartOnly))
+	}
```
linkvt: This could be implemented directly close to the struct by implementing the TextMarshaller interface so that MustParseServing stays clean:
```go
func (m *DirectPodRoutingMode) UnmarshalText(text []byte) error {
    switch DirectPodRoutingMode(text) {
    case DirectPodRoutingDisabled, DirectPodRoutingColdStartOnly:
        *m = DirectPodRoutingMode(text)
        return nil
    default:
        return fmt.Errorf("invalid value %q: must be %q or %q",
            text, DirectPodRoutingDisabled, DirectPodRoutingColdStartOnly)
    }
}
```

See https://github.com/caarlos0/env/blob/a72d89a8930fc800372a6a338a1acf33e5cc3a56/example_test.go#L212-L218

interceptor/config/serving.go
```diff
@@ -53,10 +66,21 @@ type Serving struct {
 	EnableColdStartHeader bool `env:"KEDA_HTTP_ENABLE_COLD_START_HEADER" envDefault:"true"`
 	// LogRequests enables/disables logging of incoming requests
 	LogRequests bool `env:"KEDA_HTTP_LOG_REQUESTS" envDefault:"false"`
+	// DirectPodRouting controls when the interceptor routes directly to a pod IP
+	// instead of the ClusterIP service. Valid values: "disabled", "cold-start-only".
+	DirectPodRouting DirectPodRoutingMode `env:"KEDA_HTTP_DIRECT_POD_ROUTING" envDefault:"disabled"`
```
linkvt: Couldnt this be a simple boolean? Internally we later only use a boolean as well.
Is there a reason for why we would use direct pod routing only on cold starts according to this config (didnt check the full PR yet)?

config/crd/bases/http.keda.sh_interceptorroutes.yaml
```diff
@@ -195,15 +195,18 @@ spec:
                 description: Backend service to route traffic to.
                 properties:
                   port:
-                    description: Port number on the Service. Mutually exclusive with
-                      portName.
+                    description: |-
+                      Port number on the Service. Mutually exclusive with portName.
+                      Note: direct-pod routing (when enabled on the interceptor) requires portName;
+                      routes using a numeric port will always be forwarded via the Service ClusterIP.
```
linkvt: This seems like a major restriction and I don't think there is a reason for it?

operator/apis/http/v1beta1/interceptorroute_types.go
```diff
@@ -104,11 +104,14 @@ type TargetRef struct {
 	// +kubebuilder:validation:MinLength=1
 	Service string `json:"service"`
 	// Port number on the Service. Mutually exclusive with portName.
+	// Note: direct-pod routing (when enabled on the interceptor) requires portName;
+	// routes using a numeric port will always be forwarded via the Service ClusterIP.
```
linkvt: This seems like a major restriction, I don't see a reason for it yet to be honest:

I see these cases:
- single port service, unnamed port
    - endpointslice contains only one port with `name: ""`
    - IR uses port number: resolve port name from Service, match by name in EndpointSlice
    - IR uses port name: not possible (service port is unnamed, nothing to reference)
- single port service, named port
    - endpointslice contains only one port where `name` is equal to the name of the port in the service
    - IR uses port number: resolve port name from Service, match by name in EndpointSlice
    - IR uses port name: exact match by name
- multi port service, unnamed ports
    - not possible, Kubernetes requires names when multiple ports are defined: https://kubernetes.io/docs/concepts/services-networking/service/#multi-port-services
- multi port service, named ports
    - endpointslice contains multiple ports, each named after the corresponding service port
    - IR uses port number: resolve port name from Service, match by name in EndpointSlice
    - IR uses port name: exact match by name

In all cases, we can resolve to the correct EndpointSlice port without requiring `portName` in the IR.
The "IR uses port number" cases all work the same way: resolve the numeric port to its port name via the Service object.

The routing middleware (`resolveUpstreamURL`) already does that Service lookup to build the upstream URL, so it could pass the resolved port name down via context at the same time.

pkg/http/transport_pool.go
```diff
@@ -6,35 +6,56 @@ import (
...
+
 // TransportPool manages a pool of nethttp.Transport instances,
-// reusing them based on response header timeout configuration.
+// reusing them based on response header timeout and TLS server name.
+// serverName is only applied when the base transport was explicitly configured
+// with TLS; for plain-HTTP upstreams the key degenerates to timeout-only.
 //
-// NOTE: Transports are never evicted, we expect a low cardinality of timeouts
+// NOTE: Transports are never evicted; we expect a low cardinality of (timeout, serverName) combinations.
 type TransportPool struct {
```
linkvt, line 13: I think all of these transport pool changes could be avoided when we would be using `transport.TLSDialContext` and resolve the ServerName while dialing.
This would be a smaller change than what we're doing right and IMO cleaner, but a bit more complicated.
The base transport is currently created in proxy.go outside of this whole middleware chain, not sure if it should still be placed there after this change.

pkg/k8s/ready_endpoints_cache.go
```diff
@@ -14,14 +14,31 @@ import (
...
+// ReadyEndpointsCache maintains a derived map of service -> serviceState
 // for O(1) hot-path lookups, plus a broadcast mechanism so the cold-start
 // wait function can block until a service becomes ready.
 type ReadyEndpointsCache struct {
 	lggr logr.Logger
 
-	// "namespace/service" -> *atomic.Bool
-	ready sync.Map
+	// "namespace/service" -> *atomic.Pointer[serviceState]
+	states sync.Map
```
linkvt, line 42: I read a bit more into this and noticed trhat sync.Map offers already Swap etc, we don't need atomic wrappers at all as it seems, see https://pkg.go.dev/sync#Map.Swap

pkg/k8s/ready_endpoints_cache.go
```diff
@@ -65,58 +88,101 @@ func (c *ReadyEndpointsCache) WaitForReady(ctx context.Context, serviceKey strin
...
 	if len(endpointSlices) == 0 {
-		c.ready.Delete(serviceKey)
+		c.states.Delete(serviceKey)
 		c.broadcast()
 		return
 	}
 
-	hasReady := slices.ContainsFunc(endpointSlices, hasAnyReadyEndpoint)
+	newState := collectServiceState(endpointSlices)
+
```
linkvt, line 156: I dont understand the new code completely yet but dont understand, why this hot and cold path would be needed?
Can't we just swap and broadcast if ready changes?

linkvt (changes requested):
The initially trivial idea of "just connecting to endpoints instead of cluster ips" really showed to be way more tricky...
I tried to understand all changes but didn't finish the whole PR yet as some findings will again change things quite a bit IMO. I understand this is quite complex, let me know if I should give it a try as well based on these changes to show what I'm thinking about if my comments are a bit cryptic.

interceptor/config/serving.go
AtharvaPakade: @Fedosin Suggested in the review that we should give option for different modes [comment link](https://github.com/kedacore/http-add-on/pull/1585#pullrequestreview-4197887321) section `2. Scope: cold-start-only vs. all requests`

pkg/k8s/ready_endpoints_cache.go
AtharvaPakade: With `sync.Map.Swap`, every update would call Swap on the map itself, which acquires a mutex internally even for existing keys (lock on every update). 

The atomic wrapper is the entire reason we avoid that — the map value is a stable pointer that never changes, only what it points to does (lock only on first insertion).

pkg/k8s/ready_endpoints_cache.go
AtharvaPakade: The hot/cold split exists precisely to keep the hot path lock-free — as mentioned in [comment](https://github.com/kedacore/http-add-on/pull/1585#discussion_r3238721550), if we just called `sync.Map.Swap` directly, we'd acquire a mutex on every endpoint update, defeating the entire point of the atomic wrapper.

The cold path pays the mutex cost once via `LoadOrStore` (first insertion). After that, every update stays on the hot path: `states.Load + ptr.CompareAndSwap` — no map write, no mutex.

pkg/http/transport_pool.go
AtharvaPakade: Agreed, we discussed this in a previous [comment](https://github.com/kedacore/http-add-on/pull/1585#discussion_r3136450866) and based on your recommendation i opted this approach. 

IMO we should have a seperate PR/Issue addressing the ability to configure `TLSDialContext`, `DialContext` and moving base transport outside the whole middleware chain.

operator/apis/http/v1beta1/interceptorroute_types.go
AtharvaPakade: In the `multi port service, named ports` case, there is a possibility that the user specifies port names on the Service but does not provide portName at the time of configuring the InterceptorRoute. The note was targeted at this case to avoid misconfigurations.

Based on your suggestion, I feel this case can also be resolved by resolving the `portName` in the Routing middleware and passing it via context, by adding a separate `resolvePortName` method that does the reverse lookup (numeric port → port name via `Service.Spec.Ports`). In the `EndpointResolver` middleware, we can then refer to the resolved `portName` from context instead of `ir.Spec.Target.PortName` directly.

pkg/k8s/ready_endpoints_cache.go
linkvt: I did more research on that and yes this is correct, but I don't see any impact for us.
We are reacting to events from an informer which are not happening that frequently per key (endpoints of a service don't change that much, at max a few times a second) and are delivered sequentially.
The locking which only happens on scaling events happens only on writes and doesnt affect reads.

I also compared both implementations for the update case locally in a benchmark:
- sync.Map.Swap: takes 25ns 
- sync.Map.Load+atomic.Pointer: takes 9ns

Using these numbers means we need just for the update case an additional 16ns = 0.000000016 seconds which is not really noticeable at all.

The read case happens way more often and took 0.2ns longer using sync.Map+atomic.Pointer than using sync.Map directly.

Considering these numbers and that writing concurrency safe code is really hard, I don't see how this optimization is worth the complexity we add. I trust the Go authors more than us and our LLMs to write safe and fast code and also improve that over time.

pkg/k8s/ready_endpoints_cache.go
linkvt: See my comment above, the cost which only applies to writes is extremely low (few nanoseconds) which doesn't warrant the complexity we add here IMO.

pkg/http/transport_pool.go
linkvt: I see, I think the comment there was based on the proposed options which didn't include the `TLSDialContext` option which seems way simpler IMO as they don't need all of these changes on the transportpool.

I did some research on this as well and it seems to me that `TLSDialContext` is the way to go for per connection TLS configs and is way simpler than the proposed transportpool change, it could look like this in proxy.go:
```diff
--- a/interceptor/proxy.go
+++ b/interceptor/proxy.go
@@ -50,21 +51,22 @@ func BuildProxyHandler(cfg *ProxyHandlerConfig) http.Handler {
 		}
 	}
 
-	var forwardingTLSCfg *tls.Config
-	if cfg.TLSConfig != nil {
-		forwardingTLSCfg = &tls.Config{
-			RootCAs:            cfg.TLSConfig.RootCAs,
-			Certificates:       cfg.TLSConfig.Certificates,
-			InsecureSkipVerify: cfg.TLSConfig.InsecureSkipVerify, //nolint:gosec // G402: user-configurable
-			MinVersion:         cfg.TLSConfig.MinVersion,
-			MaxVersion:         cfg.TLSConfig.MaxVersion,
-			CipherSuites:       cfg.TLSConfig.CipherSuites,
-			CurvePreferences:   cfg.TLSConfig.CurvePreferences,
-		}
-	}
-
 	// Clone DefaultTransport to inherit Go's defaults for IdleConnTimeout, ...
 	transport := http.DefaultTransport.(*http.Transport).Clone()
 	transport.DialContext = dialFunc
 	transport.ForceAttemptHTTP2 = cfg.Timeouts.ForceHTTP2
 	transport.MaxIdleConns = cfg.Timeouts.MaxIdleConns
 	transport.MaxIdleConnsPerHost = cfg.Timeouts.MaxIdleConnsPerHost
-	transport.TLSClientConfig = forwardingTLSCfg
+	if cfg.TLSConfig != nil {
+		baseTLSCfg := cfg.TLSConfig
+		transport.DialTLSContext = func(ctx context.Context, network, addr string) (net.Conn, error) {
+			conn, err := dialFunc(ctx, network, addr)
+			if err != nil {
+				return nil, err
+			}
+			tlsCfg := baseTLSCfg.Clone()
+			tlsCfg.ServerName = util.UpstreamServerNameFromContext(ctx)
+			if tlsCfg.ServerName == "" {
+				tlsCfg.ServerName, _, _ = net.SplitHostPort(addr)
+			}
+			return tls.Client(conn, tlsCfg), nil
+		}
+	}
```

Or am I missing something?

interceptor/config/serving.go
linkvt: Thanks I missed that comment, quoting parts of it as I'm missing the "why":

> it has a clear problem statement (kube-proxy lag)

To be honest this doesn't really explain why we should use the ClusterIP for warm requests and thus build a solution that is split between cold and warm requests? A single implementation just using Endpoint IPs seems simpler.
We also solve the issue, that existing pooled connections in the http.Transport are not targeting new pods. If we always connect to ClusterIP the TCP connection is kept open to the existing pods, new pods are only reached during new connections which is currently a problem.
knative also does not differentiate between cold and warm requests AFAIK.

> limits blast radius

I'm not sure what is meant here with the blast radius, but two paths (=different cold and warm paths) increase complexity compared to just one.
If this direct pod routing is broken, users would disable it completely anyways, changing it from "always" to "cold-start-only" or such would still give them broken cold start requests, right?

Let me know if I misunderstood @Fedosin 

I think we plan to enable it by default in the near future anyways, adding another step between us now and this goal is IMO not worth it which is why I see this as an on and off feature.

AtharvaPakade: > knative also does not differentiate between cold and warm requests AFAIK

True, it only differentiates based on mesh configuration modes, which can affect direct-to-pod reachability. 

I'll convert this flag to a boolean.

pkg/k8s/ready_endpoints_cache.go
```diff
@@ -14,14 +15,33 @@ import (
...
+}
+
+// serviceState is an immutable snapshot of a service's ready pod IPs.
+// Replaced atomically on every EndpointSlice event; readers hold a pointer
+// to the old snapshot safely without locks.
+type serviceState struct {
+	ready bool // true if the Service's ClusterIP is ready to serve traffic
+	// candidates is family → portName → ready (ip,port) pairs from the same slice.
+	// Family preference is supplied per-request via context, not stored here.
+	candidates map[discov1.AddressType]map[string][]endpoint
```
linkvt: Is there a reason the address type was added? Right now I don't see the benefit and would assume that it is simpler to always choose a (random) IP of all available ones.
I don't think we need to prefer a specific IP address type?

pkg/k8s/ready_endpoints_cache.go
```diff
@@ -3,9 +3,10 @@ package k8s
 import (
 	"context"
 	"fmt"
-	"slices"
+	"math/rand/v2"
```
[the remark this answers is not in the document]
linkvt: /ar no strong randomness needed here.

interceptor/config/serving.go
```diff
@@ -65,6 +65,12 @@ type Serving struct {
 	// server continues serving in-flight and new requests, giving Kubernetes
 	// time to propagate endpoint removal.
 	ShutdownDelay time.Duration `env:"KEDA_HTTP_SHUTDOWN_DELAY" envDefault:"5s"`
+	// DirectPodRouting routes all requests directly to a ready pod IP instead
+	// of through the Service ClusterIP. This bypasses kube-proxy and anything
+	// else operating at the Service layer (Service-level NetworkPolicy,
+	// session affinity, topology-aware routing). The address family is derived
+	// per-Service from Service.Spec.IPFamilies.
+	DirectPodRouting bool `env:"KEDA_HTTP_DIRECT_POD_ROUTING" envDefault:"false"`
```
linkvt, line 71: I propose to default this to true, this is also the knative behavior and direct pod routing is IMO very useful.

interceptor/proxy.go
```diff
@@ -104,3 +125,32 @@ func BuildProxyHandler(cfg *ProxyHandlerConfig) http.Handler {
...
+		if serverName == "" {
+			if tlsCfg.InsecureSkipVerify {
+				return nil, fmt.Errorf("upstream server name missing from context for TLS dial to %s (cert verification is disabled — refusing to dial without explicit SNI)", addr)
+			}
+			serverName, _, _ = net.SplitHostPort(addr)
+			util.LoggerFromContext(ctx).V(1).Info(
+				"upstream server name missing from context; falling back to dial address",
+				"addr", addr, "fallbackServerName", serverName,
+			)
+		}
```
linkvt, line 131: this feels a bit complex, can it actually happen that servername is empty?

linkvt (commented):
Left a few comments and will continue to review later, thanks for the update so far!

interceptor/config/serving.go
AtharvaPakade: >This bypasses kube-proxy and anything operating at the Service layer (Service-level NetworkPolicy, session affinity, topology-aware routing)

Because of this implication i made it false. Let me know if u still want to make is true.

pkg/k8s/ready_endpoints_cache.go
AtharvaPakade: So in case of dual-stack k8s setup, we will get 1 or 2 groups of endpoint slices 1 for ipv4 and 1 for ipv6 (this will be dictated by `.spec.ipFamilyPolicy` field on svc). If we get 2 groups then the latest group of endpoint slice might overwrite the pod's ip and port, so to avoid this i though, adding prefered group will be a good idea.

But now i realised that it is perfect eg for YAGNI. `candidates map[string][]endpoint` this will even be enough.

interceptor/proxy.go
AtharvaPakade: No it won't happen with current middleware chains. Added this condition because any middleware chain can polute the context and change servername . It will be good guard to have. i will refactor it and drop the conditions around `InsecureSkipVerify` and make simpler.

Fedosin (approved):
@AtharvaPakade Thanks for sticking with this through all the review rounds. The final result is solid. Great work!

CHANGELOG.md
```diff
@@ -34,10 +34,13 @@ This changelog keeps track of work items that have been completed and are ready
...
 
 ### Improvements
 
-- **General**: TODO ([#TODO](https://github.com/kedacore/http-add-on/issues/TODO))
+- **Interceptor**: The forwarding transport sets TLS `ServerName` per-dial via `DialTLSContext`, using the original service hostname captured in context, so SNI stays correct when the upstream URL is rewritten to a pod IP. ([#1473](https://github.com/kedacore/http-add-on/issues/1473))
+- **Interceptor**: TLS server name is captured in context by the routing middleware before any URL rewrites, so downstream transports always use the original service hostname for SNI. The routing middleware also resolves the upstream named port into context to drive direct-pod target selection. ([#1473](https://github.com/kedacore/http-add-on/issues/1473))
+- **Interceptor**: `ReadyEndpointsCache` now tracks full `(ip, port)` pairs per named port from EndpointSlices, enabling direct-pod routing (replaces the previous bool-only ready state). ([#1473](https://github.com/kedacore/http-add-on/issues/1473))
```
linkvt, line 43: The changelog is a bit too verbose, 4 entries for a single changes that leak implementation details. The top one is IMO good, the improvements are IMO not helpful.

interceptor/middleware/endpoint_resolver_test.go
```diff
@@ -133,6 +133,105 @@ func TestEndpointResolver_Fallback(t *testing.T) {
 	}
 }
 
+// TestEndpointResolver_DirectPodRouting_DoesNotOverwriteFallback verifies that
```
linkvt, line 136: It seems very high to add 583 lines for 21 added lines in the endpoint_resolver itself. There seems to be duplication (like `_HTTPSUpstream` and `_UsedOnColdStart_HTTPS`) or the HTTP+HTTPS tests that could probably be using table tests for each test case instead of full duplication?

linkvt (commented):
Thank you looks really good now, only two simple things I think we should improve (Changelog and the huge test suite).

Fedosin (approved):
I don't think we should hold this patch any longer. It is in a decent state, and we can address all remaining questions later in separate PRs.

@linkvt is on vacation, but he previously confirmed that he has no objections to merging this PR.

Thanks, @AtharvaPakade, for this work!

PR #1303, opened 2025-05-30
feat: Add custom placeholder responses for scale-from-zero scenarios

malpou:
Allows `HTTPScaledObjects` to serve configurable content (HTML, JSON, XML, plain text, etc.) while workloads scale up from zero, improving user experience during cold starts.

## Description

This PR implements customizable placeholder responses that are displayed to users while applications are scaling up from zero. This significantly improves the user experience during cold starts by providing immediate feedback instead of connection timeouts or errors.

The implementation is **content-agnostic**, allowing users to serve any type of response (HTML pages, JSON API responses, XML, plain text, etc.) without automatic modifications or assumptions about the content format.

## Key features:

- Adds `placeholderConfig` section to `HTTPScaledObject` CRD for configuring placeholder responses
- **Content-type agnostic**: Support for any response format (HTML, JSON, XML, plain text, etc.)
- Inline content or ConfigMap-based templates via environment variables
- Go template support with variables like `{{.ServiceName}}`, `{{.Namespace}}`, and `{{.RefreshInterval}}`
- Custom `Content-Type` headers to match your API format
- Custom HTTP status codes (default: 200 OK)
- No automatic content injection - your content is served exactly as provided
- Configurable via global defaults (ConfigMap) or per-HTTPScaledObject settings

## Implementation details:

- New placeholder handler in the interceptor that serves configured content when backends are unavailable
- Template variable substitution works with any content format
- Graceful fallback to upstream requests once the service is ready
- Error handling for endpoint cache failures returns 503 Service Unavailable
- Full backward compatibility - disabled by default

## Use cases:

- **HTML pages**: Show a "Service starting..." page with auto-refresh
- **JSON APIs**: Return `{"status": "initializing", "message": "Service is starting..."}` responses
- **XML services**: Provide XML-formatted status responses
- **Plain text**: Simple text-based status messages

## Checklist

- [x] Commits are signed with Developer Certificate of Origin (DCO)
- [x] Changelog has been updated and is aligned with our [changelog requirements](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#Changelog)
- [x] Any necessary documentation is added, such as:
  - [`README.md`](/README.md)
  - [The `docs/` directory](./docs)
  - [The docs repo](https://github.com/kedacore/keda-docs)

PR #1592, opened 2026-04-20
feat(interceptor): add SSE streaming callback for cold-start scaling

guimou:
## Summary

When serving LLM inference workloads behind KEDA HTTP Add-on, scaling from zero can take a significant amount of time (30s–several minutes) while the model loads into GPU memory. During this cold start, streaming clients (e.g. chat UIs, CLI tools using the OpenAI-compatible API) hold an open connection with no feedback, leading to:

- **User confusion**: no indication that the request is being processed.
- **Premature timeouts**: load balancers, proxies, or clients may drop the connection before the backend is ready.
- **Poor developer experience**: no way to distinguish "waiting for scale-up" from "request lost."

This PR adds an optional **SSE streaming callback** to the interceptor's cold-start path. When configured and the incoming request is a streaming request (`"stream": true` in the JSON body), the interceptor sends OpenAI-compatible `chat.completion.chunk` SSE keepalive events to the client while waiting for the backend to become ready. Once the backend is available, it inserts a visual separator and proxies the real upstream response through the same SSE stream.

### Configuration

- **HTTPScaledObject (v1alpha1)**: new `coldStartStreamingCallback` field with `message`, `intervalSeconds`, and `keepaliveMessage`.
- **InterceptorRoute (v1beta1)**: new `streamingCallback` field inside `coldStart` with `message`, `interval`, and `keepaliveMessage`.
- The `coldStart` validation is relaxed from requiring `fallback` to requiring at least one of `fallback` or `streamingCallback`.

### Example output (CLI client)

```
Model is waking up, hold on.........

---
Hello! How can I help you today?
```

### Changes

- **Interceptor middleware** (`interceptor/middleware/endpoint_resolver.go`): new `serveStreamingCallback` method, `isStreamingRequest` body parser, `writeSSEEvent` SSE formatter, and `headerSuppressingWriter` to avoid duplicate `WriteHeader` from the upstream proxy.
- **CRD types**: `ColdStartStreamingCallbackSpec` (v1alpha1), `StreamingCallbackSpec` (v1beta1), plus generated DeepCopy methods.
- **CRD manifests**: updated via `controller-gen`.
- **Routing table** (`pkg/routing/table.go`): converts v1alpha1 `coldStartStreamingCallback` to v1beta1 `streamingCallback` during `refreshMemory`.
- **Tests**: unit tests covering cold-start streaming, non-streaming requests, timeout, and unconfigured paths.

## Test plan

- [x] `make test` passes (unit tests including new streaming callback tests)
- [x] `make lint` passes
- [x] Deploy to a cluster with an LLM workload scaled to zero, send a streaming chat completion request, and verify keepalive SSE events are received until the backend is ready
- [x] Verify non-streaming requests still follow the normal cold-start/fallback path
- [x] Verify that when `streamingCallback` is not configured, behavior is unchanged
- [x] Verify the separator and model response appear on separate lines in CLI clients


linkvt:
Hi @guimou , thanks for the PR!
This is a very specific request with a very specific implementation, we have to consider how that fits into the HTTP Add-on.
There is a related topic of placeholder responses that I started working on a few days ago based on #1303 , I will have to check if that would fit in there. It might also be a usecase for a "placeholderService" where a user-defined service starts with the implementation. I don't have enough time to dig into that topic this week and am on vacation next week though, just a headsup.

guimou:
Hey @linkvt !
No worries, and no hurries. You're right, it's very specific, especially the implementation with a continuous SSE response. And it would definitely fit within #1303 scope. At least we can consider this like a PoC that proved to work, it's used in [this demo](https://github.com/rh-aiservices-bu/dynamic-model-autoscaling/blob/fixes/assets/docs/demo-http-addon-callback.md) already, something that will be part of Red Hat Summit's presentation.
I'll let you think about it, and happy to help in any ways!

PR #1593, opened 2026-04-21
feat: add orphan-scaledobject annotation for HTTPSO->IR migration

linkvt:
When migrating from HTTPScaledObject to InterceptorRoute, the auto-created ScaledObject is cascade-deleted with the HTTPScaledObject. GitOps tools cannot use --cascade=orphan because deletion is declarative.

The new httpscaledobject.keda.sh/orphan-scaledobject annotation tells the controller to remove the owner reference from the ScaledObject, allowing it to survive HTTPScaledObject deletion without a scaling gap.

More of a nice to have but quite simple to implement.

### Changes
- Add OrphanScaledObjectAnnotation and orphanScaledObject method
- Add AnnotationChangedPredicate to HTTPScaledObject watch predicates
- Add tests for orphan annotation with and without existing ScaledObject

### Checklist

### DIFF
--- a/operator/controllers/http/app.go
+++ b/operator/controllers/http/app.go
@@ -6,14 +6,20 @@ import (
 
 	"github.com/go-logr/logr"
 	kedav1alpha1 "github.com/kedacore/keda/v2/apis/keda/v1alpha1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	"github.com/kedacore/http-add-on/operator/apis/http/v1alpha1"
 	"github.com/kedacore/http-add-on/operator/controllers/http/config"
 )
 
-var SkipScaledObjectCreationAnnotation = "httpscaledobject.keda.sh/skip-scaledobject-creation"
+const (
+	annotationEnabled                  = "true"
+	SkipScaledObjectCreationAnnotation = "httpscaledobject.keda.sh/skip-scaledobject-creation"
+	OrphanScaledObjectAnnotation       = "httpscaledobject.keda.sh/orphan-scaledobject"
+)
 
 func (r *HTTPScaledObjectReconciler) createOrUpdateApplicationResources(
 	ctx context.Context,
@@ -32,11 +38,18 @@ func (r *HTTPScaledObjectReconciler) createOrUpdateApplicationResources(
 		httpso.Namespace,
 	)
 
+	if httpso.Annotations[OrphanScaledObjectAnnotation] == annotationEnabled {
+		logger.Info(
+			"Orphaning ScaledObject with annotation 'httpscaledobject.keda.sh/orphan-scaledobject'=true",
+			"HTTPScaledObject", httpso.Name)
+		return r.orphanScaledObject(ctx, cl, logger, httpso)
+	}
+
 	// We want to integrate http scaler with other
 	// scalers. when "httpscaledobject.keda.sh/skip-scaledobject-creation" is set to true,
 	// reconciler will skip the KEDA core ScaledObjects creation or delete scaledObject if it already exists.
 	// you can then create your own SO, and add http scaler as one of your triggers.
-	if httpso.Annotations[SkipScaledObjectCreationAnnotation] == "true" {
+	if httpso.Annotations[SkipScaledObjectCreationAnnotation] == annotationEnabled {
 		logger.Info(
 			"Skip scaled objects creation with flag 'httpscaledobject.keda.sh/skip-scaledobject-creation'=true",
 			"HTTPScaledObject", httpso.Name)
@@ -90,6 +103,45 @@ func (r *HTTPScaledObjectReconciler) deleteScaledObject(
 			"ScaledObject", &fetchedSO.Name)
 	}
 
+	return nil
+}
+
+func (r *HTTPScaledObjectReconciler) orphanScaledObject(
+	ctx context.Context,
+	cl client.Client,
+	logger logr.Logger,
+	httpso *v1alpha1.HTTPScaledObject,
+) error {
+	var fetchedSO kedav1alpha1.ScaledObject
+
+	objectKey := types.NamespacedName{
+		Namespace: httpso.Namespace,
+		Name:      httpso.Name,
+	}
+
+	if err := cl.Get(ctx, objectKey, &fetchedSO); err != nil {
+		if k8serrors.IsNotFound(err) {
+			logger.Info("ScaledObject not found, nothing to orphan",
+				"HTTPScaledObject", httpso.Name)
+			return nil
+		}
+		return err
+	}
+
+	if !isOwnerReferenceMatch(&fetchedSO, httpso) {
+		return nil
+	}
+
+	if err := controllerutil.RemoveOwnerReference(httpso, &fetchedSO, r.Scheme); err != nil {
+		return err
+	}
+
+	if err := cl.Update(ctx, &fetchedSO); err != nil {
+		return err
+	}
+
+	logger.Info("Orphaned ScaledObject by removing owner reference",
+		"ScaledObject", fetchedSO.Name)
 	return nil
 }
 
--- a/operator/controllers/http/httpscaledobject_controller.go
+++ b/operator/controllers/http/httpscaledobject_controller.go
@@ -162,6 +162,10 @@ func (r *HTTPScaledObjectReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&httpv1alpha1.HTTPScaledObject{}, builder.WithPredicates(
 			predicate.Or(
+				util.AnnotationKeyChangedPredicate{Keys: []string{
+					OrphanScaledObjectAnnotation,
+					SkipScaledObjectCreationAnnotation,
+				}},
 				predicate.GenerationChangedPredicate{},
 				util.HTTPScaledObjectReadyConditionPredicate{},
 			),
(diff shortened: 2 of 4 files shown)


operator/controllers/http/httpscaledobject_controller.go
```diff
@@ -162,6 +162,7 @@ func (r *HTTPScaledObjectReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&httpv1alpha1.HTTPScaledObject{}, builder.WithPredicates(
 			predicate.Or(
+				predicate.AnnotationChangedPredicate{},
 				predicate.GenerationChangedPredicate{},
 				util.HTTPScaledObjectReadyConditionPredicate{},
```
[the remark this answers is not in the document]
linkvt: not worth it

operator/controllers/http/httpscaledobject_controller_test.go
```diff
@@ -8,6 +8,63 @@ import (
...
+
+	err := reconciler.createOrUpdateApplicationResources(
+		testInfra.ctx, testInfra.logger, testInfra.cl,
+		config.Base{}, config.ExternalScaler{}, &testInfra.httpso,
+	)
+	r.NoError(err)
+
+	// ScaledObject must not have been created
+	_, err = getSO(testInfra.ctx, testInfra.cl, testInfra.httpso)
+	r.Error(err)
```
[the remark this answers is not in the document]
linkvt: This annotation should only be used during the migration -> invalid

operator/controllers/http/app.go
```diff
@@ -32,6 +37,13 @@ func (r *HTTPScaledObjectReconciler) createOrUpdateApplicationResources(
 		httpso.Namespace,
 	)
 
+	if httpso.Annotations[OrphanScaledObjectAnnotation] == "true" { //nolint:goconst
+		logger.Info(
+			"Orphaning ScaledObject with annotation 'httpscaledobject.keda.sh/orphan-scaledobject'=true",
+			"HTTPScaledObject", httpso.Name)
```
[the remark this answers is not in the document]
linkvt: This annotation should only be used during the migration -> invalid

operator/controllers/http/app.go
```diff
@@ -93,6 +106,45 @@ func (r *HTTPScaledObjectReconciler) deleteScaledObject(
...
+	if err := cl.Get(ctx, objectKey, &fetchedSO); err != nil {
+		if k8serrors.IsNotFound(err) {
+			logger.Info("ScaledObject not found, nothing to orphan",
+				"HTTPScaledObject", httpso.Name)
+			return nil
+		}
+		return err
+	}
+
+	prevLen := len(fetchedSO.OwnerReferences)
```
Fedosin: **Bug:** `controllerutil.RemoveOwnerReference` (controller-runtime v0.22.5) returns an error both when OwnerReferences is empty and when the specific owner ref is not found:

```go
if length < 1 {
    return fmt.Errorf("%T does not have any owner references", object)
}
// ...
if index == -1 {
    return fmt.Errorf("%T does not have an owner reference for %T", object, owner)
}
```

This means:
- The `prevLen` comparison below is dead code — it can never be reached.
- On the **second reconcile** with the orphan annotation set, the owner ref is already gone, so `RemoveOwnerReference` will error out and the reconciler will report a failure.

Suggestion: guard with `isOwnerReferenceMatch` (which already exists in this file) before calling `RemoveOwnerReference`:

```go
if !isOwnerReferenceMatch(&fetchedSO, httpso) {
    logger.Info("ScaledObject not owned by this HTTPScaledObject, nothing to orphan",
        "ScaledObject", fetchedSO.Name)
    return nil
}
```

This also lets you drop the `prevLen` check entirely.

operator/controllers/http/app.go
```diff
@@ -32,11 +38,18 @@ func (r *HTTPScaledObjectReconciler) createOrUpdateApplicationResources(
 		httpso.Namespace,
 	)
 
+	if httpso.Annotations[OrphanScaledObjectAnnotation] == annotationEnabled {
```
Fedosin, line 41: When the orphan annotation is set, this returns early without calling `createOrUpdateScaledObject`. Back in the controller, a nil error unconditionally sets the status to Ready with message `"ScaledObject created and configured"` — even if no ScaledObject was ever created (e.g. orphan annotation on a brand-new HTTPSO).

I understand the annotation is only intended for migration, but the controller should still be robust against misuse. Consider either:
- Verifying the ScaledObject actually exists before orphaning, or
- Setting a distinct status message/reason (e.g. `"ScaledObject orphaned"`) so users can tell what happened.

operator/controllers/http/httpscaledobject_controller.go
```diff
@@ -162,6 +162,7 @@ func (r *HTTPScaledObjectReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&httpv1alpha1.HTTPScaledObject{}, builder.WithPredicates(
 			predicate.Or(
+				predicate.AnnotationChangedPredicate{},
```
Fedosin: `AnnotationChangedPredicate` fires on **any** annotation change, not just the orphan annotation. In a real cluster, tools like ArgoCD, Helm, kubectl, etc. frequently add/modify annotations, and each change will now trigger a full reconcile.

The existing `skip-scaledobject-creation` annotation doesn't need this predicate because changing it bumps `.metadata.generation`. Consider either:
- A custom predicate that only fires on changes to the specific orphan annotation key, or
- Moving the orphan flag into `.spec` so `GenerationChangedPredicate` handles it naturally.

operator/controllers/http/app.go
```diff
@@ -32,11 +38,18 @@ func (r *HTTPScaledObjectReconciler) createOrUpdateApplicationResources(
...
+		logger.Info(
+			"Orphaning ScaledObject with annotation 'httpscaledobject.keda.sh/orphan-scaledobject'=true",
+			"HTTPScaledObject", httpso.Name)
+		return r.orphanScaledObject(ctx, cl, logger, httpso)
+	}
+
```
Fedosin, line 47: Nit: if both `orphan-scaledobject` and `skip-scaledobject-creation` annotations are set, the orphan path wins silently because it's checked first. These encode conflicting intent (orphan = "keep the SO alive without owner" vs skip = "delete the SO if it exists"). Consider detecting the conflict and returning an error, or at minimum documenting the precedence.

operator/controllers/http/httpscaledobject_controller_test.go
```diff
@@ -8,6 +8,63 @@ import (
...
+	r.Len(so.OwnerReferences, 1)
+
+	// Add orphan annotation and reconcile again
+	testInfra.httpso.Annotations[OrphanScaledObjectAnnotation] = "true"
+	err = reconciler.createOrUpdateApplicationResources(
+		testInfra.ctx, testInfra.logger, testInfra.cl,
+		config.Base{}, config.ExternalScaler{}, &testInfra.httpso,
+	)
+	r.NoError(err)
+
```
Fedosin, line 39: This test verifies the happy path but doesn't call `createOrUpdateApplicationResources` a third time to confirm idempotency. Given that `RemoveOwnerReference` errors when the ref is already removed (see my comment on `app.go`), a second reconcile with the orphan annotation set would actually fail. Adding an idempotency assertion here would catch that bug.

Fedosin (changes requested):
Thanks for the PR — the feature solves a real migration pain point and the implementation is concise. I found a few issues though, ranging from a bug to design concerns. See inline comments.

**Summary:**

1. **Bug: `RemoveOwnerReference` errors on subsequent reconciles.** controller-runtime v0.22.5's `RemoveOwnerReference` returns an error both when no owner references exist and when the specific ref is not found. This means (a) the `prevLen` comparison is dead code, and (b) the second reconcile with the orphan annotation set will fail because the owner ref was already removed. This makes the feature unreliable since controllers reconcile repeatedly.

2. **Early return skips ScaledObject creation — misleading Ready status.** When the orphan annotation is set, `createOrUpdateApplicationResources` returns early. The controller then unconditionally sets status to Ready with "ScaledObject created and configured", even if no ScaledObject exists.

3. **`AnnotationChangedPredicate` is too broad.** Any annotation change (from ArgoCD, kubectl, Helm, etc.) will now trigger a full reconcile. Consider a targeted predicate or moving the orphan flag to `.spec`.

4. **No conflict detection with `skip-scaledobject-creation`.** If both annotations are set, the orphan path wins silently. These encode conflicting intent.

5. **Test doesn't verify idempotency.** Given bug #1, a second reconcile after orphaning would fail, but the test doesn't cover this.

operator/controllers/http/app.go
linkvt: This annotation is only there for migrating to the IR+SO approach, so IMO we can ignore the special case of someone orphaning and then adding the skip annotation.

operator/controllers/http/httpscaledobject_controller_test.go
linkvt: Good idea, added it to the bottom of the test.

operator/controllers/http/app.go
linkvt: Good catch, fixed 👍 Was wondering why isOwnerReferenceMatch is there when there is an alternative in the library... Guess taht explains

operator/controllers/http/httpscaledobject_controller.go
linkvt: Annotation changes dont bump generations, I tried it first but it failed without the annotationchanged predicate. This means the skip-SO annotation was also not working correctly before.
I added a custom predicate for our annotations to prevent unnecessary reconciles.

operator/controllers/http/app.go
linkvt: Same behavior as with the preexisting skip-SO creation annotation, IMO good to skip as this is only used for the migration followed by a deletion of the HTTPSO, status was and still is not really relevant unfortunately, doesnt say anything about HTTPSO/IRs being loaded in all interceptors.

Fedosin (approved):

PR #1599, opened 2026-04-23
feat: add native Go fuzz tests for untrusted-input surfaces

Fedosin:
Add 6 native Go fuzz tests (`testing.F`) targeting the main untrusted-input
surfaces across the interceptor, scaler, and routing packages:

**Interceptor** (`interceptor/fuzz_test.go`):
- `FuzzParseTLSVersion` -- TLS version string parsing (crash detection)
- `FuzzParseCipherSuites` -- cipher suite name parsing (crash detection)
- `FuzzParseCurvePreferences` -- curve preference parsing (crash detection)
- `FuzzProxyHandler` -- full HTTP proxy handler chain with fuzzed method/host/path/headers, with semantic assertions (known host → 200, unknown host → error)

**Routing** (`pkg/routing/fuzz_test.go`):
- `FuzzTableMemoryRoute` -- radix-tree route matching with arbitrary hostnames, paths, and headers (crash detection + route sanity check)

**Scaler** (`scaler/fuzz_test.go`):
- `FuzzEscapeString` -- metric name escaping produces only safe characters

Each target focuses on crash detection for code that processes untrusted input,
with stronger semantic assertions where applicable.

Also adds:
- `make fuzz` Makefile target with configurable `FUZZ_TIME` (default 30s)
- `.github/workflows/fuzz.yaml` CI workflow running on every PR as a smoke check
- Automatic fuzz corpus upload as CI artifact on failure for regression seeding

### Checklist

  - [`README.md`](../README.md)

### DIFF
--- a/interceptor/fuzz_test.go
+++ b/interceptor/fuzz_test.go
@@ -1,0 +1,184 @@
+package main
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/go-logr/logr"
+	discov1 "k8s.io/api/discovery/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	"github.com/kedacore/http-add-on/interceptor/config"
+	"github.com/kedacore/http-add-on/interceptor/metrics"
+	httpv1beta1 "github.com/kedacore/http-add-on/operator/apis/http/v1beta1"
+	kedacache "github.com/kedacore/http-add-on/pkg/cache"
+	"github.com/kedacore/http-add-on/pkg/k8s"
+	"github.com/kedacore/http-add-on/pkg/queue"
+	routingtest "github.com/kedacore/http-add-on/pkg/routing/test"
+)
+
+// FuzzParseTLSVersion exercises the TLS version parser with arbitrary strings.
+// The function parses user-supplied configuration, so it must never panic.
+func FuzzParseTLSVersion(f *testing.F) {
+	f.Add("1.2")
+	f.Add("1.3")
+	f.Add("")
+	f.Add("1.0")
+	f.Add("1.1")
+	f.Add("TLSv1.2")
+	f.Add("ssl3")
+
+	f.Fuzz(func(t *testing.T, v string) {
+		_, _ = parseTLSVersion(v)
+	})
+}
+
+// FuzzParseCipherSuites exercises cipher-suite name parsing with arbitrary
+// strings. The function parses user-supplied configuration, so it must never
+// panic.
+func FuzzParseCipherSuites(f *testing.F) {
+	f.Add("TLS_AES_128_GCM_SHA256")
+	f.Add("TLS_AES_128_GCM_SHA256,TLS_AES_256_GCM_SHA384")
+	f.Add("")
+	f.Add(",,,")
+	f.Add("UNKNOWN_SUITE")
+	f.Add("TLS_AES_128_GCM_SHA256, TLS_AES_256_GCM_SHA384")
+
+	f.Fuzz(func(t *testing.T, s string) {
+		_, _ = parseCipherSuites(s)
+	})
+}
+
+// FuzzParseCurvePreferences exercises curve preference parsing with arbitrary
+// strings. The function parses user-supplied configuration, so it must never
+// panic.
+func FuzzParseCurvePreferences(f *testing.F) {
+	f.Add("CurveP256")
+	f.Add("X25519")
+	f.Add("P-256,P-384")
+	f.Add("")
+	f.Add(",,,")
+	f.Add("UNKNOWN_CURVE")
+	f.Add("CurveP256, X25519, CurveP384")
+
+	f.Fuzz(func(t *testing.T, s string) {
+		_, _ = parseCurvePreferences(s)
+	})
+}
+
+// FuzzProxyHandler sends fuzzed HTTP requests through the full proxy handler
+// chain. This exercises routing, queue interaction, and upstream proxying with
+// arbitrary method/host/path/header combinations. The primary goal is crash
+// detection (no panics, deadlocks, or goroutine leaks). Additionally, we
+// verify that requests to the known host are proxied successfully and requests
+// to unknown hosts receive an error response.
+func FuzzProxyHandler(f *testing.F) {
+	f.Add("GET", "fuzz.example.com", "/api/v1", "X-Custom", "value")
+	f.Add("POST", "fuzz.example.com", "/", "", "")
+	f.Add("DELETE", "unknown.host", "/path", "", "")
+	f.Add("GET", "", "", "", "")
+	f.Add("PATCH", "fuzz.example.com", "/api/../secret", "Host", "evil.com")
+	f.Add("GET", "[::1]:8080", "/path", "X-Forwarded-For", "127.0.0.1")
+
+	const fuzzHost = "fuzz.example.com"
+	const fuzzServiceKey = "fuzz-ns/fuzz-service"
+
+	backend := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusOK)
+	}))
+	f.Cleanup(backend.Close)
+
+	fakeQueue := queue.NewFakeCounterBuffered()
+	done := make(chan struct{})
+	go func() {
+		for {
+			select {
+			case <-fakeQueue.ResizedCh:
+			case <-done:
+				return
+			}
+		}
+	}()
+	f.Cleanup(func() { close(done) })
+
+	readyCache := k8s.NewReadyEndpointsCache(logr.Discard())
+	readyCache.Update(fuzzServiceKey, []*discov1.EndpointSlice{{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "fuzz-slice",
+			Namespace: "fuzz-ns",
+			Labels:    map[string]string{discov1.LabelServiceName: "fuzz-service"},
+		},
+		Endpoints: []discov1.Endpoint{{Addresses: []string{"1.2.3.4"}}},
+	}})
+
+	routingTable := routingtest.NewTable()
+	routingTable.Memory[fuzzHost] = &httpv1beta1.InterceptorRoute{
+		ObjectMeta: metav1.ObjectMeta{Name: "fuzz-ir", Namespace: "fuzz-ns"},
+		Spec: httpv1beta1.InterceptorRouteSpec{
+			Target: httpv1beta1.TargetRef{Service: "fuzz-service", Port: 80},
+		},
+	}
+
+	handler := BuildProxyHandler(&ProxyHandlerConfig{
+		Logger:       logr.Discard(),
+		Queue:        fakeQueue,
+		ReadyCache:   readyCache,
+		RoutingTable: routingTable,
+		Reader:       fake.NewClientBuilder().WithScheme(kedacache.NewScheme()).Build(),
+		Timeouts: config.Timeouts{
+			Connect:        500 * time.Millisecond,
+			Readiness:      1 * time.Second,
+			Request:        5 * time.Second,
+			ResponseHeader: 2 * time.Second,
+		},
+		Serving:             config.Serving{},
+		Tracing:             config.Tracing{},
+		Instruments:         metrics.NewNoopInstruments(),
+		dialAddressOverride: backend.Listener.Addr().String(),
+	})
+
+	validMethods := []string{"GET", "POST", "PUT", "DELETE", "PATCH", "HEAD", "OPTIONS", "TRACE", "CONNECT"}
+
+	f.Fuzz(func(t *testing.T, method, host, path, headerKey, headerVal string) {
+		if len(method) == 0 {
+			method = "GET"
+		} else {
+			method = validMethods[int(method[0])%len(validMethods)]
+		}
+
+		if path == "" || path[0] != '/' {
+			path = "/" + path
+		}
+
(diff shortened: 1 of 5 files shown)


interceptor/fuzz_test.go
```diff
@@ -0,0 +1,155 @@
...
+	const fuzzHost = "fuzz.example.com"
+	const fuzzIRKey = "fuzz-ns/fuzz-ir"
+	const fuzzServiceKey = "fuzz-ns/fuzz-service"
+
+	backend := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusOK)
+	}))
+	f.Cleanup(backend.Close)
+
+	fakeQueue := queue.NewFakeCounterBuffered()
```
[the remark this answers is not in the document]
Fedosin: Good catch — added a drain goroutine with `f.Cleanup`. This also improved FuzzProxyHandler throughput from ~968 to ~9000 execs/sec.

pkg/routing/fuzz_test.go
```diff
@@ -0,0 +1,140 @@
...
+	f.Add("k8s.io", "//abc/def//")
+	f.Add("[::1]:8080", "/path")
+	f.Add("host\x00name", "/path\x00with\x00nulls")
+	f.Add("a.b.c.d.e.f.g.h.i.j.example.com", "/"+string(make([]byte, 1024)))
+
+	f.Fuzz(func(t *testing.T, hostname, path string) {
+		key := NewKey(hostname, path)
+		if key == nil {
+			t.Error("NewKey returned nil")
+		}
```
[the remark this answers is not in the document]
Fedosin: Agreed — removed the dead nil check and replaced it with an assertion that the `/` separator is at the expected position after the hostname.

pkg/routing/fuzz_test.go
```diff
@@ -0,0 +1,140 @@
...
+	f.Add("example.com:8080")
+	f.Add("[::1]:8080")
+	f.Add("")
+	f.Add(":")
+	f.Add(":8080")
+	f.Add("[::1]")
+	f.Add("host:port:extra")
+
+	f.Fuzz(func(t *testing.T, host string) {
+		_ = stripPort(host)
```
JorTurFer: I'm not sure what we test here. Maybe is because I'm not fully in with fuzzy tests, but as stripPort is safe from panics and we don't check the return, which is the goal? Am I missing somthing?

JorTurFer:
Does it make sense to expose the fuzzy file in case of failures in include those cases in next iterations?

scaler/fuzz_test.go
```diff
@@ -0,0 +1,64 @@
...
+	f.Add("my-app")
+	f.Add("")
+	f.Add("name with spaces")
+	f.Add("unicode-\u00e9")
+	f.Add("emoji-\U0001f600")
+	f.Add("a/b")
+
+	f.Fuzz(func(t *testing.T, irName string) {
+		result := RateMetricName(irName)
+		if !strings.HasPrefix(result, "http_") || !strings.HasSuffix(result, "_rate") {
```
linkvt: Same as above

scaler/fuzz_test.go
```diff
@@ -0,0 +1,64 @@
...
+	f.Add("name with spaces")
+	f.Add("unicode-\u00e9")
+	f.Add("emoji-\U0001f600")
+	f.Add("a/b")
+
+	f.Fuzz(func(t *testing.T, irName string) {
+		result := ConcurrencyMetricName(irName)
+		if !strings.HasPrefix(result, "http_") || !strings.HasSuffix(result, "_concurrency") {
+			t.Errorf("ConcurrencyMetricName(%q) = %q, unexpected format", irName, result)
+		}
```
linkvt: I don't think we actually test anything here besides whether fmt.Printf is able to put the `irName` inbetween `http_` and `_concurrency`?

scaler/fuzz_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+import (
+	"regexp"
+	"strings"
+	"testing"
+)
+
+var safePattern = regexp.MustCompile(`^[-.0-9A-Za-z_]*$`)
+
+func FuzzEscapeString(f *testing.F) {
```
linkvt, line 12: I'm not sure what we test here, I think we can be quite sure that calling the same function twice produces the same result. Also comparing it against the regex above just means we're testing go's regex implementation?

pkg/routing/fuzz_test.go
```diff
@@ -0,0 +1,140 @@
...
+	f.Add("")
+	f.Add("a.b.c.d.e.f.g.h.i.j.k")
+	f.Add(".....")
+	f.Add(".leading.dot")
+	f.Add("trailing.dot.")
+
+	f.Fuzz(func(t *testing.T, hostname string) {
+		result := wildcardHostnames(hostname)
+		for _, w := range result {
+			if len(w) < 2 || w[0] != '*' || w[1] != '.' {
```
linkvt: Could maybe be simplified to `strings.HasPrefix(w, "*.")` but we're effectively just testing that `"*."+remaining` works, right?

pkg/routing/fuzz_test.go
```diff
@@ -0,0 +1,140 @@
...
+	tm := fuzzTableMemory()
+
+	f.Fuzz(func(t *testing.T, hostname, path, headerKey, headerVal string) {
+		var headers http.Header
+		if headerKey != "" {
+			headers = http.Header{headerKey: {headerVal}}
+		}
+
+		ir := tm.Route(hostname, path, headers)
+		if ir != nil && !knownRouteNames[ir.Name] {
```
linkvt, line 89: I'm not sure what we test here as we will always pass the check  `knownRouteNames[ir.Name]` as the table contains just these routes. Isn't a unit test which we probably already have enough to test that the `Route()` doesn't return unwanted results? Or should this test check if we crash?

pkg/routing/fuzz_test.go
```diff
@@ -0,0 +1,140 @@
...
+	f.Add("", "")
+	f.Add("k8s.io", "//abc/def//")
+	f.Add("[::1]:8080", "/path")
+	f.Add("host\x00name", "/path\x00with\x00nulls")
+	f.Add("a.b.c.d.e.f.g.h.i.j.example.com", "/"+string(make([]byte, 1024)))
+
+	f.Fuzz(func(t *testing.T, hostname, path string) {
+		key := NewKey(hostname, path)
+		s := key.String()
+		if len(s) == 0 || s[len(hostname)] != '/' {
```
linkvt: Doesnt this just test that we always append a slash? Otherwise we would just fuzz `strings.Builder`?

pkg/queue/fuzz_test.go
```diff
@@ -0,0 +1,84 @@
...
+import (
+	"encoding/binary"
+	"encoding/json"
+	"math"
+	"reflect"
+	"testing"
+	"time"
+)
+
+func FuzzCountsJSON(f *testing.F) {
```
linkvt: Isn't this test just testing/fuzzing `json.Marshal` and `json.Unmarshal`?

interceptor/fuzz_test.go
```diff
@@ -0,0 +1,124 @@
...
+		req.RequestURI = path
+		if headerKey != "" {
+			req.Header.Set(headerKey, headerVal)
+		}
+
+		rec := httptest.NewRecorder()
+		handler.ServeHTTP(rec, req)
+
+		code := rec.Code
+		if code < 100 || code > 599 {
```
linkvt: Are there any cases where this check would pass? I am again unsure what we're fuzzing/testing here also considering the big setup of this test.

pkg/queue/fuzz_test.go
```diff
@@ -0,0 +1,84 @@
...
+
+		if lastTime.IsZero() {
+			return
+		}
+
+		avg := buckets.WindowAverage(lastTime)
+		if math.IsNaN(avg) || math.IsInf(avg, 0) {
+			t.Errorf("WindowAverage returned %v", avg)
+		}
+		if avg < 0 {
```
linkvt: This assertion seems a bit weak, I would guess if we want to test for negative values we could add a unit test as well?

linkvt (commented):
I'm a bit unsure about most of the fuzz tests as most of them are to my understanding not testing something where fuzzing is worth it? I'm definitely not knowledgable in the whole fuzzing topic though.
I would have expected fuzzing when we would ahve implemented custom complex parsing, conversions and (de-)serializations that are hard to check with unit tests and which could actually fail somehow, I don't think this is the case for most of hte fuzz tests though?

pkg/routing/fuzz_test.go
Fedosin: You're right — `FuzzStripPort` had no assertions and `stripPort` doesn't have realistic panic risk. Removed it in the latest push.

Fedosin:
> Does it make sense to expose the fuzzy file in case of failures in include those cases in next iterations?

Great idea! Added an `actions/upload-artifact` step to the CI workflow that uploads `testdata/fuzz/**` on failure with 30-day retention, so crash-triggering inputs are preserved and can be added as regression seeds in future iterations.

Fedosin:
@linkvt Thank you for the thorough review — you're right that most of the fuzz tests were testing trivial properties or standard library behavior rather than our own logic. I've reworked the PR based on your feedback:

**Removed** (tests that were just fuzzing stdlib or trivial string operations):
- `FuzzConcurrencyMetricName`, `FuzzRateMetricName` — just testing `fmt.Sprintf`
- `FuzzCountsJSON`, `FuzzRequestsBuckets` — just testing `json.Marshal`/`json.Unmarshal`; deleted `pkg/queue/fuzz_test.go` entirely
- `FuzzNewKey` — just testing `strings.Builder`
- `FuzzStripPort` — no assertions, no realistic panic risk
- `FuzzWildcardHostnames` — just testing `"*." + string` concatenation

**Kept and improved** (targets with real value):
- `FuzzEscapeString` — removed the pointless determinism check, kept the `safePattern` assertion which verifies a real contract (output must only contain metric-safe characters)
- `FuzzProxyHandler` — replaced the weak status-code range check with semantic assertions: known-host requests must return 200, unknown-host requests must not. Primary value is crash detection through the full handler chain.
- `FuzzTableMemoryRoute` — kept for crash detection on the radix tree with arbitrary input; added doc comment clarifying intent
- `FuzzParseTLSVersion`, `FuzzParseCipherSuites`, `FuzzParseCurvePreferences` — these parse user-supplied config strings, crash detection is appropriate here

Down from 10 targets to 6, each with a clear purpose documented in its godoc comment.

scaler/fuzz_test.go
Fedosin: Fair point on the determinism check — removed it. Kept the `safePattern` assertion though, as it verifies a real contract: `escapeString` must only produce metric-safe characters regardless of input. If someone refactors the function and accidentally lets through a `/` or a space, this catches it.

scaler/fuzz_test.go
Fedosin: You're right, this was just testing `fmt.Sprintf`. Removed.

scaler/fuzz_test.go
Fedosin: Agreed, removed along with `FuzzConcurrencyMetricName`.

pkg/routing/fuzz_test.go
Fedosin: Agreed — this was trivial string concatenation testing. Removed.

pkg/routing/fuzz_test.go
Fedosin: You're right that the assertion is weak on its own. However, the primary value here is crash detection — feeding arbitrary strings into the radix-tree routing engine to catch panics, deadlocks, or unexpected behavior. I've added a doc comment clarifying this intent and kept the secondary check as a sanity guard.

pkg/routing/fuzz_test.go
Fedosin: Agreed — just testing `strings.Builder`. Removed.

pkg/queue/fuzz_test.go
Fedosin: You're right, `Counts` has no custom serialization — this was just fuzzing the stdlib. Removed; deleted `pkg/queue/fuzz_test.go` entirely.

interceptor/fuzz_test.go
Fedosin: Good point about the weak assertion. Replaced the status-code range check with semantic assertions: requests to the known host (`fuzz.example.com`) must return 200, and requests to unknown hosts must not return 200. The primary value remains crash detection — exercising the full handler chain with arbitrary method/host/path/header combinations to catch panics, deadlocks, or goroutine leaks.

pkg/queue/fuzz_test.go
Fedosin: Agreed, the assertion was too weak. Removed `FuzzRequestsBuckets` along with the rest of `pkg/queue/fuzz_test.go`.

CHANGELOG.md
```diff
@@ -36,6 +36,7 @@ This changelog keeps track of work items that have been completed and are ready
 
 ### Improvements
 
+- **General**: Add native Go fuzz tests for untrusted-input surfaces (routing, TLS config parsing, proxy handler, metric escaping) ([#1599](https://github.com/kedacore/http-add-on/pull/1599))
```
linkvt: Just noticed this changelog entry, not sure if we want to keep it as it is no user facing change?

linkvt (approved):
Left another small comment on the PR, but LGTM to me in general.

I understand that these tests are more meant as a crash detection mechanism in most cases but don't really think that running them for every commit in every open PR is helpful and worth the CPU time (e.g. TLS version parsing probably won't change anytime soon) though, but that's a bit subjective I guess.
We could also consider running it like once a week or so on a schedule but don't want to block anything here.

Issue #1427, opened 2026-01-27
upgrade e2e tests


wozniakjan:
### Proposal

add upgrade scenario to suite of e2e tests - https://github.com/kedacore/http-add-on/tree/main/tests/checks

### Is this a feature you are interested in implementing yourself?

Yes

curiousjazz77:
Hi! I'm a student in the CodePath AI301 program and I'd like to work on this issue. I'll start by reproducing the problem locally this week. Any pointers on where to start in the codebase would be appreciated!

linkvt:
Hi @curiousjazz77, nice, you could take a look at the PR referenced above and its comments as that might help.
Please make sure to understand this issue and your approach, simply generating PRs with AI doesn't really help, thanks.

JoeArias1121:
Hi @linkvt , seeing as no one is really doing anything to work on the issue at the moment I would like to work on this issue. I will work on reproducing the issue this week and might ask questions.

PR #1610, opened 2026-04-24
test(e2e): add Helm-chart upgrade scenario

Sanil2108:
## Summary

Refs #1427. Adds an upgrade scenario to \`tests/checks/\` that installs a previous release of the \`kedacore/http-add-on\` Helm chart, verifies a basic HTTPScaledObject scales on load, runs \`helm upgrade\` to a target version, and then verifies the same HTTPScaledObject still reconciles and scales after the upgrade.

## How it runs

The test is **skipped unless both env vars are set**:

- \`HTTPADDON_UPGRADE_FROM_VERSION\` — the Helm chart version to install as baseline (e.g. \`0.10.0\`).
- \`HTTPADDON_UPGRADE_TO_VERSION\` — the Helm chart version to upgrade to (e.g. \`0.11.0\`).

So the addition is a no-op for the existing e2e suite. Running it requires a dedicated CI job (or a local invocation) that sets both variables before running \`go run -tags e2e ./tests/run-all.go\` with an \`E2E_TEST_REGEX\` narrowed to \`upgrade\`.

## Why skip-by-default

The main legacy e2e suite installs the add-on once at the start (\`TestDeployKEDAHttpAddOn\`) and assumes every check runs against that single, pinned release. The upgrade scenario intentionally swaps the cluster-wide release, so it can't safely share a run with other checks (runner uses \`concurrentTests=2\`). Guarding on the env vars keeps the existing suite's assumptions intact.

## What the test verifies

1. \`helm upgrade --install http-add-on kedacore/http-add-on --version $FROM\` succeeds.
2. A plain deployment + service + HTTPScaledObject in a fresh namespace reconciles, and \`curl\` traffic through the interceptor scales it from 0 → 1.
3. \`helm upgrade --install http-add-on kedacore/http-add-on --version $TO\` succeeds and the operator / interceptor / scaler deployments all come back ready.
4. The **pre-existing** HTTPScaledObject is still honored by the upgraded operator — the deployment stays up, load still scales correctly, and scale-in still works.

## Notes / open questions for the maintainer

- This tests chart-released version → chart-released version upgrades. Upgrading **to the local dev build** needs \`make deploy-e2e\` + ko + Kind image loading, which felt out of scope for an initial upgrade check. Happy to add that as a follow-up if you want CI to also cover "released → HEAD".
- Release-branch CI could set the env vars to exercise the last two minor versions, giving us a permanent safety net for upgrade breakage.
- Deployment names in \`waitForAddonReady\` (\`keda-add-ons-http-operator\`, \`keda-add-ons-http-interceptor\`, \`keda-add-ons-http-external-scaler\`) match the current chart. If those names change historically, I can switch to label-based readiness.

Refs #1427

### DIFF
--- a/test/e2e/upgrade/upgrade_test.go
+++ b/test/e2e/upgrade/upgrade_test.go
@@ -207,7 +207,7 @@ func TestUpgrade(t *testing.T) {
 
 func undeployPreExistingAddon(t *testing.T) {
 	t.Helper()
-	// Runs from tests/checks/upgrade_httpso/ — repo root is two levels up.
+	// Runs from test/e2e/upgrade/ — repo root is three levels up.
 	out, err := ExecuteCommandWithDir("make undeploy", "../../..")
 	if err != nil {
 		// Not fatal — `make undeploy` may return non-zero on a clean cluster.
--- a/tests/checks/upgrade_httpso/upgrade_test.go
+++ b/tests/checks/upgrade_httpso/upgrade_test.go
@@ -155,6 +155,20 @@ func TestUpgrade(t *testing.T) {
 
 	kc := GetKubernetesClient(t)
 
+	// The main runner (tests/run-all.go → tests/utils/setup_test.go) installs the
+	// add-on via `make deploy-e2e` (kustomize/ko), which creates the same
+	// keda-add-ons-http-* resources Helm wants to own. Tear those down first so
+	// the Helm release can take ownership cleanly. `make undeploy` is idempotent
+	// (it swallows errors if nothing is deployed) so this also handles the case
+	// where the test runs on a fresh cluster.
+	t.Log("--- tearing down any pre-existing add-on install ---")
+	undeployPreExistingAddon(t)
+	t.Cleanup(func() {
+		// Leave the cluster without the Helm release when we're done; a follow-up
+		// `make deploy-e2e` can reinstall via kustomize/ko.
+		_, _ = ExecuteCommand(fmt.Sprintf("helm uninstall %s --namespace %s --wait", releaseName, KEDANamespace))
+	})
+
 	t.Logf("--- installing http-add-on %s (baseline) ---", fromVersion)
 	installOrUpgradeAddon(t, fromVersion)
 	waitForAddonReady(t, kc)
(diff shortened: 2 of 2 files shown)


Sanil2108:
Thanks for the review! Addressed both points in the follow-up commit:

1. **Release-ownership conflict (L203):** Added a `make undeploy` call at the start to clear the kustomize/ko install before the first `helm upgrade --install`, plus a `helm uninstall` in `t.Cleanup` so the cluster is left without the Helm release at the end. A follow-up `make deploy-e2e` can restore the regular kustomize install if the dedicated job is running alongside anything else.

2. **Flaky post-upgrade replica assertion (L183):** The previous flow was relying on leftover pending load to keep the deployment at `maxReplicaCount` across the upgrade, which isn't safe given `scaledownPeriod: 10s`. Restructured: baseline scale-out → baseline scale-in → upgrade → fresh scale-out → fresh scale-in. The post-upgrade scale-out is now the real "HTTPScaledObject survives the upgrade" assertion.

test/e2e/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,268 @@
+//go:build e2e
```
linkvt, line 1: Hi @Sanil2108 , thanks for the PR but the tests/ folder contains only the last few tests of the deprecated HTTPScaledObject.
New tests are in the test/ subfolder - could you adjust the code?

linkvt (changes requested):

Sanil2108:
Thanks for the pointer, @linkvt! Moved the test from `tests/checks/upgrade_httpso/` to `test/e2e/upgrade/` as suggested. The `make undeploy` relative path (`../../..`) is unchanged since both locations are three levels below the repo root.

linkvt:
Sorry but it seems you just moved the code there without making any effort to understand how the e2e tests work in general.
This seems like an AI generated change, at least the principle "The human-in-the-loop must fully understand all code." does not seem to apply here.
I don't really see why I then should make an effort reviewing this PR.

linkvt:
Closing due to inactivity and the current approach being still a bit away from a mergable solution.

PR #1626, opened 2026-05-08
feat: add coldStart.placeholder support for static responses

### LINKED ISSUE #874: "Please hold" landing pages for slow scale from zero scenarios
### Proposal

If the service is scaling from zero and maybe needs to wait for a new node to provision a nicer user XP might be to just land on a page that says "Teamcity is starting up... please wait". Could be enabled and configured in the http scaled object resource. MVP could be a very simple page saying "{SomeValueFromHttpResource} is start up please wait" and it polls.

### Use-Case

Nodes can be slow (well a cpl of minutes) to start up. The experience scaling from zero in this case is pretty ordinary. The default wait time is only 30s and then you Bad Gateway or something like that.

I nice web page telling people to sit tight and wait for the magic would be a lot nicer.

Maybe

### Anything else?

_No response_

linkvt:
When scaling from zero, the interceptor can now return a static HTTP response immediately instead of blocking until the backend is ready. The response is configured via coldStart.placeholder.response on the InterceptorRoute CRD.

The current approach is flexible with these things in mind:
- we could add a similar `coldStart.fallback.response` field
- we could add `coldStart.placeholder.service` similar to the `coldStart.fallback.service`
- static responses for individual paths can have a similar structure as the response struct

### Example Resource
```yaml
apiVersion: http.keda.sh/v1beta1
kind: InterceptorRoute
metadata:
  name: whoami-ir
  namespace: default
spec:
  coldStart:
    placeholder:
      response:
        # body: "<h1>Loading...</h1>" # this OR bodyFromConfigMap or none of both for an empty response body
        bodyFromConfigMap:
          key: index.html
          name: placeholder-page
        headers:
          Content-Type: text/html
          Refresh: "5"
        statusCode: 503
...
```

### Decisions
There are a few tricky things when attempting to reimplement the solution proposed in #1303 with the same feature set, which is why this PR does not include the following yet:
- options for loading the body from a ConfigMap
	- use API client without a cache -> potentially high API load on burst triggering cold-start, general higher load as we always have to talk to the API server
	- use cached API client -> informer cache would cache all configmaps in all watched namespaces (by default all) which can increase memory usage possibly quite a lot
		- can be improved e.g. by filtering for configmaps with a specific label (current approach)
		- **👉 label filtering could be opt-in for simplicity? bit unsure whether we should have that as opt-in, opt-out or always-on**
	- use a custom solution -> IMO unneeded complexity
- templating responses
	- choosing `text/template` vs `html/template`
		- `html/template` is safe against XSS but might modify the template input in an unexpected way when using it in a json response -> not content type agnostic
		- the developer/admin controls in theory the template and the variables, so we could use `text/template`
		- what about custom input vars? probably optional/future feature
	- caching reduces load, cache expiration is something to consider though and should be based on the InterceptorRoute AND ConfigMap 
	- questions:
		- is templating really useful/wanted?
		- which fields would we expose, that are helpful for users? I.e. is `ServiceName` or `Namespace` something a user could use?
	- **👉 currently out of scope but easy to add in case we want it**
- global default template
	- difficult topic IMO, I'm more a fan of an explicit configuration as an implicit global default template could break everything at once
	- how would the global default template be configured? how/when/if would the template be reloaded? do we need to watch the file system in case of a mount and invalidate all template caches?
	- **👉 currently not included, not sure if we should include it? Is a `placeholder.service` maybe a better choice?**
- should `refreshInterval` be a dedicated field in the CRD?
	- pro
		- can be used inside of a template (-> requires templating)
	- con
		- equivalent approach is to set the [`Refresh` header](https://developer.mozilla.org/de/docs/Web/HTTP/Reference/Headers/Refresh) directly:
		  ```yaml
		  response:
		    headers:
		      Refresh: "5"
		  ```
	- **👉 the Refresh header is IMO so simple that a specific field for the interval is not really useful right now, refreshInterval also would require templating**
- should we have a default placeholder
	- personal preference for the body: explicit configuration especially when it's really simple to do, otherwise users might rely on our defaults that can change over time
	- should we require a body? IMO no, empty body is a valid response
	- should we require a status code? Bit unsure, 503 seems like a great candidate
	- **👉 default to empty response and status code 503 right now**
- caching headers
	- the consensus seems to be to not cache 500 or more specific 503 headers
	- **👉 IMO no need to have explicit caching header defaults**
- header implementation
	- HTTP allows to have multiple headers with the same key in theory
	- We could patch it later e.g. by splitting the header value by `\n` which seems to be a disallowed character for header values
	- **👉 for simplicity support a single header right now**

### Ideas
- derive content-type header from configmap file name if unset
	- decide whether this is always a good choice

### Changes
- Add StaticResponse, ConfigMapKeyRef, ColdStartPlaceholder types to CRD
- Add placeholder middleware between counting and endpoint resolver
	- I though a dedicated middleware keeps things simple and allows adding placeholder.service support without bloating the endpoint resolver middleware doing the fallback handling right now
- Support inline body or body from a labeled ConfigMap
	- inline body is limited to about 32k characters which is usually 32KB
	- this limit is an arbitrary choice, could even be way longer to put larger responses into a ConfigMap
- Scope ConfigMap cache with ByObject label selector to avoid caching all ConfigMaps cluster-wide
- Add unit tests, e2e test, and test helpers

### Checklist

- [x] Any necessary documentation is added, such as the [`README.md`](../README.md) or the [keda-docs](https://github.com/kedacore/keda-docs/tree/main/content/http-add-on)

Fixes #874
Supersedes #1303 (thanks again @malpou for the initial implementation ❤️ )


interceptor/middleware/placeholder.go
```diff
@@ -0,0 +1,103 @@
...
+		w.Header().Set(k, v)
+	}
+
+	statusCode := int(resp.StatusCode)
+	if statusCode == 0 {
+		statusCode = http.StatusServiceUnavailable
+	}
+	w.WriteHeader(statusCode)
+	if len(body) > 0 {
+		if _, err := w.Write(body); err != nil {
```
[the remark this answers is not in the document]
linkvt: /ar the body is controlled by the developer/admin and not by the user which makes it  safe IMO.

interceptor/middleware/placeholder.go
```diff
@@ -0,0 +1,104 @@
...
+	p.next.ServeHTTP(w, r)
+}
+
+func (p *Placeholder) serveStaticResponse(w http.ResponseWriter, r *http.Request, ir *httpv1beta1.InterceptorRoute, resp *httpv1beta1.StaticResponse) {
+	body, err := p.resolveBody(r, ir, resp)
+	if err != nil {
+		logger := util.LoggerFromContext(r.Context())
+		logger.Error(err, "failed to resolve placeholder body")
+		http.Error(w, "Internal Server Error", http.StatusInternalServerError)
+		return
```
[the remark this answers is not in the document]
linkvt: static is a full handler and expects the stream (upstream url) to be set, we dont need/have both here.

operator/apis/http/v1beta1/interceptorroute_types.go
```diff
@@ -141,18 +141,65 @@ type ServiceRef struct {
...
+}
+
+// ConfigMapKeyRef references a key within a ConfigMap.
+type ConfigMapKeyRef struct {
+	// Name of the ConfigMap.
+	// +kubebuilder:validation:MinLength=1
+	Name string `json:"name"`
+	// Key within the ConfigMap.
+	// +kubebuilder:validation:MinLength=1
+	Key string `json:"key"`
```
linkvt: TODO: consider supporting multiple keys (index.html, styles.css) or no keys e.g. to support all paths within a configmap?

This wouldn't require devs to put all their placeholder code into a single file.

Slashes can't be in configmap keys, `css/styles.css` could e.g. be URL encoded in the config map keys?
Path could be used for a configmap key lookup, empty path = index.html?

interceptor/proxy.go
```diff
@@ -81,6 +81,8 @@ func BuildProxyHandler(cfg *ProxyHandlerConfig) http.Handler {
 		EnableColdStartHeader: cfg.Serving.EnableColdStartHeader,
 	})
 
+	h = middleware.NewPlaceholder(h, cfg.ReadyCache, cfg.Reader)
```
Fedosin, line 84: Nit: the middleware position here is architecturally significant — it must be after Counting (so the scaler sees demand) and before EndpointResolver (to avoid blocking until the backend scales up). A short comment would help future maintainers understand why it's placed here. Something like:

```go
// Serve a static placeholder response when configured and no endpoints are ready.
// Placed after counting (so the scaler sees demand) and before endpoint resolution
// (to avoid blocking until the backend scales up).
h = middleware.NewPlaceholder(h, cfg.ReadyCache, cfg.Reader)
```

pkg/k8s/labels.go
```diff
@@ -0,0 +1,7 @@
+package k8s
+
+import "k8s.io/apimachinery/pkg/labels"
+
+// ResponseBodyLabels is the label set that ConfigMaps must carry to be
+// included in the interceptor's informer cache.
+var ResponseBodyLabels = labels.Set{"http.keda.sh/response-body": "true"}
```
Fedosin, line 7: Nit: the comment says the label key must be present, but the selector actually requires the value to be exactly `"true"`. Users might think any value works. Consider updating to something like:

```go
// ResponseBodyLabels is the label set that ConfigMaps must carry to be
// included in the interceptor's informer cache.
// ConfigMaps must have the label "http.keda.sh/response-body" set to "true".
var ResponseBodyLabels = labels.Set{"http.keda.sh/response-body": "true"}
```

interceptor/middleware/placeholder_test.go
```diff
@@ -0,0 +1,262 @@
...
+}
+
+func newPlaceholderRequest(t *testing.T, ir *httpv1beta1.InterceptorRoute) *http.Request {
+	t.Helper()
+
+	req := httptest.NewRequest("GET", "/test", nil)
+	ctx := util.ContextWithLogger(req.Context(), logr.Discard())
+	ctx = util.ContextWithInterceptorRoute(ctx, ir)
+	return req.WithContext(ctx)
+}
```
Fedosin, line 299: Could you add a test for the case where both `fallback` and `placeholder` are configured on the same InterceptorRoute? Since `Placeholder` sits upstream of `EndpointResolver` (which handles fallback), the placeholder should fire first when there are no ready endpoints. A test confirming this interaction would be valuable — e.g. verifying that the placeholder response is returned immediately and the fallback is never reached when endpoints aren't ready.

pkg/k8s/labels.go
linkvt: I don't see where the comment says, that the label **key** must be present - it says that the label **set** must be carried by the ConfigMap

interceptor/proxy.go
linkvt: I think this applies to any middleware, but think that verifying the scaling in an e2e test is a good idea and achieves the same 👍

interceptor/middleware/placeholder_test.go
linkvt: Good idea, should be an easy e2e test 👍

linkvt: E2E test was added

interceptor/proxy.go
linkvt: E2E test added that verifies that scaling up is triggered when using placeholder responses, this ensures Counting is called before Placeholder.

linkvt:
Hi @JorTurFer and @wozniakjan , both of you did review the initial placeholder response PR #1303 - I would appreciate your thoughts on this variant if you find time :)

Fedosin (approved):

interceptor/middleware/placeholder.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	statusCode := int(resp.StatusCode)
+	if statusCode == 0 {
+		statusCode = http.StatusServiceUnavailable
+	}
+	w.WriteHeader(statusCode)
+
+	if body != "" {
+		if _, err := io.WriteString(w, body); err != nil {
```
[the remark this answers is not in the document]
linkvt: Still not an issue



/ar the body is controlled by the developer/admin and not by the user which makes it safe IMO.

linkvt:
As an experiment I made the key optional and added auto detection so that you can put `index.html` and `style.css` into a configmap which get's auto loaded.

Could also move this into a separate PR, I kept it in a separate commit in case we don't want it.

Fedosin (approved):

linkvt:
Push since approve was just a rebase on main, merging now.

Issue #965, opened 2024-04-04
Observability for all the components


JorTurFer:
We have to implement and document observability metrics for all the components:
- [ ] Operator
- [x] https://github.com/kedacore/http-add-on/issues/910
- [ ] External Scaler
- [ ] https://github.com/kedacore/http-add-on/issues/354
- [x] https://github.com/kedacore/http-add-on/issues/1031

linkvt:
Unstale

Fedosin:
Remaining observability was added in KHA 0.15.0.

PR #1634, opened 2026-05-13
feat(scaler): add observability (metrics + tracing) to the external scaler

Fedosin:
Add OpenTelemetry-based metrics and distributed tracing to the external
scaler component, which previously had no observability instrumentation.

Shared observability infrastructure is extracted into `pkg/observability/`
so both the interceptor and scaler reuse the same tracing setup, metrics
provider, and configuration types.

**Metrics:**
- `scaler.pinger.fetch.duration` (histogram) — duration of each queue pinger fetch cycle
- `scaler.pinger.fetch.errors` (counter) — total failed pinger fetch cycles
- `scaler.pinger.endpoints` (gauge) — number of interceptor endpoints being polled
- Prometheus `/metrics` endpoint on port 2223 (configurable via `OTEL_PROM_EXPORTER_PORT`)
- Optional OTLP HTTP metrics export (via `OTEL_EXPORTER_OTLP_METRICS_ENABLED`)

**Tracing:**
- OTEL tracing SDK with console, HTTP/protobuf, and gRPC exporters
- `otelgrpc` stats handler for automatic gRPC server span instrumentation
- W3C TraceContext + Baggage propagation

**Configuration env vars** (same as the interceptor):
- `OTEL_PROM_EXPORTER_ENABLED` (default: `true`)
- `OTEL_PROM_EXPORTER_PORT` (default: `2223`)
- `OTEL_EXPORTER_OTLP_METRICS_ENABLED` (default: `false`)
- `OTEL_EXPORTER_OTLP_TRACES_ENABLED` (default: `false`)
- `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL` (default: `console`)

### Checklist

  - [`README.md`](../README.md)
  - [keda-docs](https://github.com/kedacore/keda-docs/tree/main/content/http-add-on)

Part of https://github.com/kedacore/http-add-on/issues/965

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ This changelog keeps track of work items that have been completed and are ready 
 
 - **General**: Add `coldStart.placeholder` support for returning static HTTP responses during scale-from-zero ([#874](https://github.com/kedacore/http-add-on/issues/874))
 - **General**: TODO ([#TODO](https://github.com/kedacore/http-add-on/issues/TODO))
+- **Scaler**: Add OpenTelemetry metrics and distributed tracing to the external scaler ([#965](https://github.com/kedacore/http-add-on/issues/965))
 
 ### Improvements
 
--- a/interceptor/config/metrics.go
+++ b/interceptor/config/metrics.go
@@ -1,21 +1,9 @@
 package config
 
 import (
-	"github.com/caarlos0/env/v11"
+	"github.com/kedacore/http-add-on/pkg/observability"
 )
 
-// Metrics is the configuration for configuring metrics in the interceptor.
-type Metrics struct {
-	// Sets whether or not to enable the Prometheus metrics exporter
-	OtelPrometheusExporterEnabled bool `env:"OTEL_PROM_EXPORTER_ENABLED" envDefault:"true"`
-	// Sets the port which the Prometheus compatible metrics endpoint should be served on
-	OtelPrometheusExporterPort int `env:"OTEL_PROM_EXPORTER_PORT" envDefault:"2223"`
-	// Sets whether or not to enable the OTEL metrics exporter
-	OtelHTTPExporterEnabled bool `env:"OTEL_EXPORTER_OTLP_METRICS_ENABLED" envDefault:"false"`
-}
-
-// MustParseMetrics parses standard configs and returns the
-// newly created config. Panics if parsing fails.
-func MustParseMetrics() Metrics {
-	return env.Must(env.ParseAs[Metrics]())
-}
+// Metrics is an alias for the shared observability metrics config.
+// Kept to avoid refactoring all existing interceptor code that references config.Metrics.
+type Metrics = observability.MetricsConfig
--- a/scaler/config.go
+++ b/scaler/config.go
@@ -4,6 +4,8 @@ import (
 	"time"
 
 	"github.com/caarlos0/env/v11"
+
+	"github.com/kedacore/http-add-on/pkg/observability"
 )
 
 type config struct {
@@ -30,6 +32,9 @@ type config struct {
 	ProfilingAddr string `env:"PROFILING_BIND_ADDRESS" envDefault:""`
 	// StreamIntervalMS is the interval in milliseconds between stream ticks
 	StreamIntervalMS int `env:"KEDA_HTTP_SCALER_STREAM_INTERVAL_MS" envDefault:"200"`
+
+	Metrics observability.MetricsConfig `envPrefix:""`
+	Tracing observability.TracingConfig `envPrefix:""`
 }
 
 func mustParseConfig() config {
--- a/scaler/main.go
+++ b/scaler/main.go
@@ -13,10 +13,13 @@ import (
 	"net/http"
 	_ "net/http/pprof" //nolint:gosec // G108: pprof intentionally exposed, gated by --profiling-addr
 	"os"
+	"runtime"
 	"time"
 
 	"github.com/go-logr/logr"
 	"github.com/kedacore/keda/v2/pkg/scalers/externalscaler"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc"
 	"golang.org/x/sync/errgroup"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/health"
@@ -29,8 +32,11 @@ import (
 
 	"github.com/kedacore/http-add-on/pkg/build"
 	kedacache "github.com/kedacore/http-add-on/pkg/cache"
+	kedahttp "github.com/kedacore/http-add-on/pkg/http"
 	"github.com/kedacore/http-add-on/pkg/k8s"
+	"github.com/kedacore/http-add-on/pkg/observability"
 	"github.com/kedacore/http-add-on/pkg/util"
+	"github.com/kedacore/http-add-on/scaler/metrics"
 )
 
 var setupLog = ctrl.Log.WithName("setup")
@@ -40,6 +46,7 @@ var setupLog = ctrl.Log.WithName("setup")
 // +kubebuilder:rbac:groups=http.keda.sh,resources=interceptorroutes,verbs=get;list;watch
 
 func main() {
+	defer os.Exit(1)
 	cfg := mustParseConfig()
 	namespace := cfg.TargetNamespace
 	svcName := cfg.TargetService
@@ -53,10 +60,33 @@ func main() {
 
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
 
+	setupLog.Info(
+		"starting scaler",
+		"metricsConfig", cfg.Metrics,
+		"tracingConfig", cfg.Tracing,
+	)
+
+	provider, err := observability.NewMeterProvider(metrics.ServiceName, cfg.Metrics)
+	if err != nil {
+		setupLog.Error(err, "failed to create meter provider")
+		runtime.Goexit()
+	}
+	defer func() {
+		if err := provider.Shutdown(context.Background()); err != nil {
+			setupLog.Error(err, "error shutting down meter provider")
+		}
+	}()
+
+	instruments, err := metrics.NewInstruments(provider)
+	if err != nil {
+		setupLog.Error(err, "failed to create metric instruments")
+		runtime.Goexit()
+	}
+
 	k8sCfg, err := ctrl.GetConfig()
 	if err != nil {
 		setupLog.Error(err, "Kubernetes client config not found")
-		os.Exit(1)
+		runtime.Goexit()
 	}
 
 	ctrlCache, err := cache.New(k8sCfg, cache.Options{
@@ -66,15 +96,28 @@ func main() {
 	})
 	if err != nil {
 		setupLog.Error(err, "creating cache")
-		os.Exit(1)
+		runtime.Goexit()
 	}
 
-	pinger := newQueuePinger(ctrl.Log, k8s.EndpointsFuncForControllerClient(ctrlCache), namespace, svcName, deplName, targetPortStr)
+	pinger := newQueuePinger(ctrl.Log, k8s.EndpointsFuncForControllerClient(ctrlCache), namespace, svcName, deplName, targetPortStr, instruments)
 
 	ctx := ctrl.SetupSignalHandler()
 	ctx = util.ContextWithLogger(ctx, setupLog)
 
 	eg, ctx := errgroup.WithContext(ctx)
+
+	if cfg.Tracing.Enabled {
+		shutdown, err := observability.SetupTracing(ctx, metrics.ServiceName, cfg.Tracing)
+		if err != nil {
+			setupLog.Error(err, "error setting up tracer")
+			runtime.Goexit()
+		}
+		defer func() {
+			if shutdownErr := shutdown(context.Background()); shutdownErr != nil {
+				setupLog.Error(shutdownErr, "error during tracer shutdown")
+			}
+		}()
+	}
 
 	// start the controller-runtime cache
 	eg.Go(func() error {
@@ -96,7 +139,7 @@ func main() {
 	// Wait for cache to sync before starting components that depend on it
 	if !ctrlCache.WaitForCacheSync(ctx) {
 		setupLog.Error(nil, "cache failed to sync")
-		os.Exit(1)
+		runtime.Goexit()
 	}
 
 	eg.Go(func() error {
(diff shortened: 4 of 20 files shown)


scaler/queue_pinger.go
```diff
@@ -152,7 +155,11 @@ func (q *queuePinger) fetchAndSaveCounts(ctx context.Context) error {
 	q.pingMut.Lock()
 	defer q.pingMut.Unlock()
 
+	fetchStart := time.Now()
 	perPod, err := fetchCountsPerPod(ctx, q.lggr, q.getEndpointsFn, q.interceptorNS, q.interceptorSvcName, q.adminPort)
+	if q.instruments != nil {
+		q.instruments.RecordFetch(time.Since(fetchStart), len(perPod), err)
```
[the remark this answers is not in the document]
Fedosin: Fixed in the latest push. `fetchCountsPerPod` now returns a `fetchResult` struct that includes `endpointCount` (the number of endpoints targeted for polling), which is populated even when the fetch errors out. The gauge now correctly reflects how many endpoints were polled regardless of success/failure.

scaler/metrics/provider.go
```diff
@@ -0,0 +1,51 @@
...
+			sdkmetric.NewPeriodicReader(otlpExporter),
+		))
+	}
+
+	options = append(options, sdkmetric.WithResource(
+		resource.NewSchemaless(
+			attribute.String("service.name", serviceName),
+			attribute.String("service.version", build.Version()),
+		),
+	))
```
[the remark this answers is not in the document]
Fedosin: Fixed. Both metrics and tracing now use the same `"keda-http-external-scaler"` value via a shared exported constant (`metrics.ServiceName`) so telemetry correlates cleanly in backends.

scaler/main.go
```diff
@@ -53,10 +59,36 @@ func main() {
...
+	if err != nil {
+		setupLog.Error(err, "failed to create metric instruments")
+		runtime.Goexit()
+	}
+
 	k8sCfg, err := ctrl.GetConfig()
 	if err != nil {
 		setupLog.Error(err, "Kubernetes client config not found")
-		os.Exit(1)
+		runtime.Goexit()
```
linkvt, line 89: Is there a reason for using runtime.Goexit now? If this is intended we should probably also add the `defer os.Exit(1)` at top as in the interceptor to also stop the grpc server etc after `runtime.Goexit` has been called.

scaler/metrics/instruments.go
```diff
@@ -0,0 +1,81 @@
...
+
+	// ServiceName is the OTEL service.name used for both metrics and tracing.
+	ServiceName = "keda-http-external-scaler"
+
+	MetricPingerFetchDuration = "scaler.pinger.fetch.duration"
+	MetricPingerFetchErrors   = "scaler.pinger.fetch.errors"
+	MetricPingerEndpoints     = "scaler.pinger.endpoints"
+
+	AttrNamespace = "namespace"
+	AttrService   = "service"
```
linkvt: unused vars

scaler/tracing/tracing.go
```diff
@@ -0,0 +1,91 @@
+package tracing
```
linkvt: This looks like a copy of interceptor/tracing/tracing.go, we should deduplicate this code

scaler/metrics/provider.go
```diff
@@ -0,0 +1,51 @@
+package metrics
```
linkvt: This looks like a copy of interceptor/provider/metrics.go, we should deduplicate this code.

scaler/queue_pinger.go
```diff
@@ -152,14 +155,20 @@ func (q *queuePinger) fetchAndSaveCounts(ctx context.Context) error {
 	q.pingMut.Lock()
 	defer q.pingMut.Unlock()
 
-	perPod, err := fetchCountsPerPod(ctx, q.lggr, q.getEndpointsFn, q.interceptorNS, q.interceptorSvcName, q.adminPort)
+	fetchStart := time.Now()
+	result, err := fetchCountsPerPod(ctx, q.lggr, q.getEndpointsFn, q.interceptorNS, q.interceptorSvcName, q.adminPort)
+	if q.instruments != nil {
```
linkvt: We could use the same pattern as in the interceptor instruments with `NewNoopInstruments()` and pass that into the components? This makes the code cleaner as we avoid adding a special case for instruments being nil.

scaler/metrics/instruments.go
```diff
@@ -0,0 +1,81 @@
...
+	"go.opentelemetry.io/otel/attribute"
+	api "go.opentelemetry.io/otel/metric"
+	sdkmetric "go.opentelemetry.io/otel/sdk/metric"
+)
+
+const (
+	meterName = "keda-external-scaler"
+
+	// ServiceName is the OTEL service.name used for both metrics and tracing.
+	ServiceName = "keda-http-external-scaler"
```
linkvt, line 16: We should probably also align the service name of the interceptor to use this naming scheme

scaler/config.go
```diff
@@ -30,6 +30,20 @@ type config struct {
 	ProfilingAddr string `env:"PROFILING_BIND_ADDRESS" envDefault:""`
 	// StreamIntervalMS is the interval in milliseconds between stream ticks
 	StreamIntervalMS int `env:"KEDA_HTTP_SCALER_STREAM_INTERVAL_MS" envDefault:"200"`
+
```
linkvt, line 35: I guess we could also deduplicate the config to ensure it is consistent across components?

scaler/config.go
```diff
@@ -30,6 +30,20 @@ type config struct {
...
+	Metrics metricsConfig `envPrefix:""`
+	Tracing tracingConfig `envPrefix:""`
+}
+
+type metricsConfig struct {
+	OtelPrometheusExporterEnabled bool `env:"OTEL_PROM_EXPORTER_ENABLED" envDefault:"true"`
+	OtelPrometheusExporterPort    int  `env:"OTEL_PROM_EXPORTER_PORT" envDefault:"2224"`
```
linkvt: Is there a reason to not use the same port we use for the interceptor metrics?

scaler/metrics/instruments.go
```diff
@@ -0,0 +1,81 @@
...
+	return &Instruments{
+		pingerFetchDuration: pingerFetchDuration,
+		pingerFetchErrors:   pingerFetchErrors,
+		pingerEndpoints:     pingerEndpoints,
+	}, nil
+}
+
+// RecordFetch records a completed pinger fetch cycle.
+func (i *Instruments) RecordFetch(duration time.Duration, endpointCount int, fetchErr error) {
+	attrs := api.WithAttributeSet(attribute.NewSet())
```
linkvt: Can be removed as there are no attributes

scaler/metrics/instruments.go
```diff
@@ -0,0 +1,81 @@
+package metrics
```
linkvt, line 1: We should probably also add a test like the prometheus_test.go the interceptor has right now?

linkvt (changes requested):
I left a few comments, we can probably reduce the size quite a bit after deduplicating code we already have in the interceptor.

The PR description should also probably not say "Fixes #..." to avoid auto closing the issue.
We should also keep in mind to update the helm chart and the resources in config/ in this repo.

scaler/main.go
Fedosin: Fixed — added `defer os.Exit(1)` at the top of main (same pattern as the interceptor). All subsequent failures use `runtime.Goexit()` to ensure defers run.

scaler/metrics/instruments.go
Fedosin: Removed.

scaler/tracing/tracing.go
Fedosin: Deduplicated — extracted shared tracing setup into `pkg/observability/tracing.go`. Both interceptor and scaler now delegate to it.

scaler/metrics/provider.go
Fedosin: Deduplicated — extracted shared meter provider factory into `pkg/observability/metrics.go`. Both interceptor and scaler delegate to it with their respective service names.

scaler/queue_pinger.go
Fedosin: Done — added `NewNoopInstruments()` and all tests now use it instead of nil. The nil check in `fetchAndSaveCounts` is removed.

scaler/metrics/instruments.go
Fedosin: Good point. The interceptor already uses `keda-http-interceptor` as its service name (set in `interceptor/tracing/tracing.go`). I think that's consistent with the scaler's `keda-http-external-scaler`. Should we rename the interceptor to something like `keda-http-interceptor-proxy` or is `keda-http-interceptor` fine?

scaler/config.go
Fedosin: Done — extracted `MetricsConfig` and `TracingConfig` into `pkg/observability/config.go`. Both interceptor and scaler use type aliases to it.

scaler/config.go
Fedosin: Changed to 2223 (same as the interceptor).

scaler/metrics/instruments.go
Fedosin: Removed.

scaler/metrics/instruments.go
Fedosin: Added `scaler/metrics/prometheus_test.go` that verifies the histogram, counter, and gauge are correctly emitted.

Fedosin:
Thanks for the detailed review @linkvt — addressed all inline comments in the latest force-push (`f8a351b`):

- deduplicated tracing + meter provider into `pkg/observability/`
- deduplicated metrics/tracing config types (`MetricsConfig`, `TracingConfig`)
- switched scaler metrics default port to `2223` (same as interceptor)
- added `defer os.Exit(1)` + `runtime.Goexit()` pattern in scaler main
- removed unused vars and empty attribute set
- added `NewNoopInstruments()` and removed nil special-casing
- added `scaler/metrics/prometheus_test.go`
- updated PR description to use `Part of #965` instead of `Fixes`

On your note about Helm/config resources: this repo currently has `config/` manifests but no Helm chart directory. Since these env vars are optional and have defaults, runtime behavior is unchanged unless users set them. I can add explicit env var wiring to `config/scaler/deployment.yaml` in this PR too if you’d prefer that visibility.

scaler/metrics/instruments.go
linkvt: Yes but the interceptors instrument are using the name `"interceptor-proxy"` which we should align with this pattern

CHANGELOG.md
```diff
@@ -33,6 +33,7 @@ This changelog keeps track of work items that have been completed and are ready
 ### New
 
 - **General**: TODO ([#TODO](https://github.com/kedacore/http-add-on/issues/TODO))
+- **Scaler**: Add OpenTelemetry metrics and distributed tracing to the external scaler ([#965](https://github.com/kedacore/http-add-on/issues/965))
```
linkvt, line 37: We should also need to update the helm chart and the resources in config/ in this repo.

scaler/metrics/provider.go
linkvt: Is there a reason/benefit for this indirection or couldnt we just use `observability.NewMeterProvider` directly? Using it directly seems simpler.

interceptor/config/metrics.go
```diff
@@ -1,21 +1,14 @@
...
 // Metrics is the configuration for configuring metrics in the interceptor.
-type Metrics struct {
-	// Sets whether or not to enable the Prometheus metrics exporter
-	OtelPrometheusExporterEnabled bool `env:"OTEL_PROM_EXPORTER_ENABLED" envDefault:"true"`
-	// Sets the port which the Prometheus compatible metrics endpoint should be served on
-	OtelPrometheusExporterPort int `env:"OTEL_PROM_EXPORTER_PORT" envDefault:"2223"`
-	// Sets whether or not to enable the OTEL metrics exporter
-	OtelHTTPExporterEnabled bool `env:"OTEL_EXPORTER_OTLP_METRICS_ENABLED" envDefault:"false"`
-}
+type Metrics = observability.MetricsConfig
```
linkvt, line 9: Is there a reason for the type alias? If the code using this config is also from the observability package I see no reason to alias it?
Same with the Tracing config.

scaler/metrics/instruments.go
Fedosin: Done — renamed the interceptor metrics service name from `"interceptor-proxy"` to `"keda-http-interceptor"`, aligning with the `keda-http-*` convention used by the scaler (`keda-http-external-scaler`).

CHANGELOG.md
Fedosin: Added the metrics port (2223) to `config/scaler/deployment.yaml`. There's no Helm chart in this repo — the chart lives in the [kedacore/charts](https://github.com/kedacore/charts) repo and would need a separate PR there once this merges.

scaler/metrics/provider.go
Fedosin: Agreed — removed the indirection. The scaler now calls `observability.NewMeterProvider(metrics.ServiceName, cfg.Metrics)` directly in `main.go`. The `scaler/metrics/provider.go` file is deleted.

interceptor/config/metrics.go
Fedosin: The type alias exists only to avoid a mass-refactor of all interceptor files that reference `config.Metrics` / `config.Tracing` (15+ files). It's zero-cost at runtime and keeps this PR focused on scaler observability. The `MustParse*` wrapper functions are removed though — `interceptor/main.go` now calls `observability.MustParseMetricsConfig()` directly.

Happy to remove the aliases entirely in a follow-up PR if you prefer.

scaler/metrics/provider.go
linkvt: interceptor/metrics/provider.go still wraps observability.NewMeterProvider

scaler/tracing/tracing.go
linkvt: Now interceptor/tracing/tracing.go is still wrapping `observability.SetupTracing` while the scaler uses it directly

CHANGELOG.md
linkvt: Yes I know, this PR is pretty final from a config perspective so I guess there is nothing blocking the other PR so that this topic doesn't fall of the radar.

Fedosin:
@linkvt All review concerns have been addressed in the latest push (`05e4acab`):

- Deduplicated tracing, metrics provider, and config into `pkg/observability/`
- Removed `scaler/metrics/provider.go` indirection — now calling `observability.NewMeterProvider` directly
- Aligned interceptor service name to `keda-http-interceptor`
- Added metrics port to `config/scaler/deployment.yaml`
- Removed MustParse wrappers (type aliases kept to avoid a 15+ file refactor in this PR)
- Port set to 2223 (same as interceptor)
- PR description updated to "Part of #965"

Ready for another look when you get a chance. Thanks!

scaler/metrics/provider.go
Fedosin: Done — removed the indirection entirely. `scaler/metrics/provider.go` is deleted and `main.go` calls `observability.NewMeterProvider(metrics.ServiceName, cfg.Metrics)` directly.

interceptor/config/metrics.go
Fedosin: The type alias is only kept to avoid a 15+ file refactor of existing interceptor code that references `config.Metrics` / `config.Tracing` — it's zero-cost and keeps this PR scoped to scaler changes. The `MustParse*` wrapper functions are removed though. Happy to drop the aliases in a follow-up if preferred.

Fedosin:
@linkvt Pushed `f7fb8905` — removed the `interceptor/metrics/provider.go` wrapper as well. Both the interceptor and scaler now call `observability.NewMeterProvider(serviceName, cfg)` directly with no indirection layer.

The `ServiceName` constant is exported from both `interceptor/metrics` and `scaler/metrics` so callers pass it explicitly.

linkvt (approved):
Looks good, thanks!
Also just to make sure, we shouldnt forget the helm chart update :)

PR #1649, opened 2026-05-20
feat: add SLSA Level 3 provenance to release pipeline

### LINKED ISSUE #1604: SLSA Build L2 provenance attestations for release images
Add [SLSA](https://slsa.dev/) (Supply-chain Levels for Software Artifacts) provenance attestations to the release workflow, bringing the project from SLSA Build L0 to at least Build L2.
The following is the result from an AI analysis in the repo and might contain mistakes, I checked it quickly and think it is a good start that contains valuable information.

### Background

The project currently does not generate SLSA provenance attestations, putting it at **SLSA Build Level 0** for provenance.

We already have strong foundations:
- **Cosign image signing** — keyless OIDC via GitHub Actions (proves *who* published the image)
- **SBOMs** — generated automatically by ko in SPDX format (describes *what* went into it)
- **SHA-pinned GitHub Actions** — all actions pinned to full SHA256 hashes
- **Minimal workflow permissions** — least-privilege across all workflows

What's missing is **provenance** — a signed attestation describing *how* the image was built (source commit, workflow, build platform). Cosign signing alone doesn't prevent a compromised maintainer from building locally with modified source; SLSA provenance does.

### SLSA Build Levels

| Level | Requirements | Our status |
|-------|--------------|------------|
| **L0** | No requirements | Current state for provenance |
| **L1** | Provenance exists and is distributed to consumers (may be unsigned) | Not met — no provenance document is generated |
| **L2** | L1 + builds run on a hosted platform + provenance is cryptographically signed | Partially met — we already build on GitHub Actions (hosted) and sign images with cosign, but no provenance document exists to sign |
| **L3** | L2 + cross-build isolation + signing keys not accessible to user-defined build steps | Not met — would require splitting the workflow so provenance is generated in an isolated reusable workflow |

### Proposed approach: Implement SLSA Build L2

Add `actions/attest-build-provenance` after each `ko build` in `build_release.yml`. This generates signed provenance using GitHub's OIDC identity. Low effort (~20 lines of YAML per image), high value.

Consumers verify with:
```bash
gh attestation verify oci://ghcr.io/kedacore/http-add-on-operator:0.9.0 -R kedacore/http-add-on
```

#### Optional: SLSA Build L3 (full isolation)

<details>
  <summary>This is likely something we don't want to do right as it is way more complex - this is absolutely OK right now IMO.
</summary>

Use `slsa-framework/slsa-github-generator`'s `generator_container_slsa3.yml` as a reusable workflow. Split the release workflow into:
1. **Build job**: ko build → capture image digests
2. **Provenance jobs (3x)**: slsa-github-generator reusable workflow (isolated, signing keys not accessible to build steps)
3. **Sign job**: cosign sign
4. **Release job**: upload manifests

This adds ~2-3 min to the release pipeline but provides the strongest guarantees.

Consumers verify with:
```bash
slsa-verifier verify-image ghcr.io/kedacore/http-add-on-operator:0.15.0 \
  --source-uri github.com/kedacore/http-add-on
```

</details>

### Tasks
- [ ] Implement SLSA Build L2

AtharvaPakade:
Implements SLSA Build Level 3 provenance for all release artifacts, resolving the provenance gap identified in the issue. The implementation goes directly to L3 rather than L2, as discussed in the issue thread.

  ### Changes

  The release workflow (`build_release.yml`) is restructured from a single monolithic job into a multi-job pipeline:

  - **`template-build-sign.yml`** — new reusable workflow that builds all three images with `ko` and signs each by immutable index digest via cosign keyless OIDC. Both `build_release.yml` and `build_canary.yml` delegate to it.
  - **`provenance-{operator,interceptor,scaler}` jobs** — each calls `slsa-framework/slsa-github-generator`'s `generator_container_slsa3.yml` as a reusable workflow on an isolated VM. The signing key never touches the build job, satisfying the L3 cross-build isolation requirement.
  - **`manifests` job** — generates deployment YAML with pinned digest refs, computes SHA-256 hashes, signs blobs with `cosign sign-blob`, and uploads artifacts. Runs manifest validation to fail fast if any `ko://` placeholder survives substitution.
  - **`provenance-manifests` job** — generates SLSA L3 generic provenance for the deployment manifests via `generator_generic_slsa3.yml`.
  - **`release` job** — gated on all provenance jobs completing, ensuring consumers always find attestations alongside release assets.

  Makefile changes:
  - Replace `publish-*` targets (signed by mutable version/SHA tags) with `ko-publish-*` targets that print `image@sha256:<digest>` to stdout and tag by full commit SHA instead of short SHA.
  - Replace `release` target with `generate-manifests` (sed-substitutes pinned digest refs into kustomize output) with post-generation assertions.
  - `sign-images` now signs by immutable digest ref (`image@sha256:...`) instead of mutable `:version` and `:sha` tags.

  Renovate: exclude `slsa-framework/slsa-github-generator` from digest pinning per the project's [own
  guidance](https://github.com/slsa-framework/slsa-github-generator/blob/4d014fae4dbd39eb09e8d40348b73db095e6ba9a/RENOVATE.md).

  `SECURITY.md` updated with artifact verification instructions for both `slsa-verifier` and `cosign`, covering container images and deployment manifests. Canary images are explicitly noted as cosign-signed only (no SLSA provenance by design).

  ### Verifying the attestations

  ```bash
  VERSION=<release-version>
  COMPONENT=interceptor   # or: operator, scaler
  IMAGE="ghcr.io/kedacore/http-add-on-${COMPONENT}"

  digest=$(crane digest "${IMAGE}:${VERSION}")

  slsa-verifier verify-image "${IMAGE}@${digest}" \
    --source-uri github.com/kedacore/http-add-on
  ```
 ### Proof of concept run

  [Workflow run 26163858450](https://github.com/AtharvaPakade/keda-http-add-on/actions/runs/26163858450) — all provenance jobs passed on the fork. The images and manifests below can be used to reproduce verification against the fork's registry:

  **Container images**

  ```bash
  # Operator
  slsa-verifier verify-image \
    ghcr.io/atharvapakade/http-add-on-operator@sha256:30ba4de42d996528121a8e1397c8dda385dfa0dd1ce842f50f49390606892311 \
    --source-uri github.com/AtharvaPakade/keda-http-add-on
  # Verified build using builder "https://github.com/slsa-framework/slsa-github-generator/.github/workflows/generator_container_slsa3.yml@refs/tags/v2.1.0" at commit 966bdf40aed1d1141b617899cb1b574be5a7cfd1
  # PASSED: SLSA verification passed

  # Interceptor
  slsa-verifier verify-image \
    ghcr.io/atharvapakade/http-add-on-interceptor@sha256:7139d74d4b397b6de72a4111d7792be1024ce09565f98805e2ba551629215fcb \
    --source-uri github.com/AtharvaPakade/keda-http-add-on
  # Verified build using builder "https://github.com/slsa-framework/slsa-github-generator/.github/workflows/generator_container_slsa3.yml@refs/tags/v2.1.0" at commit 966bdf40aed1d1141b617899cb1b574be5a7cfd1
  # PASSED: SLSA verification passed

  # Scaler
  slsa-verifier verify-image \
    ghcr.io/atharvapakade/http-add-on-scaler@sha256:eb3701d676d3c3d81bff201f0a5d3425672d78bf7a2c4f376381bf43bc98de4f \
    --source-uri github.com/AtharvaPakade/keda-http-add-on
  # Verified build using builder "https://github.com/slsa-framework/slsa-github-generator/.github/workflows/generator_container_slsa3.yml@refs/tags/v2.1.0" at commit 966bdf40aed1d1141b617899cb1b574be5a7cfd1
  # PASSED: SLSA verification passed
  ```

  **Deployment manifests** (download artifacts from the run, then verify)

  ```bash
  # Download artifacts (requires gh CLI)
 gh run download 26163858450 --repo AtharvaPakade/keda-http-add-on --name "manifests-0.0.4-test"
 gh run download 26163858450 --repo AtharvaPakade/keda-http-add-on --name "keda-http-add-on-manifests.intoto.jsonl"
  # Full deployment manifest
  slsa-verifier verify-artifact "keda-add-ons-http-0.0.4-test.yaml" \
    --provenance-path "keda-http-add-on-manifests.intoto.jsonl" \
    --source-uri github.com/AtharvaPakade/keda-http-add-on
  # Verified build using builder "https://github.com/slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@refs/tags/v2.1.0" at commit 966bdf40aed1d1141b617899cb1b574be5a7cfd1
  # Verifying artifact keda-add-ons-http-0.0.4-test.yaml: PASSED
  # PASSED: SLSA verification passed

  # CRD manifest
  slsa-verifier verify-artifact "keda-add-ons-http-0.0.4-test-crds.yaml" \
    --provenance-path "keda-http-add-on-manifests.intoto.jsonl" \
    --source-uri github.com/AtharvaPakade/keda-http-add-on
  # Verified build using builder "https://github.com/slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@refs/tags/v2.1.0" at commit 966bdf40aed1d1141b617899cb1b574be5a7cfd1
  # Verifying artifact keda-add-ons-http-0.0.4-test-crds.yaml: PASSED
  # PASSED: SLSA verification passed
  ```

  ### Notes for Reviewers

  - The `slsa-github-generator` reusable workflows are referenced by semver tag (`@v2.1.0`) rather than a commit SHA. This is the approach explicitly recommended by the generator project — their release tags are protected from force-push. Renovate is configured to skip digest pinning for this package.
  - Three explicit provenance jobs (`provenance-operator`, `provenance-interceptor`, `provenance-scaler`) are used instead of a matrix strategy. GitHub Actions matrix jobs share the same job ID namespace, so downstream jobs cannot reference individual matrix job outputs by component name (e.g., `needs.provenance.outputs.operator-digest` is not resolvable when `provenance` is a matrix job). Three named jobs give each component a distinct job ID and a statically addressable output, which is required for the `release` job's `needs` dependency and the digest forwarding in `get-digests`.
  - `actionlint` will emit warnings on `needs.build.outputs.*` in the `get-digests` pass-through job — this is a known tooling limitation with reusable workflow output resolution; the workflow is correct at runtime.
  - Canary builds intentionally have no SLSA provenance. This was discussed in the issue and is documented in `SECURITY.md`.

  ### References

  - [SLSA Build L3 specification](https://slsa.dev/spec/v1.1/levels#build-l3)
  - [SLSA Build Track basics](https://slsa.dev/spec/v1.2/build-track-basics)
  - [slsa-github-generator — container builder](https://github.com/slsa-framework/slsa-github-generator/tree/4d014fae4dbd39eb09e8d40348b73db095e6ba9a/internal/builders/container)
  - [slsa-github-generator — generic builder](https://github.com/slsa-framework/slsa-github-generator/tree/4d014fae4dbd39eb09e8d40348b73db095e6ba9a/internal/builders/generic)
  - [Referencing the SLSA generator](https://github.com/slsa-framework/slsa-github-generator/blob/4d014fae4dbd39eb09e8d40348b73db095e6ba9a/internal/builders/container/README.md#referencing-the-slsa-generator)
  - [slsa-github-generator Renovate config guidance](https://github.com/slsa-framework/slsa-github-generator/blob/4d014fae4dbd39eb09e8d40348b73db095e6ba9a/RENOVATE.md)

### Checklist

- [ ] Changelog is updated per our [changelog requirements](https://github.com/kedacore/http-add-on/blob/main/CONTRIBUTING.md#updating-the-changelog)

 Fixes #1604


.github/workflows/build_release.yml
```diff
@@ -3,80 +3,222 @@ name: Publish official image to GitHub Container Registry
 on:
   push:
     tags: ["v[0-9]+.[0-9]+.[0-9]+"]
+  # Added for fork testing — allows manual trigger without pushing a real semver tag
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "Version to use (without v prefix, e.g. 0.0.0-test)"
+        required: true
+        default: "0.0.0-test"
```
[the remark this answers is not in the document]
AtharvaPakade: `workflow_dispatch` is temporary for testing the SLSA pipeline end-to-end and will be removed before merge.

.github/workflows/build_release.yml
```diff
@@ -3,80 +3,222 @@ name: Publish official image to GitHub Container Registry
...
-        with:
-          go-version: "stable"
-      - uses: ko-build/setup-ko@d006021bd0c28d1ce33a07e7943d48b079944c8d # v0.9
-
       - name: Get the version
         id: get_version
         run: |
-          echo "VERSION=${GITHUB_REF#refs/tags/v}" >> "$GITHUB_OUTPUT"
+          if [ "${{ github.event_name }}" = "workflow_dispatch" ]; then
+            echo "VERSION=${{ github.event.inputs.version }}" >> "$GITHUB_OUTPUT"
```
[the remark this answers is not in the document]
AtharvaPakade: Won't be relevant once `workflow_dispatch` is dropped before merge.

.github/workflows/build_release.yml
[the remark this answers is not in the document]
AtharvaPakade: dropped `workflow_dispatch`

Makefile
```diff
@@ -265,32 +267,36 @@ undeploy:
...
 sign-images: ## Sign KEDA images published on GitHub Container Registry
-	cosign sign $(COSIGN_FLAGS) $(IMAGE_OPERATOR_VERSIONED_TAG)
-	cosign sign $(COSIGN_FLAGS) $(IMAGE_OPERATOR_SHA_TAG)
-	cosign sign $(COSIGN_FLAGS) $(IMAGE_INTERCEPTOR_VERSIONED_TAG)
-	cosign sign $(COSIGN_FLAGS) $(IMAGE_INTERCEPTOR_SHA_TAG)
-	cosign sign $(COSIGN_FLAGS) $(IMAGE_SCALER_VERSIONED_TAG)
-	cosign sign $(COSIGN_FLAGS) $(IMAGE_SCALER_SHA_TAG)
+	cosign sign $(COSIGN_FLAGS) $(IMAGE_OPERATOR_REF)
+	cosign sign $(COSIGN_FLAGS) $(IMAGE_INTERCEPTOR_REF)
+	cosign sign $(COSIGN_FLAGS) $(IMAGE_SCALER_REF)
```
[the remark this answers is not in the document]
AtharvaPakade: done

.github/workflows/template-build-sign.yml
```diff
@@ -0,0 +1,122 @@
...
+      scaler-digest:      ${{ steps.publish-scaler.outputs.digest }}
+    env:
+      KO_RELEASE_PLATFORMS: linux/amd64,linux/arm64
+
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+      - uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
+        with:
+          go-version: "stable"
+      - uses: ko-build/setup-ko@d006021bd0c28d1ce33a07e7943d48b079944c8d # v0.9
```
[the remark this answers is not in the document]
AtharvaPakade: done

Makefile
```diff
@@ -265,32 +267,39 @@ undeploy:
...
 
-publish: publish-operator publish-interceptor publish-scaler
-
-release: manifests ## Produce new KEDA Http Add-on release in keda-add-ons-http-$(VERSION).yaml file.
+generate-manifests: ## Generate deployment YAML with pinned digest refs (requires OPERATOR_REF, INTERCEPTOR_REF, SCALER_REF).
 	kustomize build config/crd > keda-add-ons-http-$(VERSION).yaml
 	echo '---' >> keda-add-ons-http-$(VERSION).yaml
-	kustomize build config/operator | KO_DOCKER_REPO=$(IMAGE_OPERATOR) ko resolve --bare --platform=$(KO_RELEASE_PLATFORMS) --tags=$(VERSION) -f - >> keda-add-ons-http-$(VERSION).yaml
+	kustomize build config/operator \
+	  | sed "s|ko://github.com/kedacore/http-add-on/operator|$(OPERATOR_REF)|g" \
```
linkvt, line 283: Is there a reason why we dont use the `IMAGE_OPERATOR_REF` vars and similar we define above?

Makefile
```diff
@@ -17,13 +17,15 @@ GIT_COMMIT       ?= $(shell git rev-list -1 HEAD)
 GIT_COMMIT_SHORT ?= $(shell git rev-parse --short HEAD)
```
linkvt, line 17: This seems to be unused now.

SECURITY.md
```diff
@@ -15,7 +15,95 @@ We have a few preventive measures in place to detect security vulnerabilities:
   - Dependabot will automatically identify vulnerabilities based on the GitHub Advisory Database and open PRs with patches
   - Automated [secret scanning](https://docs.github.com/en/enterprise-cloud@latest/code-security/concepts/secret-security/about-secret-scanning) and alerts
 - Container images are signed with [cosign](https://github.com/sigstore/cosign) using keyless OIDC via GitHub Actions
-- All GitHub Actions are pinned to full commit SHAs to prevent supply chain attacks
+- All GitHub Actions are pinned to full commit SHAs where supported to prevent supply chain attacks; `slsa-framework/slsa-github-generator` reusable workflows are referenced by semver tag per their [official recommendation](https://github.com/slsa-framework/slsa-github-generator/blob/4d014fae4dbd39eb09e8d40348b73db095e6ba9a/README.md#referencing-slsa-builders-and-generators)
```
linkvt, line 18: nit: We could IMO link to main branch instead of a specific commit here, future readers will probably be interested in the future state?

SECURITY.md
```diff
@@ -15,7 +15,95 @@ We have a few preventive measures in place to detect security vulnerabilities:
...
+- All **release** artifacts (container images and deployment manifests) have [SLSA Level 3](https://slsa.dev/spec/v1.0/levels#build-l3) provenance attestations generated by [`slsa-github-generator`](https://github.com/slsa-framework/slsa-github-generator). Canary images (published from `main`) are signed with cosign but intentionally have no SLSA provenance attestation.
```
linkvt, line 19: Are there good reasons not to add provenance to canary images as well? IMO there are no technical blockers?
Big bonus would be that we can verify the attestation before doing an actual release on the canary builds.

SECURITY.md
```diff
@@ -15,7 +15,95 @@ We have a few preventive measures in place to detect security vulnerabilities:
...
+
+## Verifying Release Artifacts
```
linkvt, line 21: This whole section should IMO be added to the [keda-docs](https://github.com/kedacore/keda-docs), this file should mostly be used for the vulnerability related infos. Extending the prevention list above is IMO good.

.github/workflows/build_canary.yml
```diff
@@ -9,45 +9,24 @@ permissions:
...
-        run: cosign version
-
-      - name: Sign KEDA images published on GitHub Container Registry
-        # This step uses the identity token to provision an ephemeral certificate
-        # against the sigstore community Fulcio instance.
-        run: make sign-images
-        env:
-          VERSION: canary
+      id-token: write
+    secrets: inherit
```
linkvt, line 32: Do we still need this after switching from `secrets.GITHUB_TOKEN` to `github.token`?

.github/workflows/template-build-sign.yml
```diff
@@ -0,0 +1,122 @@
...
+    permissions:
+      contents: read
+      packages: write
+      id-token: write   # cosign keyless signing via GitHub OIDC
+    outputs:
+      operator-digest:    ${{ steps.publish-operator.outputs.digest }}
+      interceptor-digest: ${{ steps.publish-interceptor.outputs.digest }}
+      scaler-digest:      ${{ steps.publish-scaler.outputs.digest }}
+    env:
+      KO_RELEASE_PLATFORMS: linux/amd64,linux/arm64
```
linkvt, line 41: Isn't this redundant to what we have in the Makefile?

.github/workflows/build_canary.yml
```diff
@@ -9,45 +9,24 @@ permissions:
...
 jobs:
-  build:
+  get-repo:
     runs-on: ubuntu-latest
+    outputs:
+      image-repo: ${{ steps.get_repo.outputs.IMAGE_REPO }}
+    steps:
+      - name: Get lowercase image repo
+        id: get_repo
+        run: echo "IMAGE_REPO=${GITHUB_REPOSITORY_OWNER,,}" >> "$GITHUB_OUTPUT"
```
linkvt, line 19: Arent we already doing a lower case in the template?

.github/workflows/template-build-sign.yml
```diff
@@ -0,0 +1,122 @@
...
+      # ko build --bare prints exactly one line to stdout: image@sha256:<digest>.
+      # Build progress and errors go to stderr, which remains visible in workflow logs.
+      # make -s suppresses the recipe echo so only ko's output reaches stdout,
+      # keeping the captured ref clean for parsing.
+      - name: Publish operator image
+        id: publish-operator
+        env:
+          IMAGE_REGISTRY: ${{ inputs.image-registry }}
+          IMAGE_REPO_INPUT: ${{ inputs.image-repo }}
+          VERSION: ${{ inputs.version }}
```
linkvt, line 66: We could also move these to the jobs env vars to avoid repeating them for the steps?

.github/workflows/build_release.yml
```diff
@@ -8,75 +8,201 @@ permissions:
...
+  get-digests:
+    needs: [build, get-version]
+    runs-on: ubuntu-latest
+    outputs:
+      operator-digest:    ${{ needs.build.outputs.operator-digest }}
+      interceptor-digest: ${{ needs.build.outputs.interceptor-digest }}
+      scaler-digest:      ${{ needs.build.outputs.scaler-digest }}
+      image-repo:         ${{ needs.get-version.outputs.image-repo }}
+    steps:
+      - run: echo "Forwarding build outputs"
```
linkvt, line 57: This seems like a hack, are we sure this is needed?
Why not something like this?
```yml
# Before (with get-digests):
provenance-operator:
    needs: [get-digests]
    with:
      image: ghcr.io/${{ needs.get-digests.outputs.image-repo }}/http-add-on-operator
      digest: ${{ needs.get-digests.outputs.operator-digest }}

# After (without get-digests):
provenance-operator:
    needs: [build, get-version]
    with:
      image: ghcr.io/${{ needs.get-version.outputs.image-repo }}/http-add-on-operator
      digest: ${{ needs.build.outputs.operator-digest }}
```

.github/workflows/build_release.yml
```diff
@@ -8,75 +8,201 @@ permissions:
   contents: read
 
 jobs:
-  build:
+  # Resolve the version string and lowercase registry namespace once so all
+  # downstream jobs share the same values.
+  get-version:
```
linkvt, line 13: nit: Name is a bit misleading asi it also resolves the registry

linkvt (changes requested):
Quite a big change especially compared to the level 2 approach, guess this can't be done in a simpler way though so thanks at first!

Something I learned signing is always based on the actual digest behind a tag/version/..., so the previous signing for version AND digest was redundant.

I hope my comments make sense, otherwise let me know.

Makefile
AtharvaPakade: `IMAGE_OPERATOR_REF` forces to use the image with digest. For local generations we usually use tags. 

But given that we already have a seperate rule `deploy`,  and the `generate-manifests` will alway be used by workflows; we can go ahead use `IMAGE_OPERATOR_REF` and similar refs.

Makefile
AtharvaPakade: I will drop this.

SECURITY.md
AtharvaPakade: I though it will be better to point out a exact ref doc that use for implementation. Additionaly IMO is better to tag a particular version because sometimes the changes recommended by the doc might not be aligned with ours. Let me know if that makes sense to you.

SECURITY.md
AtharvaPakade: Yes, there is no technical blocker following are few points i want to point out. 
- Verification become difficult as usually the verification happens through a tag (this can be dropped and we can just verify it belong to a github project ). But since we will be using it for testing this won't be problem. 
- This add ~2-4min to your current canary job. 
- `.att` piles up in the registery as for every merge to main will generate a `.att`.
- we will only be having `container provenance` for the canary release. As there is no place to push the signed manifest. I read somewhere we can also have canary tag (which keeps getting updated) and we can cut releases around it.

SECURITY.md
AtharvaPakade: Sure, i will drop this section and open a separate pr in the keda-docs repo.

.github/workflows/build_canary.yml
AtharvaPakade: Yes, there is a lot of redundancy around this, i will streamline it.

.github/workflows/template-build-sign.yml
AtharvaPakade: In a way true, But IMO it is always better to be expilict in the workflows. My only concern is everytime we make changes to the makefile, changes to the workflow will also be needed. Let me know what you think.

.github/workflows/template-build-sign.yml
AtharvaPakade: My idea behind separating them was to make the env explicitly visible so that we know what each step requires.

.github/workflows/build_release.yml
AtharvaPakade: Yes, this was just a hack to avoid lint warning on each step.

.github/workflows/template-build-sign.yml
linkvt: I understand but having them here 4 times seems quite a bit redundant

.github/workflows/template-build-sign.yml
linkvt: Which kind of negative effects would you see with using the env var from the makefile? I rather see that we might forget to add a new platform here as it is duplicated.

PR #1663, opened 2026-06-02
feat: add s390x to release platforms

### LINKED ISSUE #1662: Enable support for s390x
Keda http-add-on requires support to run on the s390x architecture

linkvt:
Simply cross building for s390x is IMO good enough as well as we don't expect that our quite normal go code suddenly would not work on that platform anymore.
We test amd64 and arm64 in CI already, it is extremely unlikely that s390x would not work considering we dont use any unsafe/C code or anything archtiecture specific.

### Changes:
- Add linux/s390x to KO_RELEASE_PLATFORMS in Makefile

### Checklist

  - see https://github.com/kedacore/keda-docs/pull/1783

Fixes #1662

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ This changelog keeps track of work items that have been completed and are ready 
 
 ### New
 
+- **General**: Add s390x to supported release platforms ([#1662](https://github.com/kedacore/http-add-on/issues/1662))
 - **General**: Add `coldStart.placeholder` support for returning static HTTP responses during scale-from-zero ([#874](https://github.com/kedacore/http-add-on/issues/874))
 - **General**: TODO ([#TODO](https://github.com/kedacore/http-add-on/issues/TODO))
 - **Scaler**: Add OpenTelemetry metrics and distributed tracing to the external scaler ([#965](https://github.com/kedacore/http-add-on/issues/965))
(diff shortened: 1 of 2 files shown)


Fedosin (approved):

JorTurFer:
We have s390x machines, if you want to run e2e test, it's feasible

linkvt:
Thanks for the offer, but at least for me the effort of having custom runners isn't worth it right now. Also considering s390x doesn't have many users and that we're doing nothing magical in the HTTP Addon I'm personally trusting Gos cross compilation.

JorTurFer:
> Thanks for the offer, but at least for me the effort of having custom runners isn't worth it right now. Also considering s390x doesn't have many users and that we're doing nothing magical in the HTTP Addon I'm personally trusting Gos cross compilation.

Fine for me, but we should be clear in public docs about the expected support in this regard top avoid misunderstandings :)

linkvt:
Good news, we're doing exactly that 🙂 

>   - Supported architectures: `amd64`, `arm64`, or `s390x` (CI-tested on `amd64` and `arm64`)
> 
> https://github.com/kedacore/keda-docs/blob/7c8c7a707f9cb00b8770db3608cc25eef6668c43/content/http-add-on/0.15/operations/installation.md?plain=1#L12

JorTurFer:
sorry, I didn't see it 😓

PR #1670, opened 2026-06-04
fix: tolerate partial interceptor pod failures in scaler

### LINKED ISSUE #1667: Scaler restarts on single unreachable interceptor endpoint
### Report

This issue was created based on the PR #903 from January 2024 which I'll close as the suggested solution isn't quite good.
**I used AI to analyze the related PR and related code with additional manual checks, I didn't reproduce it yet though besides the sample test case below.**

When any single interceptor pod endpoint is temporarily unreachable (e.g., spot node eviction, rolling update race window before EndpointSlice updates), `fetchCountsPerPod` fails the entire metrics fetch because it uses `errgroup`, which returns on the first error. This sets `PingerERROR`, the gRPC health check reports `NOT_SERVING`, and Kubernetes restarts the scaler pod after ~30s when the pod isn't reachable for that long.

The restart destroys all in-memory state (`prevPodCounts`, `rateBuckets`), meaning RPS-based scaling produces zero/incorrect values for up to a full window duration (default: 1 minute) after each restart.

A partial failure (one pod unreachable) is treated identically to a total failure (all pods unreachable). I'm not sure if this is what we want.

### Expected Behavior

Unclear, should be proposed as part of that issue.

Underreporting by ignoring answers from 1+ interceptors might cause a scale down, so this might not be a good approach as well.

### Actual Behavior

One unreachable interceptor pod causes a full scaler restart, losing all rate calculation state and disrupting scaling for all workloads.

1. Deploy the HTTP add-on with multiple interceptor replicas
2. Evict or kill a node running one interceptor pod (or simulate by blocking network to one pod)
3. Observe the scaler logs: `"fetching all counts failed"` followed by health check reporting `NOT_SERVING`
4. After ~30s the scaler pod is restarted by Kubernetes liveness probe

Example test for `scaler/queue_pinger_test.go`:
```go
func TestFetchCountsPerPod_FailsEntirelyWhenOnePodUnreachable(t *testing.T) {
	r := require.New(t)
	ctx := t.Context()
	const adminPort = "8081"

	// Pod A: healthy, returns valid counts.
	podA := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
		_ = json.NewEncoder(w).Encode(queue.Counts{
			"host1": {Concurrency: 5, RequestCount: 100},
		})
	}))
	defer podA.Close()

	// Pod B: unreachable (closed listener, simulates spot node eviction).
	podBListener, err := net.Listen("tcp", "127.0.0.1:0")
	r.NoError(err)
	podBAddr := podBListener.Addr().String()
	podBListener.Close()

	withPatchedDefaultTransport(t, map[string]string{
		"pod-a:" + adminPort: podA.Listener.Addr().String(),
		"pod-b:" + adminPort: podBAddr,
	})

	endpointsFn := func(context.Context, string, string) (k8s.Endpoints, error) {
		return k8s.Endpoints{
			ReadyAddresses: []string{"pod-a", "pod-b"},
		}, nil
	}

	_, err = fetchCountsPerPod(ctx, logr.Discard(), endpointsFn, "testns", "testsvc", adminPort)

	r.NoError(err, "one unreachable pod should not fail the entire fetch")
}

```

_No response_

### HTTP Add-on Version

0.14.0

### Kubernetes Version

all

### Platform

Any

None

### Anything else?

_No response_

Fedosin:
When any single interceptor pod endpoint is temporarily unreachable
(e.g. spot node eviction, rolling update race window), `fetchCountsPerPod`
previously failed the entire metrics fetch because it used `errgroup`,
which returns on the first error. This set `PingerERROR`, the gRPC health
check reported `NOT_SERVING`, and Kubernetes restarted the scaler pod
after ~30s -- destroying all in-memory state (`prevPodCounts`,
`rateBuckets`) and producing zero/incorrect RPS values for up to a full
window duration (default: 1 minute).

This PR makes two changes to address the problem:

### 1. Tolerate partial interceptor pod failures

Replace `errgroup` with `sync.WaitGroup.Go()` (available since Go 1.25)
and buffered channels so that individual pod failures are logged and
skipped. An error is only returned when **all** pods are unreachable.

### 2. Cache last-known counts for unreachable pods

To prevent underreporting (which could cause premature scale-down or
prevent scale-from-zero), the scaler now caches the last successful
response from each interceptor pod. When a pod is unreachable but still
present in the EndpointSlice, its cached concurrency is included in the
aggregated metrics. The cache is pruned when the pod actually leaves
the EndpointSlice.

A new OTel gauge `scaler.pinger.unreachable_pods` reports the number of
interceptor pods that were unreachable in the last fetch cycle, giving
operators visibility into partial degradation before it becomes a total
outage.

### Checklist

- [x] Changelog is updated per our changelog requirements
- [x] Documentation is updated if needed (README.md, keda-docs)

Fixes #1667

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@ This changelog keeps track of work items that have been completed and are ready 
 - **General**: TODO ([#TODO](https://github.com/kedacore/http-add-on/issues/TODO))
 - **Interceptor**: Fix panic with "invalid concurrent Body.Read call" when reverse proxy fails without consuming the request body ([#1668](https://github.com/kedacore/http-add-on/issues/1668))
 - **Scaler**: Fix incorrect HPA values when both concurrency and requestRate metrics are configured ([#1659](https://github.com/kedacore/http-add-on/issues/1659))
+- **Scaler**: Tolerate partial interceptor pod failures instead of restarting the scaler on a single unreachable endpoint; cache last-known concurrency for unreachable pods to prevent underreporting ([#1667](https://github.com/kedacore/http-add-on/issues/1667))
 
 ### Deprecations
 
--- a/scaler/metrics/instruments.go
+++ b/scaler/metrics/instruments.go
@@ -15,16 +15,18 @@ const (
 	// ServiceName is the OTEL service.name used for both metrics and tracing.
 	ServiceName = "keda-http-external-scaler"
 
-	MetricPingerFetchDuration = "scaler.pinger.fetch.duration"
-	MetricPingerFetchErrors   = "scaler.pinger.fetch.errors"
-	MetricPingerEndpoints     = "scaler.pinger.endpoints"
+	MetricPingerFetchDuration   = "scaler.pinger.fetch.duration"
+	MetricPingerFetchErrors     = "scaler.pinger.fetch.errors"
+	MetricPingerUnreachablePods = "scaler.pinger.unreachable_pods"
+	MetricPingerEndpoints       = "scaler.pinger.endpoints"
 )
 
 // Instruments holds all metric instruments for the external scaler.
 type Instruments struct {
-	pingerFetchDuration api.Float64Histogram
-	pingerFetchErrors   api.Int64Counter
-	pingerEndpoints     api.Int64Gauge
+	pingerFetchDuration   api.Float64Histogram
+	pingerFetchErrors     api.Int64Counter
+	pingerUnreachablePods api.Int64Gauge
+	pingerEndpoints       api.Int64Gauge
 }
 
 // NewNoopInstruments returns Instruments backed by a no-op provider, for use in tests.
@@ -60,6 +62,14 @@ func NewInstruments(provider *sdkmetric.MeterProvider) (*Instruments, error) {
 		return nil, fmt.Errorf("creating pinger fetch errors counter: %w", err)
 	}
 
+	pingerUnreachablePods, err := meter.Int64Gauge(
+		MetricPingerUnreachablePods,
+		api.WithDescription("Number of interceptor pods that were unreachable in the last fetch cycle"),
+	)
+	if err != nil {
+		return nil, fmt.Errorf("creating pinger unreachable pods gauge: %w", err)
+	}
+
 	pingerEndpoints, err := meter.Int64Gauge(
 		MetricPingerEndpoints,
 		api.WithDescription("Number of interceptor endpoints the scaler is polling"),
@@ -69,16 +79,18 @@ func NewInstruments(provider *sdkmetric.MeterProvider) (*Instruments, error) {
 	}
 
 	return &Instruments{
-		pingerFetchDuration: pingerFetchDuration,
-		pingerFetchErrors:   pingerFetchErrors,
-		pingerEndpoints:     pingerEndpoints,
+		pingerFetchDuration:   pingerFetchDuration,
+		pingerFetchErrors:     pingerFetchErrors,
+		pingerUnreachablePods: pingerUnreachablePods,
+		pingerEndpoints:       pingerEndpoints,
 	}, nil
 }
 
 // RecordFetch records a completed pinger fetch cycle.
-func (i *Instruments) RecordFetch(duration time.Duration, endpointCount int, fetchErr error) {
+func (i *Instruments) RecordFetch(duration time.Duration, endpointCount, failedPods int, fetchErr error) {
 	i.pingerFetchDuration.Record(context.Background(), duration.Seconds())
 	i.pingerEndpoints.Record(context.Background(), int64(endpointCount))
+	i.pingerUnreachablePods.Record(context.Background(), int64(failedPods))
 	if fetchErr != nil {
 		i.pingerFetchErrors.Add(context.Background(), 1)
 	}
--- a/scaler/queue_pinger.go
+++ b/scaler/queue_pinger.go
@@ -9,10 +9,10 @@ import (
 	"net/http"
 	"net/url"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/go-logr/logr"
-	"golang.org/x/sync/errgroup"
 
 	"github.com/kedacore/http-add-on/pkg/k8s"
 	"github.com/kedacore/http-add-on/pkg/queue"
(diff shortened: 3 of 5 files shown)


linkvt:
I wrote the following in the issue:
> Underreporting by ignoring answers from 1+ interceptors might cause a scale down, so this might not be a good approach as well.

This PR seems to do exactly that, was this done on purpose or did I miss anything?

interceptor/admin.go
```diff
@@ -5,21 +5,26 @@ import (
...
 
-	queue.AddCountsRoute(
-		logger,
-		mux,
-		counter,
-	)
+	mux.Handle("/readyz", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if err := pingTracker.HealthCheck(r.Context()); err != nil {
+			http.Error(w, err.Error(), http.StatusServiceUnavailable)
+			return
```
linkvt: Before, when a single interceptor wasn't reachable scaling didn't work anymore but requests were still served -> not an issue unless traffic spiked.
Not sure if I'm missing something but isn't this now worse as we effectively stop all traffic going through all interceptors when a scaler isn't working anymore?

Fedosin: Yeah, it's a bit controversial. We can discuss whether we need this feature, but it's probably better to implement it in a separate PR.

scaler/metrics/instruments.go
```diff
@@ -17,13 +17,15 @@ const (
 
 	MetricPingerFetchDuration = "scaler.pinger.fetch.duration"
 	MetricPingerFetchErrors   = "scaler.pinger.fetch.errors"
+	MetricPingerPodErrors     = "scaler.pinger.pod.errors"
```
linkvt: Im wondering if we can choose a better metric name here as `pod.errors` doesn't really say which pod is problematic.

scaler/queue_pinger.go
```diff
@@ -37,6 +38,13 @@ type aggregatedCount struct {
 	RequestRate float64
 }
 
+// cachedPodEntry pairs a pod's last-known counts with the wall-clock
+// time they were fetched, so stale entries can be expired via cacheTTL.
+type cachedPodEntry struct {
+	counts    queue.Counts
+	fetchedAt time.Time
```
linkvt: Is there a reason why we want to expire the cached records? Does it improve a specific case or is it potentially more harmful than not expiring until the endpoint slices are updated again?

scaler/queue_pinger.go
```diff
@@ -204,14 +225,44 @@ func (q *queuePinger) fetchAndSaveCounts(ctx context.Context) error {
...
+	// Only concurrency is replayed — request rate deltas are not
+	// synthesised because we have no way to know what traffic the
+	// unreachable pod handled. This is a deliberate trade-off:
+	// concurrency is a point-in-time gauge (stale > zero), while
+	// rate is a derivative that would be misleading if fabricated.
+	for _, epKey := range result.endpointKeys {
+		if _, reachable := perPod[epKey]; reachable {
+			continue
+		}
+		if entry, ok := q.cachedPodCounts[epKey]; ok && now.Sub(entry.fetchedAt) <= q.cacheTTL {
```
linkvt: Should we also log a warning or similar in when we're using cached counts? Might be helpful to know when cached data is returned.

linkvt (commented):
Good idea with caching the responses, only left a few minor comments.

scaler/metrics/instruments.go
Fedosin: Renamed to `scaler.pinger.fetch.pod_failures`. This is an aggregate counter (not per-pod), so it tells you "how many individual pod fetches failed" rather than identifying a specific pod. The per-pod details are already in the error log line which includes `interceptorAddress`.

scaler/queue_pinger.go
Fedosin: Good point. You are right, the TTL is more harmful than helpful here. If the pod is still in the EndpointSlice, we should use the cached data regardless of age. If it leaves the EndpointSlice, the cache is pruned anyway.

Removed the TTL entirely in the latest push. The cache now lives exactly as long as the pod is present in the EndpointSlice.

scaler/queue_pinger.go
Fedosin: Added a V(1) debug-level log line that fires for each unreachable pod whose cached counts are used. It includes the pod key and the number of cached hosts. This way operators can see it when they bump the log level, without adding noise at the default level.

scaler/metrics/instruments.go
Fedosin: Actually changed it to `scaler.pinger.unreachable_pods` of gauge type. Now it records the number of unreachable pods each fetch cycle. It naturally goes back to 0 when all pods recover, which makes it directly usable for alerting (e.g. "alert if scaler_pinger_unreachable_pods > 0 for 5 minutes").

CHANGELOG.md
```diff
@@ -45,14 +44,13 @@ This changelog keeps track of work items that have been completed and are ready
 ### Fixes
 
 - **General**: TODO ([#TODO](https://github.com/kedacore/http-add-on/issues/TODO))
-- **Interceptor**: Fix graceful shutdown so in-flight requests are drained before the process exits. The interceptor now waits for active handlers to complete (bounded by `KEDA_HTTP_DRAIN_TIMEOUT`), marks the readiness probe unhealthy immediately on SIGTERM, and delays listener closure by `KEDA_HTTP_SHUTDOWN_DELAY` to let Kubernetes propagate endpoint removal. ([#1636](https://github.com/kedacore/http-add-on/issues/1636))
 - **Interceptor**: Fix panic with "invalid concurrent Body.Read call" when reverse proxy fails without consuming the request body ([#1668](https://github.com/kedacore/http-add-on/issues/1668))
 - **Scaler**: Fix incorrect HPA values when both concurrency and requestRate metrics are configured ([#1659](https://github.com/kedacore/http-add-on/issues/1659))
+- **Scaler**: Tolerate partial interceptor pod failures instead of restarting the scaler on a single unreachable endpoint; cache last-known counts for unreachable pods to prevent underreporting ([#1667](https://github.com/kedacore/http-add-on/issues/1667))
```
linkvt: nit: we only cache the concurrency not the full counts the scaler receives.

linkvt (approved):

PR #1709, opened 2026-07-15
refactor: align scaler main.go with interceptor main.go pattern

linkvt:
Align scaler/main.go with interceptor/main.go error-handling pattern to avoid using the more complex runtime.Goexit+defer os.Exit instead of just returning the error and exiting based on that.

### Changes
- Extract run() error from main(), returning wrapped errors instead of calling runtime.Goexit()
- Start the pprof server via kedahttp.ServeContext so it shuts down with the rest of the app instead of leaking past context cancel
- Use util.IsIgnoredErr consistently to filter expected shutdown errors (context.Canceled, http.ErrServerClosed)

### Checklist

### DIFF
--- a/scaler/main.go
+++ b/scaler/main.go
@@ -6,14 +6,12 @@ package main
 
 import (
 	"context"
-	"errors"
 	"flag"
 	"fmt"
 	"net"
 	"net/http"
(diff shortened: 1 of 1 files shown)


Fedosin (approved):
LGTM

PR #1729, opened 2026-07-28
fix(scaler): bound queue counts requests to interceptor pods

### LINKED ISSUE #1730: Scaler crash-loops when an interceptor pod stops responding on /queue
### Report

The scaler fetches queue counts from every interceptor pod with `http.DefaultClient` and no request context, so nothing bounds the call. An interceptor that accepts the connection but never answers blocks the fetch goroutine forever, stalling the polling loop.

### Expected Behavior

A single unresponsive interceptor pod should be treated as unreachable for that tick, as already happens for pods that refuse connections (#1667). The scaler should keep serving metrics from the remaining pods.

### Actual Behavior

`wg.Wait()` in `fetchCountsPerPod` never returns, so `lastPingTime` stops advancing. The health loop flips **liveness** to `NOT_SERVING` after two ticks and kubelet restarts the scaler. The pod is still wedged, so this repeats. Since the scaler is a cluster-wide singleton, every HTTP-autoscaled app loses its metric source while it crash-loops.

Note that #1667 only covers pods that are unreachable — a refused connection returns immediately, a black hole never does — so that recovery path is never reached.

1. Deploy the add-on with 2+ interceptor replicas and any `InterceptorRoute`.
2. Make one interceptor accept TCP on the admin port but never respond on `/queue` (e.g. `SIGSTOP` the process, or blackhole its responses with a netfilter rule).
3. Watch the scaler: `lastPingTime` stops advancing and the pod begins restarting on liveness failures.

### HTTP Add-on Version

main (post-0.15.0)

### Kubernetes Version

Any

### Platform

Any

Yes

### Anything else?

Relevant code: `queue.GetCounts` in `pkg/queue/queue_rpc.go`, the fan-out in `scaler/queue_pinger.go`, and the health loop in `scaler/main.go`.

Fedosin:
The scaler polls every interceptor pod's `/queue` endpoint on each tick to build the concurrency and request-rate metrics it serves to KEDA. Those requests were issued with `http.DefaultClient`, whose `Timeout` is zero, and no context was attached. `http.DefaultTransport` bounds connection setup (30s dial, 10s TLS handshake) but sets no `ResponseHeaderTimeout`, so once a connection was established nothing bounded the exchange.

An interceptor pod that accepts the connection but never answers — GC pause, deadlock, wedged informer, a node in a degraded network state — therefore parks its fetch goroutine indefinitely. `wg.Wait()` in `fetchCountsPerPod` never returns, the tick never completes, and `lastPingTime` stops advancing. The health loop in `scaler/main.go` then flips **liveness** to `NOT_SERVING` after two ticks and kubelet restarts the scaler. The wedged pod is still wedged, so it repeats. Since the scaler is a cluster-wide singleton, every HTTP-autoscaled app in the cluster loses its metric source while it crash-loops.

#1667 made the scaler tolerate interceptor pods that are unreachable, but a refused connection returns immediately while a black hole never does, so that recovery path was never reached.

### Changes

- `queue.GetCounts` now takes a `context.Context` and uses `http.NewRequestWithContext`, so cancellation propagates and the request can be bounded.
- `queuePinger` uses a dedicated `*http.Client` with an explicit `Timeout`, which covers the whole exchange including the response body read. An unresponsive pod is now treated exactly like an unreachable one: it counts as failed for that tick, and the existing cached-counts path keeps reporting its last known concurrency.
- The bound is `KEDA_HTTP_QUEUE_TICK_DURATION`, clamped up to a 250ms floor. No new setting is introduced: there is no point waiting longer than the interval after which we ask again, since the next tick supersedes the answer. The floor matters because request-rate deltas are deliberately not synthesised for pods we failed to reach, so letting a very short tick mark slow-but-healthy pods unreachable would quietly under-report the rate.
- Non-200 responses are rejected explicitly instead of surfacing as a confusing JSON decode error, and the body is capped before decoding.

The client's `Transport` is deliberately left `nil` so it resolves `http.DefaultTransport` per request, keeping the connection pool shared across ticks.

### Tests

Seven new test functions (ten cases with the table subtests):

- `pkg/queue` pins the RPC contract: an unresponsive interceptor, a cancelled context, and a non-200 response.
- `scaler` covers one unresponsive pod among healthy ones, all pods unresponsive, and the timeout derivation including its clamp at zero, negative, and very short tick durations.
- `TestFetchAndSaveCounts_UnresponsivePodKeepsPingerFresh` is the one that matters most: it asserts `lastPingTime` keeps advancing across consecutive ticks while a pod hangs, which is exactly what the health check depends on and exactly what broke.

The unresponsive-pod tests fail against the current behavior — with the unbounded client they hang and only fail once the Go test timeout expires. With the fix they pass in under a second.

### Follow-up

While tracing this I noticed the health loop is a related hazard worth its own issue: it ties **liveness** rather than readiness to a dependency being reachable, and its staleness threshold of `2 × KEDA_HTTP_QUEUE_TICK_DURATION` gets very tight at small tick values, where a healthy fan-out across many pods can exceed it. A restart cannot fix an unreachable interceptor, and it discards the rate history in `prevPodCounts`/`rateBuckets`. Out of scope here.

### Checklist

No new configuration is introduced, but `KEDA_HTTP_QUEUE_TICK_DURATION` gains a second effect, so its keda-docs entry deserves a sentence noting it also bounds each request to an interceptor. Happy to open that PR.

Fixes #1730

### DIFF
--- a/scaler/queue_pinger.go
+++ b/scaler/queue_pinger.go
@@ -95,9 +95,7 @@ func newQueuePinger(lggr logr.Logger, getEndpointsFn k8s.GetEndpointsFunc, ns, s
 		interceptorSvcName:     svcName,
 		interceptorServiceName: deplName,
 		adminPort:              adminPort,
-		// Transport is deliberately left nil so the client resolves
-		// http.DefaultTransport per request, keeping its connection pool
-		// shared across ticks.
+		// Leave Transport nil to preserve the existing http.DefaultTransport behavior.
 		httpCl:          &http.Client{Timeout: fetchTimeout},
 		lggr:            lggr,
 		instruments:     instruments,
--- a/scaler/queue_pinger.go
+++ b/scaler/queue_pinger.go
@@ -28,6 +28,13 @@ const (
 
 	defaultWindow      = time.Minute
 	defaultGranularity = time.Second
+
+	// minFetchTimeout is the lower clamp on a single queue counts request.
+	// The timeout tracks the tick duration, but a very short tick must not
+	// make healthy pods on a loaded node look unreachable: that would drop
+	// their request-rate deltas, which are never synthesised for pods we
+	// could not reach, and quietly under-report the rate.
+	minFetchTimeout = 250 * time.Millisecond
 )
 
 // aggregatedCount holds the scaler's computed view of a single
@@ -46,7 +53,7 @@ type aggregatedCount struct {
 //
 // Sample usage:
 //
-//	pinger := newQueuePinger(lggr, getEndpointsFn, ns, svcName, deplName, adminPort, instruments)
+//	pinger := newQueuePinger(lggr, getEndpointsFn, ns, svcName, deplName, adminPort, fetchTimeout, instruments)
 //	go pinger.start(ctx, ticker)
 type queuePinger struct {
 	getEndpointsFn         k8s.GetEndpointsFunc
@@ -54,6 +61,7 @@ type queuePinger struct {
 	interceptorSvcName     string
 	interceptorServiceName string
 	adminPort              string
+	httpCl                 *http.Client
 	pingMut                sync.RWMutex
 	lastPingTime           time.Time
 	allCounts              map[string]aggregatedCount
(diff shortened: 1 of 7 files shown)


scaler/queue_pinger.go
```diff
@@ -76,19 +84,27 @@ type queuePinger struct {
...
-		lggr:                   lggr,
-		instruments:            instruments,
-		allCounts:              map[string]aggregatedCount{},
-		prevPodCounts:          map[string]map[string]int64{},
-		cachedPodCounts:        map[string]queue.Counts{},
-		rateBuckets:            map[string]*queue.RequestsBuckets{},
+		// Transport is deliberately left nil so the client resolves
+		// http.DefaultTransport per request, keeping its connection pool
+		// shared across ticks.
+		httpCl:          &http.Client{Timeout: fetchTimeout},
```
linkvt, line 99: nit: the comment is a bit misleading as the connection pool of a transport would always be shared, even if we're explicit, right? All ticks use this http client and would also use the implicit or explicit transport.

linkvt (approved):
LGTM, left a nit comment and we should probably also add a small note to the docs about the timeout behavior.

PR #1731, opened 2026-07-28
feat: add custom CA trust for outbound backend TLS

### LINKED ISSUE #1728: Support custom CA trust for interceptor backend connections
## Proposal

The interceptor has no way to trust additional/custom CA certificates (e.g. a corporate CA or an internal cluster CA) when it dials backend Kubernetes Services over HTTPS. Only the system CA pool is used, so backends whose certs are signed by a private CA can't be verified — the only workaround today is disabling certificate verification entirely.

## Use-Case

Environments that terminate backend TLS with certs issued by an internal/corporate CA (rather than a public CA) need the interceptor to trust that CA specifically, without turning off verification altogether.

## Is this a feature you are interested in implementing yourself?

Yes

## Anything else?

Existing related mechanism: `KEDA_HTTP_PROXY_TLS_CERT_STORE_PATHS` already appends its loaded certs to the trust pool, but only for cert+key pairs — a CA-only file with no matching private key fails to load at all, so there's no way to add a trust-only CA without also supplying a key it doesn't need.

linkvt:
Backend connections previously had no way to trust custom CAs, forcing `KEDA_HTTP_TLS_SKIP_VERIFY` as the only workaround for private PKI. This also separates outbound (interceptor-to-backend) TLS config from serving (client-to-interceptor) TLS config, which were previously conflated in a single tls.Config shared between both directions.

The chosen approach of accepting CA dirs is based on how [KEDA accepts the configuration](https://github.com/kedacore/keda/blob/8360d22a2f294107a95e4af528d4a343b4332cd6/cmd/operator/main.go#L359) (e.g. from the OLM operator).

I decided to move the TLS config to a shared package out of the interceptor package which is causing a few moves here but makes it easier to apply the same TLS base policy to the TLS config we use e.g. for the scaler.

### Changes

- Add KEDA_HTTP_TLS_CA_DIRS to load custom CA certs for outbound backend connections, independent of whether serving TLS is enabled
- Split ProxyHandlerConfig's TLS into ServingTLS/OutboundTLS
- Apply TLS version, cipher suite, and curve policy to outbound connections (previously only applied to serving)
- Create pkg/tls as shared TLS policy package with Policy.NewConfig() factory for both directions
- Fail fast on misconfigured CA dirs (unreadable path, non-PEM content, or zero certs loaded) rather than silently degrading
- Update e2e TLS tests to use a real CA bundle instead of KEDA_HTTP_TLS_SKIP_VERIFY

### Checklist

	- https://github.com/kedacore/keda-docs/pull/1830
	- https://github.com/kedacore/charts/pull/891

Fixes #1728


pkg/tls/policy.go
```diff
@@ -0,0 +1,137 @@
...
+	return env.Must(env.ParseAs[Policy]())
+}
+
+// NewConfig returns a tls.Config with version, cipher suite, and curve
+// preference fields set according to the policy. Fields left empty in the
+// policy preserve Go's defaults.
+func (p Policy) NewConfig() (*tls.Config, error) {
+	cfg := &tls.Config{
+		InsecureSkipVerify: p.SkipVerify, //nolint:gosec // G402: user-configurable
+	}
```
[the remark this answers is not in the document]
linkvt: 1. we are not exposing the interceptor to the internet, there is usually an ingress before it
2. tls 1.2 is good enough, it is overwritable

/ar not exposed to the internet and 1.2 is good enough and overwritable

pkg/tls/ca.go
```diff
@@ -0,0 +1,65 @@
...
+func BuildCAPool(caDirs string) (*x509.CertPool, error) {
+	pool, err := x509.SystemCertPool()
+	if err != nil {
+		return nil, fmt.Errorf("loading system CA pool: %w", err)
+	}
+	if err := appendCADirs(pool, caDirs); err != nil {
+		return nil, err
+	}
+	return pool, nil
+}
```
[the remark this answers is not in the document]
linkvt: When systemcertpool doesn't work something is seriously broken.

interceptor/main.go
```diff
@@ -236,20 +238,26 @@ func run() error {
 
 	// --- Proxy goroutines (use proxyCtx) ---
```
Fedosin, line 240:
```suggestion
	if tlsPolicyCfg.CADirs != "" && !servingCfg.ProxyTLSEnabled {
		setupLog.Info("WARNING: KEDA_HTTP_TLS_CA_DIRS is set but KEDA_HTTP_PROXY_TLS_ENABLED is false; " +
			"backend connections are plaintext HTTP and the CA bundle is ignored")
	}
```

interceptor/main.go
```diff
@@ -269,7 +277,7 @@ func run() error {
 
 	proxyEg.Go(func() error {
 		setupLog.Info("starting the proxy server", "port", servingCfg.ProxyPort)
-		if err := runProxyServer(proxyCtx, ctrl.Log, queues, readyCache, routingTable, ctrlCache, timeoutCfg, servingCfg, servingCfg.ProxyPort, nil, tracingCfg, instruments, &draining); !util.IsIgnoredErr(err) {
+		if err := runProxyServer(proxyCtx, ctrl.Log, queues, readyCache, routingTable, ctrlCache, timeoutCfg, servingCfg, servingCfg.ProxyPort, nil, outboundTLS, tracingCfg, instruments, &draining); !util.IsIgnoredErr(err) {
```
Fedosin, line 280: Why do we pass `outboundTLS` to this server? As far as I can tell it can never be
used here: this listener gets `nil` serving TLS, so the routing middleware resolves
upstreams as `http://`, and `net/http` only calls `DialTLSContext` for `https` URLs
(`transport.go`: `if cm.scheme() == "https" && t.hasCustomTLSDialer()`). So we build
a transport with a TLS dialer that never fires.

I think `nil` reads better here, with a one-line comment saying why — as written it
looks like outbound TLS is active on the plain listener, which is what tripped me up
on first read.

Unless the intent is forward-compatibility for when the upstream scheme stops being
derived from serving TLS? If so that's fine by me.

linkvt: Yes exactly, this is also why I separated the TLS configs.
IMO there is no reason to derive the outbound request scheme from the inbound one so my rough idea (which isn't written down yet) is that we add a `scheme`/`protocol` field the determines the protocol we use for the outbound connections.

interceptor/main.go
linkvt: IMO preexisting as this concerns also all other TLS related vars, while this one is only correct until we allow cross protocol (HTTPS->HTTP and vice versa) routing.

Fedosin (approved):

PR #1733, opened 2026-07-30
feat(interceptor): bound the cold-start hold queue

### LINKED ISSUE #1732: Cold-start hold queue is unbounded (risk of interceptor OOM)
### Report
There is currently no limit on the interceptor's cold-start queue: requests held while an application scales from zero accumulate without any bound. This can potentially lead to an overflow and an OOM kill of the interceptor, taking down routing for every other application routed through it.

### Expected Behavior
The cold-start queue is bounded, so a burst of requests against a scaled-to-zero application cannot exhaust the interceptor's memory.

### Actual Behavior
Requests held during a cold start accumulate without limit. Each held request keeps a goroutine, a client connection, and its buffers.

1. Deploy an application that scales to zero
2. Send a large burst of concurrent requests to it
3. Held requests pile up in the interceptor with no cap, growing memory usage until the pod can be OOM-killed

Yes

### Anything else?
#1038 covers limiting requests released to *warm* replicas; this issue is specifically about the unbounded hold queue during scale-from-zero.

Fedosin:
Requests for a scaled-to-zero app park in the interceptor with no cap and, at default timeouts, no deadline. A spike against one cold app can exhaust the shared interceptor's memory and file descriptors, taking down routing for every other tenant.

This adds per-route admission control in a dedicated `ColdStart` middleware, enforced only while the target has no ready endpoints, so warm-path traffic is unaffected:

- `coldStart.maxPendingRequests` limits held requests per route, falling back to the new `KEDA_HTTP_COLD_START_MAX_PENDING_REQUESTS` global default (`0` = unlimited, preserving current behavior on upgrade).
- The limit applies **per interceptor replica** (each pod enforces against its in-memory counter), so effective cluster-wide capacity is `maxPendingRequests` × interceptor replicas. This matches the threat model: the limit sizes each pod's memory/FD budget, so protection holds regardless of replica count.
- `coldStart.overflow` (`Reject` | `Placeholder`) selects the overflow behavior: `Reject` returns `503`; `Placeholder` serves the route's placeholder response.
- A route with a placeholder, `overflow: Placeholder`, and an explicit `maxPendingRequests` now holds requests up to the limit and serves the placeholder only to overflow — previously every request was short-circuited immediately. All other placeholder configs keep their current behavior.
- Rejections are tracked by the new `interceptor.cold_start.rejections` counter.

The `ColdStart` middleware absorbs the former `Placeholder` middleware so all cold-start handling (immediate placeholder, bounded holding, overflow response) lives in one place. It runs after the `Counting` middleware, so held and placeholder-served requests are still counted — that count is what drives scale-from-zero. `Counting` itself is unchanged.

Companion PRs:

- Docs: kedacore/keda-docs#1835
- Helm chart (`interceptor.coldStartMaxPendingRequests` value): kedacore/charts#896

### Checklist

- [x] Documentation is updated if needed ([`README.md`](../README.md), [keda-docs](https://github.com/kedacore/keda-docs/tree/main/content/http-add-on)) — kedacore/keda-docs#1835

Fixes #1732


interceptor/middleware/counting.go
```diff
@@ -35,6 +56,18 @@ func (cm *Counting) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		util.LoggerFromContext(ctx).V(1).Info("cold-start hold queue full, request not admitted", "key", key)
+		if overflowServesPlaceholder(ir) {
+			serveStaticResponse(w, r, cm.reader, ir, ir.Spec.ColdStart.Placeholder.Response, http.StatusServiceUnavailable)
+		} else {
+			w.Header().Set("Retry-After", queueOverflowRetryAfter)
+			http.Error(w, "cold-start hold queue full", http.StatusServiceUnavailable)
+		}
+		return
+	}
+
```
linkvt: The counting middleware is now doing more than counting, I don't think that we should add that logic to this middleware so that it continues to have a single responsibility only.

interceptor/middleware/counting.go
```diff
@@ -35,6 +56,18 @@ func (cm *Counting) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	key := k8s.ResourceKey(ir.Namespace, ir.Name)
 
+	if cm.holdQueueFull(ir, key) {
+		cm.instruments.RecordColdStartRejection(ir.Name, ir.Namespace)
+		util.LoggerFromContext(ctx).V(1).Info("cold-start hold queue full, request not admitted", "key", key)
```
linkvt: We should probably not log in the request path, if the queue is full we might still impact the interceptor pod I guess.

interceptor/middleware/counting.go
```diff
@@ -3,26 +3,47 @@ package middleware
...
 	"github.com/kedacore/http-add-on/interceptor/metrics"
+	httpv1beta1 "github.com/kedacore/http-add-on/operator/apis/http/v1beta1"
 	"github.com/kedacore/http-add-on/pkg/k8s"
 	"github.com/kedacore/http-add-on/pkg/queue"
 	"github.com/kedacore/http-add-on/pkg/util"
 )
 
+// queueOverflowRetryAfter is the Retry-After value (in seconds) sent with the
+// 503 response when a route's cold-start hold queue is full.
+const queueOverflowRetryAfter = "5"
```
linkvt: Is there a reason we set this header? I mean I understand the purpose but if the user didn't specify something I'm not sure if we should just set a default that might not fit?

interceptor/middleware/placeholder.go
```diff
@@ -37,11 +38,20 @@ func (p *Placeholder) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	if ir.Spec.ColdStart != nil && ir.Spec.ColdStart.Placeholder != nil && ir.Spec.ColdStart.Placeholder.Response != nil {
 		serviceKey := ir.Namespace + "/" + ir.Spec.Target.Service
-		if !p.readyCache.HasReadyEndpoints(serviceKey) {
+		if !p.readyCache.HasReadyEndpoints(serviceKey) && !holdsRequests(ir) {
```
linkvt: Do we need/want to have the request holding/rejection logic in two middlewares? Similar to the counting middleware comment, we should maybe keep that in a dedicated middleware?

operator/apis/http/v1beta1/interceptorroute_types.go
```diff
@@ -200,6 +215,26 @@ type ColdStartSpec struct {
...
+	// unlimited). When set together with a placeholder and
+	// overflow: Placeholder, requests are held up to this limit instead of
+	// receiving the placeholder immediately; only overflowing requests
+	// receive it.
+	// The limit applies per interceptor replica: the effective cluster-wide
+	// hold capacity is maxQueueDepth multiplied by the number of
+	// interceptor replicas.
+	// +optional
+	// +kubebuilder:validation:Minimum=1
+	MaxQueueDepth *int32 `json:"maxQueueDepth,omitzero"`
```
linkvt: `MaxQueueDepth` sounds really technical, I'm wondering if there isn't a better name like max pending requests or such?

interceptor/metrics/instruments.go
```diff
@@ -18,6 +18,7 @@ const (
 	MetricRequestConcurrency = "interceptor.request.concurrency"
 	MetricRequestCount       = "interceptor.request.count"
 	MetricRequestDuration    = "interceptor.request.duration"
+	MetricColdStartRejected  = "interceptor.coldstart.rejected"
```
linkvt: Maybe something like `...rejections` or so to stay more consistent with these and the scaler metric names?

linkvt (commented):
Good idea, I left a few comments but I think it makes sense adding as this is something we can't do in ingress controllers, i.e. limitting only pending requests.

Fedosin:
@linkvt Thanks for the review! All points addressed in a2c3fb9:

1. **Counting single responsibility** — the admission logic moved out of `Counting` (which is now byte-for-byte back to main) into a dedicated `ColdStart` middleware. Since the hold/immediate-placeholder decision was split across two middlewares, `ColdStart` also absorbs `Placeholder`, so everything that happens to a request while the backend is cold lives in one file. It runs after `Counting`, so held and placeholder-served requests are still counted and drive scale-from-zero. One side effect to be aware of: rejected requests now briefly pass through the counter (they're rejected *after* the increment, since the placeholder responses must stay counted), so they add to the monotonic request-count used by rate-based scaling — arguably correct, since they represent real demand.
2. **No logging in the request path** — dropped the log line entirely; the `interceptor.coldstart.rejections` metric is the signal.
3. **Retry-After** — removed; a plain 503 now. If a use case for a retry hint comes up later, we can add it as a per-route option rather than an arbitrary default.
4. **Naming** — renamed to `coldStart.maxPendingRequests` and `KEDA_HTTP_COLD_START_MAX_PENDING_REQUESTS` (matches `targetPendingRequests` vocabulary), and the metric to `interceptor.coldstart.rejections`.

interceptor/metrics/instruments.go
```diff
@@ -15,9 +15,10 @@ const (
 	meterName   = "keda-interceptor-proxy"
 	ServiceName = "keda-http-interceptor"
 
-	MetricRequestConcurrency = "interceptor.request.concurrency"
-	MetricRequestCount       = "interceptor.request.count"
-	MetricRequestDuration    = "interceptor.request.duration"
+	MetricRequestConcurrency  = "interceptor.request.concurrency"
+	MetricRequestCount        = "interceptor.request.count"
+	MetricRequestDuration     = "interceptor.request.duration"
+	MetricColdStartRejections = "interceptor.coldstart.rejections"
```
linkvt: Really minor but we should probably also use `cold_start` here similar to how we split multiple words in the scaler metrics: https://github.com/kedacore/http-add-on/blob/1fd725ac13003e073e5b9777792a34705b69ff1e/scaler/metrics/instruments.go#L20

interceptor/middleware/cold_start_test.go
```diff
@@ -0,0 +1,586 @@
...
+			},
+			wantNextCalled: false,
+		},
+		"LimitWithoutPlaceholderOverflowServesImmediately": {
+			coldStart: &httpv1beta1.ColdStartSpec{
+				MaxPendingRequests: &limit,
+				Placeholder: &httpv1beta1.ColdStartPlaceholder{
+					Response: &httpv1beta1.StaticResponse{Body: &body},
+				},
+			},
```
linkvt, line 200: I think nothing changes when we set overflow to reject here just to make sure we dont test an impossible code path, as CRD validation enforces reject/placeholder.

linkvt (approved):
Looks good to me! I found these 2 minor things we should probably fix as it is quite simple but approving already 👍

Fedosin:
@linkvt Both follow-ups done in 7ef3599:

1. Metric renamed to `interceptor.cold_start.rejections` — good catch, matches `scaler.pinger.unreachable_pods` & co.
2. Test now sets `overflow: Reject` explicitly, so it exercises the post-defaulting object shape rather than the impossible unset path.

Thanks for the approval!

PR #1742, opened 2026-08-05
fix(interceptor): detect stalled routing table refresh in readiness probe

### LINKED ISSUE #1741: Interceptor readiness probe cannot detect a stalled routing table refresh
### Report

The interceptor's `/readyz` probe is backed by a single health check: the routing
table's `HasSynced()`. That is a latch that can never go false once it has passed —
the code acknowledges this itself:

```go
// pkg/routing/table.go
func (t *table) HealthCheck(_ context.Context) error {
	// TODO: HasSynced never fails after passing once, it is not testing health over time
	if !t.HasSynced() {
		return errNotSyncedTable
	}

	return nil
}
```

If the refresh loop stops making progress — a wedged informer, a signaler that
stopped firing, an error path in `refreshMemory` — the interceptor keeps serving an
increasingly stale routing table while reporting ready.

### Expected Behavior

`/readyz` reflects health over time: when the routing table has not been
successfully refreshed for longer than a multiple of the informer resync period
(`KEDA_HTTP_SCALER_CONFIG_MAP_INFORMER_RSYNC_PERIOD`, default 60m), the probe
should fail so the pod is pulled out of rotation.

### Actual Behavior

Once the initial sync has passed, `/readyz` returns 200 forever, no matter how
stale the routing table becomes. The user-facing symptom is 404s, or traffic sent
to deleted backends, on one interceptor pod out of many — exceptionally hard to
diagnose without this signal.

1. Deploy the interceptor with at least one `InterceptorRoute`/`HTTPScaledObject`.
2. Disrupt the refresh path (e.g. block informer events) without killing the process.
3. Observe that `/readyz` keeps returning 200 while the served routing table
   diverges from the actual route objects.

Note: identified by code review of `main` @ `2ca9930c`, not reproduced at runtime.

```
n/a — the defect is the absence of a failure signal
```

### HTTP Add-on Version

`main` @ `2ca9930c` (post-0.15.0)

### Kubernetes Version

Any

### Platform

Any

Yes

### Anything else?

Related closed issues for the opposite failure mode of the same check (failing at
startup): #1452, #1459.

A proposed fix exists: record `lastRefreshTime` and route count on each successful
`refreshMemory`, re-sync the table on the resync-period cadence (an informer
watching zero objects delivers no resync events, so a periodic refresh is needed
to avoid false positives on empty tables), and fail `HealthCheck` when the last
successful refresh is older than a multiple of that period.

Fedosin:
The interceptor's `/readyz` probe used the routing table's `HasSynced()` latch as its only input, and that latch can never go false once it has passed. A refresh loop that stops making progress left the pod serving an increasingly stale routing table while reporting ready, surfacing only as unexplained 404s or traffic to deleted backends on one pod out of many.

What changed:

- The routing table records the time of the last successful rebuild, and `HealthCheck` fails once that time is older than `MaxStaleness`, so a stalled refresh loop surfaces as 503 on `/readyz`.
- The table also rebuilds itself periodically, because informer resync delivers no events for an empty store — a purely signal-driven freshness check would false-positive on interceptors watching zero routes.
- The knobs travel together as a `routing.HealthConfig` struct, wired from `KEDA_HTTP_ROUTING_TABLE_REFRESH_INTERVAL` (default `1m`) and `KEDA_HTTP_ROUTING_TABLE_MAX_STALENESS` (default `5m`), following the `ColdStartConfig` + env var pattern from #1733. Setting the staleness to `0` restores the previous behavior.
- `MustParseServing` rejects a staleness deadline with no periodic rebuild behind it, or one shorter than two rebuild intervals, since both fail readiness on a healthy pod.
- New `interceptor.routing_table.last_rebuild_timestamp` metric, so a stalling loop can be alerted on before readiness pulls the pod out of the Service.
- The probe logs a failing check on onset and then every five minutes rather than every second, this being the first health check that can keep failing indefinitely.

### What this does and does not detect

It detects a refresh loop that stopped making progress. It does not detect a stale informer cache: after the initial sync, reads from an informer return immediately whether or not its watch is still healthy, so a rebuild can succeed against data that no longer reflects the cluster. Catching that needs a watch-level signal (`cache.Options.DefaultWatchErrorHandler`) and is left as follow-up.

### Why the knobs are generous and configurable

Failing readiness is not free. The scaler polls only ready interceptor endpoints, and zero ready endpoints is deliberately not an error path, so it reports zero pending requests for every metric and KEDA scales the workloads toward `minReplicaCount`. The conditions detected here are per-pod, so they should not correlate across replicas, but the env vars exist so the deadline can be widened or the check disabled outright.

### Changes since the last review

- The two durations are no longer hardcoded in `main`, and no longer derived from `CacheSyncPeriod`: they are env vars with startup validation.
- `lastRefresh` became `lastRebuild`, and the doc comments, error text and changelog entry now say "rebuild" and state what the check does not prove.
- Added the rebuild-timestamp metric, reported through a `HealthConfig.OnRebuild` hook so `pkg/routing` keeps no dependency on a metrics package, rather than re-adding the accessor that was dropped earlier.
- Fixed the every-second log spam that this change would otherwise have introduced in `handler.Probe`.
- Test boilerplate moved into a `newTestTable` helper, and `TestTablePeriodicRefresh` now observes the hook instead of reaching into private state.
- Drive-by: `TestProbePeriodicCheck` had an unsynchronized counter that races under `-race`, which CI does not run.
- Rebased onto `main`, resolving conflicts with #1733 in `CHANGELOG.md`, `interceptor/config/serving.go` and `interceptor/metrics/instruments.go`.

### Checklist

- [x] Documentation is updated if needed ([`README.md`](../README.md), [keda-docs](https://github.com/kedacore/keda-docs/tree/main/content/http-add-on)) — keda-docs needs the two new env vars and the new metric, to follow in a docs PR

Fixes #1741


pkg/routing/test/table.go
```diff
@@ -36,6 +37,14 @@ func (t Table) HasSynced() bool {
...
 }
 
+func (t Table) LastRefreshTime() time.Time {
+	return time.Time{}
+}
+
+func (t Table) RouteCount() int {
+	return len(t.Memory)
+}
+
```
linkvt: Do we really need those? It seems they arent used in the health check at all and following YAGNI it would make sense to me to not add them if we don't need them.

interceptor/main.go
```diff
@@ -124,7 +124,16 @@ func run() error {
...
-	routingTable := routing.NewTable(ctrlCache, queues)
+	// Besides reacting to informer events, the routing table rebuilds itself
+	// on the cache sync period cadence. Readiness then fails once no
+	// successful refresh happened within twice that period, surfacing a
+	// stalled refresh loop instead of serving a silently stale table.
+	routingTable := routing.NewTable(
+		ctrlCache,
+		queues,
+		routing.WithRefreshInterval(servingCfg.CacheSyncPeriod),
+		routing.WithMaxStaleness(2*servingCfg.CacheSyncPeriod),
```
linkvt: PR 1733 used a config struct, is that maybe something we could also use here to be more consistent or is that intended?
I would also guess we always need to set these two values together so making them potentially optional would not be intended, or wdyt?

interceptor/main.go
```diff
@@ -124,7 +124,16 @@ func run() error {
 	}
 
 	queues := queue.NewMemory()
-	routingTable := routing.NewTable(ctrlCache, queues)
+	// Besides reacting to informer events, the routing table rebuilds itself
+	// on the cache sync period cadence. Readiness then fails once no
+	// successful refresh happened within twice that period, surfacing a
+	// stalled refresh loop instead of serving a silently stale table.
```
linkvt: nit: comment is quite verbose and restates what table.go already has in its comments, but up to you.

interceptor/main.go
```diff
@@ -124,7 +124,16 @@ func run() error {
...
+	routingTable := routing.NewTable(
+		ctrlCache,
+		queues,
+		routing.WithRefreshInterval(servingCfg.CacheSyncPeriod),
```
linkvt: I'm not sure if we should use `servingCfg.CacheSyncPeriod` considering we would get unhealthy after 2h by default? See https://github.com/kedacore/http-add-on/blob/d8b96cea0c871beebbc10874b64d47278d047096/interceptor/config/serving.go#L21-L23

pkg/routing/table_test.go
```diff
@@ -150,6 +150,135 @@ func TestTableHealthCheck(t *testing.T) {
 	}
 }
 
+func TestTableHealthCheckStaleness(t *testing.T) {
+	cl := newTestClient()
+	tbl := NewTable(cl, queue.NewMemory(), WithMaxStaleness(100*time.Millisecond))
```
linkvt: I think we should add a test using both config knobs as well as this is the way we do it by default.

pkg/routing/test/table.go
Fedosin: Good point, removed in 8735cd8 — both methods are gone from the interface and the fake, and the route count tracking with them. The health check only needs the internal timestamp. We can bring accessors back when the /routes endpoint or metrics work needs them.

interceptor/main.go
Fedosin: Agreed on both counts — done in 8735cd8. NewTable now takes a routing.HealthConfig struct (same shape as ColdStartConfig in #1733), so the two values are always set together rather than being independently optional options.

interceptor/main.go
Fedosin: Trimmed in 8735cd8 — it is a one-liner now, the rationale lives in table.go.

interceptor/main.go
Fedosin: Fair point — decoupled in 8735cd8. The cadence no longer piggy-backs on the cache sync period (which is really a cache performance knob, and has a removal TODO of its own). The table now re-syncs every minute and readiness fails after five minutes without a successful refresh, so a stall is detected in minutes rather than 2h. Not adding an env var for it — happy to if you would rather have it configurable.

pkg/routing/table_test.go
Fedosin: Added in 8735cd8: TestTableHealthCheckStaysHealthyWithPeriodicRefresh constructs the table with both knobs set (the way main wires it) and asserts the table is still healthy well past MaxStaleness because the periodic refresh keeps it fresh.

interceptor/metrics/instruments.go
```diff
@@ -15,11 +15,12 @@ const (
...
-	MetricRequestCount        = "interceptor.request.count"
-	MetricRequestDuration     = "interceptor.request.duration"
-	MetricColdStartDuration   = "interceptor.cold_start.duration"
-	MetricColdStartRejections = "interceptor.cold_start.rejections"
+	MetricRequestConcurrency      = "interceptor.request.concurrency"
+	MetricRequestCount            = "interceptor.request.count"
+	MetricRequestDuration         = "interceptor.request.duration"
+	MetricColdStartDuration       = "interceptor.cold_start.duration"
+	MetricColdStartRejections     = "interceptor.cold_start.rejections"
+	MetricRoutingTableLastRebuild = "interceptor.routing_table.last_rebuild_timestamp"
```
linkvt, line 23: Is there a value in exposing this as a metric? To me this seems to be quite an implementation detail that might change.

interceptor/handler/probe.go
```diff
@@ -68,18 +78,26 @@ func (ph *Probe) check(ctx context.Context) {
...
 			ph.healthy.Store(false)
+			ph.consecutiveFailures++
 
-			// Log at info level before the first successful check to avoid
-			// noisy error logs during normal startup sequencing.
-			if ph.hasBeenHealthy.Load() {
-				logger.Error(err, "health check failed")
-			} else {
-				logger.Info("waiting for health check to pass", "error", err)
+			if ph.consecutiveFailures == 1 || ph.consecutiveFailures%failureLogInterval == 0 {
```
linkvt, line 83: This seems to be a bit complex compared to just logging when it goes from healthy to unhealthy considering the 300s interval above assuming checks every second.

linkvt (commented):
I just reviewed this PR again and while I think the changes are very clear, I'm not really sure what exact issue we would fix and if it isn't a bit too overengineered for that (e.g. considering the metric, a callback or special log handling).

Can you tell me again which bug would cause "a wedged informer, a signaler that stopped firing, an error path in refreshMemory" (from the issue description)?
From my understanding:
- Errors in refreshMemory are propagated and cause a pod restart.
- How would a signaler stop firing?
	- This would be a go bug I guess?
- How would an informer get stuck?
	- If that would be the case and if controller-runtime would not handle this which is what I would expect, we would still read from the stale cache I think and don't see the issue even with this approach?

So with this in mind I'm really unsure about what we actually fix with these changes.

PR #1777, opened 2026-09-01
fix(deps): update KEDA to v2.20.2 and align Kubernetes libraries

Fedosin:
Supersedes #1703 and #1527, which are deadlocked against each other and cannot merge independently.

## Why both bumps have to move together

KEDA's `go.mod` pins `k8s.io/*` to v0.35.5 and `controller-runtime` to v0.23.3 using `replace` directives. Go does not apply a dependency's `replace` directives to consuming modules, so we resolve the un-pinned `require` versions that KEDA itself never builds against:

- **#1703** (KEDA v2.20.0 alone) resolves `client-go` v0.36.0 against `controller-runtime` v0.23.3. That pair does not compile — `client-go` added `HasSyncedChecker` to `ResourceEventHandlerRegistration`, which `controller-runtime` v0.23.3 does not implement.
- **#1527** (`controller-runtime` v0.24.1 alone) leaves KEDA at v2.18.3, whose webhook code calls the older `ctrl.NewWebhookManagedBy` signature.

Each PR fails on the opposite half of the same mismatch, so this updates them as one set.

## Changes

| Module | From | To |
|---|---|---|
| `github.com/kedacore/keda/v2` | v2.18.3 | v2.20.2 |
| `sigs.k8s.io/controller-runtime` | v0.22.5 | v0.24.1 |
| `k8s.io/{api,apimachinery,client-go,apiserver}` | v0.34.11 | v0.37.0 |
| `go` directive | 1.25.6 | 1.26.0 |

KEDA v2.20.2 rather than the v2.20.0 in #1703: it is the latest patch and matches the `KEDA_VERSION ?= 2.20.2` Helm chart that the e2e jobs already deploy, so the Go module and the deployed KEDA agree.

The advisory behind #1703 (GHSA-6w3m-4hhp-775q / CVE-2026-53572) is in KEDA's PostgreSQL scaler. This project imports only `apis/keda/v1alpha1` and `pkg/scalers/externalscaler`, so the vulnerable code was never reachable here — the bump clears the scanner alert rather than fixing an exploitable path.

### Follow-on changes required by the bump

- **Go 1.26 is now required to build** (KEDA v2.20.2 mandates it). Raising the `go` directive activates the `modernize` analyzer's `newexpr` and `errorsastype` rules that `.golangci.yml` already enables, producing `ptr.To(x)` → `new(x)` and `errors.As` → `errors.AsType` conversions. These cannot be split into a preparatory PR, since the new idioms only compile under Go 1.26. All of them land in tests and test helpers; no production logic changed.
- **CRDs regenerated** under `controller-tools` v0.21.0. The diff is only the `controller-gen.kubebuilder.io/version` annotation — no schema changes — but without it `make verify-manifests` fails.
- **`controller-runtime` v0.24 deprecates `pkg/scheme.Builder`** (SA1019), which fails `make lint`. Suppressed with `//nolint:staticcheck` rather than migrated, because `SchemeBuilder` is exported from the API packages and switching to `runtime.NewSchemeBuilder` changes its exported type — a breaking change for importers that deserves its own PR. Flagging explicitly for reviewer input.

## Verification

Run locally on Go 1.26.0, all clean:

- `go build ./...`
- `go vet ./...` and `go vet -tags e2e ./...`
- `golangci-lint run` — 0 issues
- `go test ./...` — all packages pass
- `make verify-manifests` — config up to date
- `hack/validate-changelog.sh` — OK

### Checklist


linkvt:
Please check https://github.com/kedacore/http-add-on/pull/1703#issuecomment-5237370667

I don't think we need to force anything just a few days before the 2.21 release.

PR #1780, opened 2026-09-03
fix(interceptor): report 499 instead of 502 for client-cancelled requests

### LINKED ISSUE #1779: Interceptor reports 502 when the client cancels the request
### Report

When a client goes away before the response is ready, the interceptor answers 502 Bad Gateway
A common way to trigger this is a web map: the browser cancels the image requests still in flight as soon as the user pans or zooms.

The proxy error handler already separates timeouts out into 504, but there is no case for a cancelled request, so those fall through to the 502 default:

https://github.com/kedacore/http-add-on/blob/v0.15.0/interceptor/handler/upstream.go#L130-L140

When the client disconnects, `httputil.ReverseProxy` calls this handler with `context.Canceled`.

### Expected Behavior

A request the client cancelled should be counted differently from a request the backend failed to answer, so that 502 keeps meaning "the backend failed".

nginx already does this. It uses 499 (client closed request) for exactly this
case.

### Actual Behavior

Cancelled requests are counted as 502, mixed in with real backend failures.

1. Put any app behind the add-on and let it scale up to at least one ready replica. Use an endpoint that takes a second or two to answer.
2. Start a request and hang up before the response arrives, for example `curl --max-time 0.2 http://<host>/slow-endpoint`.
3. Scrape the interceptor metrics endpoint on port 2223 and look at `interceptor_request_count_total`. The cancelled request is counted with `code="502"`, even though the backend never failed.

```
example
```

### HTTP Add-on Version

0.15.0 (chart keda-add-ons-http 0.15.0)

### Kubernetes Version

1.34

### Platform

Amazon Web Services

Yes

### Anything else?

_No response_

pujitha24:
Report the interceptor's proxy `ErrorHandler` distinguishes a client-cancelled request (`context.Canceled`, e.g. the browser aborts an in-flight request) from a genuine backend failure, returning `499 Client Closed Request` (nginx's convention) instead of the generic `502 Bad Gateway`. This keeps `interceptor_request_count_total{code="502"}` meaning "the backend failed" instead of being mixed with requests the client itself walked away from.

### Checklist

Fixes #1779

### DIFF
--- a/interceptor/handler/static.go
+++ b/interceptor/handler/static.go
@@ -3,6 +3,7 @@ package handler
 import (
 	"net/http"
 
+	kedahttp "github.com/kedacore/http-add-on/pkg/http"
 	"github.com/kedacore/http-add-on/pkg/k8s"
 	"github.com/kedacore/http-add-on/pkg/routing"
 	"github.com/kedacore/http-add-on/pkg/util"
@@ -18,6 +19,10 @@ func NewStatic(statusCode int, err error) http.Handler {
 		stream := util.UpstreamURLFromContext(ctx)
 
 		statusText := http.StatusText(statusCode)
+		if statusText == "" && statusCode == kedahttp.StatusClientClosedRequest {
+			// 499 isn't a registered IANA status code, so http.StatusText doesn't know it.
+			statusText = "Client Closed Request"
+		}
 		routingKey := routing.NewKeyFromRequest(r)
 		namespacedName := k8s.NamespacedNameFromObject(ir)
 		logger.Error(err, statusText, "routingKey", routingKey, "namespacedName", namespacedName, "stream", stream)
--- a/interceptor/handler/static.go
+++ b/interceptor/handler/static.go
@@ -20,8 +20,7 @@ func NewStatic(statusCode int, err error) http.Handler {
 
 		statusText := http.StatusText(statusCode)
 		if statusText == "" && statusCode == kedahttp.StatusClientClosedRequest {
-			// 499 isn't a registered IANA status code, so http.StatusText doesn't know it.
-			statusText = "Client Closed Request"
+			statusText = kedahttp.StatusTextClientClosedRequest
 		}
 		routingKey := routing.NewKeyFromRequest(r)
 		namespacedName := k8s.NamespacedNameFromObject(ir)
--- a/interceptor/handler/upstream.go
+++ b/interceptor/handler/upstream.go
@@ -6,6 +6,7 @@ import (
 	"net"
 	"net/http"
 	"net/http/httputil"
+	"sync/atomic"
 	"time"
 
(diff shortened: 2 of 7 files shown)


interceptor/handler/static.go
```diff
@@ -18,6 +19,10 @@ func NewStatic(statusCode int, err error) http.Handler {
 		stream := util.UpstreamURLFromContext(ctx)
 
 		statusText := http.StatusText(statusCode)
+		if statusText == "" && statusCode == kedahttp.StatusClientClosedRequest {
+			// 499 isn't a registered IANA status code, so http.StatusText doesn't know it.
+			statusText = "Client Closed Request"
```
linkvt: Could we move this status text also to a const close to the status code?

interceptor/handler/upstream.go
```diff
@@ -35,9 +36,14 @@ type Upstream struct {
...
 	tracingCfg             config.Tracing
+	draining               *atomic.Bool
 }
 
-func NewUpstream(baseTransport *http.Transport, reader client.Reader, tracingCfg config.Tracing, responseHeaderTimeout time.Duration) *Upstream {
+// NewUpstream creates an Upstream that forwards requests to the URL resolved by
+// upstream routing middleware. draining, if non-nil, is set while the interceptor
+// is shutting down; it is consulted to avoid misreporting a request cancelled by
+// the server's own shutdown drain as a client-cancelled request.
+func NewUpstream(baseTransport *http.Transport, reader client.Reader, tracingCfg config.Tracing, responseHeaderTimeout time.Duration, draining *atomic.Bool) *Upstream {
```
linkvt: We set draining to true when we start the graceful shutdown. This doesnt mean that requests are immediately cancelled, requests can still be accepted and handled correctly.

The sequence is:
1. set draining to true
2. wait for shutdown delay
3. start the shutdown
4. wait for draintimeout to evict open connections

This also means: when draining is true, a request could arrive and a client could cancel it (during step 2).

If the idea was to track metrics for requests of actual server shutdowns, I don't think we need to handle this as it would only affect the time step 4 takes to close open connections.
Considering that:
- using 502 for draining requests is not more correct than 499
- the interceptor is draining anyways and shutting down in the next few seconds anyways so its questionable if this metric get's picked up anyways
- Code becomes simpler without the wiring for draining into upstream

linkvt (commented):
Hi, thanks for your PR!

It looks correct to me in general for the normal case, the draining case seems not correct to me though, see my review comments.
IMO we should remove the draining wiring, the rest then looks good to me 🙂

linkvt (approved):
Thanks, LGTM but leaving unmerged in case other reviewers take a look at this the next few days.

End of human interaction 🤖

pujitha24:
This is approved with green CI and no outstanding comments as far as I can tell — ready whenever you have a moment. Happy to rebase first if you'd like it freshened.

PR #1793, opened 2026-09-11
fix(scaler): suppress duplicate external-push activation events

### LINKED ISSUE #1792: external-push: workloads never scale to zero when many routes are active at once
## Report

With `external-push` ScaledObjects driven by `InterceptorRoute`, workloads stop scaling to
zero once a number of routes have been active at the same time. `lastActiveTime` is
refreshed continuously on every route -- including routes whose condition reads
`Active=False` / `ScalerCooldown` -- so `cooldownPeriod` never elapses and nothing reaches
`idleReplicaCount`. The HPA metric for those triggers reads `0` and the pods receive no
requests.

With 60 routes it happens every time. With 3 routes we could not reproduce it. It also
happens with both `requestRate` and `concurrency`. We have not found the cause.

## Expected Behavior

After traffic stops, the workload scales to `idleReplicaCount` once `cooldownPeriod`
elapses.

## Actual Behavior

60 routes, traffic on for 90s then off, sampled every 30s, `cooldownPeriod: 120`:

```
sample   Active=True   lastActiveTime age   replicas
t+30s    46 / 60       all 60 under 10s     60 x 1
t+120s   46 / 60       all 60 under 10s     60 x 1
t+300s   42 / 60       all 60 under 10s     60 x 1
```

`lastActiveTime` tracks the sample clock exactly (13:33:28 -> 13:37:59) and no route ever
scales down. We left it over an hour with no recovery.

Install KEDA 2.20.1 and http-add-on chart 0.14.1 (appVersion 0.14.0).

**1. Create 60 targets** (3 does not reproduce it):

```sh
kubectl create ns keda-latch
for i in $(seq 1 60); do N=$(printf 't%03d' $i); cat <<YAML | kubectl apply -f -
apiVersion: apps/v1
kind: Deployment
metadata: { name: $N, namespace: keda-latch, labels: { app: $N } }
spec:
  selector: { matchLabels: { app: $N } }
  template:
    metadata: { labels: { app: $N } }
    spec:
      containers:
        - { name: whoami, image: traefik/whoami:v1.10, args: ["--port=80"], ports: [{ containerPort: 80 }] }
---
apiVersion: v1
kind: Service
metadata: { name: $N, namespace: keda-latch }
spec:
  selector: { app: $N }
  ports: [{ port: 80, targetPort: 80 }]
---
apiVersion: http.keda.sh/v1beta1
kind: InterceptorRoute
metadata: { name: $N, namespace: keda-latch }
spec:
  target: { service: $N, port: 80 }
  rules: [{ hosts: [$N.keda.local], paths: [{ value: "/" }] }]
  scalingMetric:
    requestRate: { targetValue: 5, window: 1m, granularity: 1s }
---
apiVersion: keda.sh/v1alpha1
kind: ScaledObject
metadata: { name: $N, namespace: keda-latch }
spec:
  scaleTargetRef: { name: $N }
  pollingInterval: 15
  cooldownPeriod: 120
  initialCooldownPeriod: 20
  minReplicaCount: 1
  idleReplicaCount: 0
  maxReplicaCount: 1
  triggers:
    - type: external-push
      metadata:
        scalerAddress: keda-add-ons-http-external-scaler.keda:9090
        interceptorRoute: $N
YAML
done
```

**2. Wait for all 60 to reach 0 replicas.**

**3. Send traffic to all 60 routes for ~90s:**

```sh
kubectl -n keda-latch run loadgen --image=curlimages/curl:8.7.1 --restart=Never -- sh -c '
URL=http://keda-add-ons-http-interceptor-proxy.keda.svc.cluster.local:8080/
while true; do
  for i in $(seq 1 60); do
    curl -s -o /dev/null -m 5 -H "Host: $(printf t%03d $i).keda.local" "$URL" || true
  done
done'
```

**4. Stop the traffic:** `kubectl -n keda-latch delete pod loadgen --now`

**5. Watch for a few minutes:**

```sh
kubectl -n keda-latch get scaledobject \
  -o custom-columns='NAME:.metadata.name,ACTIVE:.status.conditions[?(@.type=="Active")].status,LAST:.status.lastActiveTime'
kubectl -n keda-latch get deploy
```

All 60 stay at 1 replica, with `lastActiveTime` moving forward every few seconds.

## Workaround

Restarting `keda-operator` clears it -- the timestamps freeze and all 60 scale to zero on
the next cooldown. Restarting the scaler or interceptor does not help.

## Ruled out

- Component restarts. The reproduction above involves none.
- Number of ScaledObjects. 260 objects with only 3 active drained fine; what matters is how
  many are active at once.
- `scaler.replicas` (reproduced with 1 and 3), and scaler OOM (1Gi limit, no restarts).
- Request rate. 3 routes drained cleanly at both ~58 and ~0.02 req/s per route.

## Environment

KEDA 2.20.1, http-add-on chart 0.14.1 (appVersion 0.14.0), Kubernetes 1.36 (Docker
Desktop). Also seen on EKS 1.34.

linkvt:
Repeated StreamIsActive true events caused KEDA to process redundant activation and status updates, delaying scale-to-zero cooldown.

I reproduced the issue (see issue description) with 20 pods already and verified that this PR fixes it.

**Changes:**
- Emit push events only on inactive-to-active transitions.
- Add deterministic transition tests.
- Update stream tests for selective event delivery.
- Document fix in Unreleased changelog.

### Checklist

Fixes #1792 

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed scaler activation notifications so they are emitted only when transitioning from inactive to active.
  - Prevented repeated active updates from delaying the scale-to-zero cooldown.
  - Inactive and repeated active states no longer produce unnecessary notifications, improving scaling responsiveness.

- **Tests**
  - Added coverage for activation-transition behavior.
  - Updated streaming checks to verify expected timeouts for inactive states and responses for active states.


scaler/handlers_httpso_test.go
```diff
@@ -291,18 +293,28 @@ func TestStreamIsActive_HTTPSO(t *testing.T) {
...
 			}
 
-			// First will see if we can establish the stream and handle this
-			// error.
-			streamClient, err := client.StreamIsActive(ctx, testRef)
+			streamCtx := ctx
+			if !tc.expected && !tc.expectedErr {
+				var cancel context.CancelFunc
+				streamCtx, cancel = context.WithTimeout(ctx, 500*time.Millisecond)
+				defer cancel()
```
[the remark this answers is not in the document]
linkvt: Tests are going to be deleted very soon with v1 release as httpso specific, ignored.

linkvt:
@coderabbitai review

Just testing the tooling we enabled

scaler/handlers.go
```diff
@@ -92,17 +92,30 @@ func (e *scalerHandler) StreamIsActive(scaledObject *externalscaler.ScaledObject
...
+			if err := server.Send(active); err != nil {
 				e.lggr.Error(err, "error sending the active result in stream")
 				return err
 			}
 		}
 	}
 }
 
+// shouldSendActive returns the next stream state and whether active is a new activation.
+func shouldSendActive(previousActive, active bool) (nextActive, send bool) {
```
Fedosin: `nextActive` always equals `active`

scaler/handlers.go
```diff
@@ -92,17 +92,30 @@ func (e *scalerHandler) StreamIsActive(scaledObject *externalscaler.ScaledObject
...
 			}
-			err = server.Send(&externalscaler.IsActiveResponse{
-				Result: active.Result,
-			})
-			if err != nil {
+			var send bool
+			previousActive, send = shouldSendActive(previousActive, active.Result)
+			if !send {
+				continue
+			}
```
Fedosin:
```suggestion
    newlyActive := active.Result && !previousActive
    previousActive = active.Result
    if !newlyActive {
        continue
    }
```

Fedosin (changes requested):
Nice work tracking this down, and thanks for verifying it on a cluster.

One concern before merge: the tests don't seem to cover the part of the change that actually fixes the bug. If I break the state threading at the call site so `previousActive` never gets updated, the behavior goes straight back to sending an event every tick, and the whole `./scaler` package still passes. `TestShouldSendActive` only covers the helper in isolation, and nothing exercises the loop that carries the state across ticks.

Could you add something at the stream level that would fail in that case? As it stands, a later refactor of `StreamIsActive` could reintroduce #1792 with a green CI run.

scaler/handlers.go
linkvt: I simplified the code a bit and update the test, thanks for the feedback!

Fedosin (approved):
Thanks! Snyk failure is irrelevant.

